PR #1, opened 2025-03-17
chore: Update getting started

NickBlow:
The getting started script had an incorrect field, and a missing required field


sam-goodwin:
Thanks so much!

PR #35, opened 2025-04-04
feat(aws): add VPC resource

naorpeled:


alchemy/src/aws/vpc.ts
```diff
@@ -0,0 +1,384 @@
...
+              Ipv6IpamPoolId: props.ipv6IpamPoolId,
+              Ipv6NetmaskLength: props.ipv6NetmaskLength,
+            })
+          );
+
+          if (!createResponse.Vpc) {
+            throw new Error(
+              "Failed to create VPC: No VPC returned from AWS API"
+            );
+          }
```
sam-goodwin, line 412: What are the various failure cases we need to handle? 

Worth looking at terraform aws provider and including that as context when generating the resource.

https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/ec2/vpc_.go

alchemy/src/aws/vpc.ts
```diff
@@ -0,0 +1,525 @@
...
+
+          // Update modifiable attributes
+          // DNS Hostnames
+          if (props.enableDnsHostnames !== undefined) {
+            await modifyVpcDnsHostnames(
+              client,
+              vpcId,
+              props.enableDnsHostnames
+            );
+          }
```
sam-goodwin, line 366: What if last time we deployed when it was `true`, and then swapped to `undefined`?

sam-goodwin:
```ts
// old props (last time we created/updated)
this.props.enableDnsHostnames !== props.enableDnsHostnames
```

sam-goodwin: `this.props` (old props)
`props` (new props)

PR #58, opened 2025-04-18
fix: lambda function url

sam-goodwin:
The Lambda Function URL was not configured properly, this change inokes the create function url command

### DIFF
--- a/alchemy/src/aws/function.ts
+++ b/alchemy/src/aws/function.ts
@@ -1,14 +1,19 @@
 import {
+  AddPermissionCommand,
   Architecture,
   CreateFunctionCommand,
+  CreateFunctionUrlConfigCommand,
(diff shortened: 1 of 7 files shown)


alchemy/src/aws/function.ts
```diff
@@ -340,11 +366,157 @@ export const Function = Resource(
...
+                          AllowOrigins: props.url.cors.allowOrigins,
+                          ExposeHeaders: props.url.cors.exposeHeaders,
+                          MaxAge: props.url.cors.maxAge,
+                        }
+                      : undefined,
+                  })
+                );
+                functionUrl = createResult.FunctionUrl;
+
+                // Add public access permission for function URL
```
nickbalestra, line 483: noice

PR #59, opened 2025-04-19
add d1 migrations

lockr7:


alchemy/src/cloudflare/d1-database.ts
```diff
@@ -57,6 +58,18 @@ export interface D1DatabaseProps extends CloudflareApiOptions {
...
+   * Name of the table used to track migrations. Only used if migrationsDir is specified.
+   * This is analogous to wrangler's `migrations_table`.
+   */
+  migrationsTable?: string;
+
+  /**
+   * Directory containing migration SQL files. If not set, no migrations will be applied.
+   * This is analogous to wrangler's `migrations_dir`.
+   */
+  migrationsDir?: string;
```
sam-goodwin, line 79: We may want to pass in the list of files and their hash instead of the directory, so it will trigger update when they change and reapply migrations. 

Maybe something like this:
```ts
await D1Database("..", {
  migrations: await Migrations("./migrations")
});
```

And then it's just a function that lists the directory and passes in the files?

That way when a new file is added, it will reapply?

sam-goodwin: I do like the `migrationsDir` interface. 

Perhaps we should have a wrapper:

```ts
export async function D1Database(id: string, props) {
  const migrationFiles = await list(props.migrationsDir)
  return D1DatabaseResource(id, {
    ..props,
    migrationFiles
  })
}
```

sam-goodwin:
Thanks for putting together. We might want to add a test too. Maybe generate a simple migration and then check it applied correctly?

PR #60, opened 2025-04-20
removed hard-coded references

nickbalestra:
This pr to fix the bundle to allow for output to mjs and support of main outdir/outfile esbuild params. 
- Remove creating a magic "dist" if none is provided, throw if neither a outdir or outfile is passed, if both are passed, esbuild will correctly bark at you. 

Tested with the following options: 

```
const bundle = await Bundle("example-bundle", {
  entryPoint: "./src/index.mts",
  outdir: "./dist",
  format: "esm",
  platform: "node",
  target: "node20",
  external: ["@aws-sdk/*"],
  options: {
    outExtension: { ".js": ".mjs" },
  },
});
```

```
const bundle = await Bundle("example-bundle", {
  entryPoint: "./src/index.mts",
  outfile: "./dist/index.mjs",
  format: "esm",
  platform: "node",
  target: "node20",
  external: ["@aws-sdk/*"]
});
```

### DIFF
--- a/alchemy/src/esbuild/bundle.ts
+++ b/alchemy/src/esbuild/bundle.ts
@@ -138,7 +138,23 @@ export const Bundle = Resource(
     // Check that bundle created output an file for the given entrypoint
     // Use bundle metada data to retrieve its content
     const bundleOutputPath = Object.entries(result.metafile.outputs).find(
-      ([file, output]) => output.entryPoint === props.entryPoint
+      ([file, output]) => {
+        if (output.entryPoint === undefined) {
+          return false;
+        }
+        // resolve to absolute and then relative to ensure consistent result (e.g. ./src/handler.ts instead of src/handler.ts)
+        const relativeOutput = path.relative(
+          process.cwd(),
+          path.resolve(output.entryPoint)
+        );
+        return (
+          relativeOutput ===
+          path.relative(
+            process.cwd(),
+            path.resolve(process.cwd(), props.entryPoint)
+          )
+        );
+      }
     )?.[0];
 
     if (!bundleOutputPath) {
--- a/alchemy/src/esbuild/bundle.ts
+++ b/alchemy/src/esbuild/bundle.ts
@@ -119,24 +119,28 @@ export const Bundle = Resource(
     props: BundleProps
   ): Promise<Bundle> {
     // Determine output path
-    const outDirPath = getOutDirPath(props);
+    const outDirPath =
+      props.outfile != null ? path.dirname(props.outfile) : props.outdir;
+
+    if (outDirPath == null) {
+      throw new Error(
+        `You need to specify either outfile or outdir in your bundle configuration ${JSON.stringify(props)}`
+      );
+    }
 
     // Ensure output directory exists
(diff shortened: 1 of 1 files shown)


alchemy/src/esbuild/bundle.ts
```diff
@@ -170,12 +182,31 @@ export async function bundle(props: BundleProps) {
...
+        await fs.promises.unlink(filePath);
+      } else if (stats.isDirectory()) {
+        await fs.promises.rmdir(filePath, { recursive: true });
+      }
+    }
+  } catch (err) {
+    console.error(`Error cleaning directory ${dirPath}:`, err);
+    throw err;
+  }
+}
```
sam-goodwin: Maybe we don't need this function?

```ts
import fs from "node:fs/promises"

await fs.rm(dir, { recursive: true });
```

alchemy/src/esbuild/bundle.ts
```diff
@@ -170,12 +182,31 @@ export async function bundle(props: BundleProps) {
...
-      props.outdir || "dist",
-      path.basename(props.entryPoint, path.extname(props.entryPoint)) + ".js"
-    )
+function getOutDirPath(props: BundleProps) {
+  if (props.outfile != null) return path.dirname(props.outfile);
+  if (props.outdir != null) return props.outdir;
+
+  throw new Error(
+    `You need to specify either outfile or outdir in your bundle configuration ${JSON.stringify(props)}`
   );
```
sam-goodwin: Let's inline this where it's used since it's only used once.

alchemy/src/esbuild/bundle.ts
```diff
@@ -119,41 +119,53 @@ export const Bundle = Resource(
...
+    // Use bundle metada data to retrieve its content
+    let bundleOutputPath;
+    for (const file in result.metafile.outputs) {
+      if (result.metafile.outputs[file].entryPoint === props.entryPoint) {
+        bundleOutputPath = file;
+      }
+    }
+    if (!bundleOutputPath) {
+      throw new Error(`Unable to find a compiled file`);
+    }
```
sam-goodwin: I don't understand why this is necessary? Can't we deterministically know where it will place the bundle?

How sure are we that `result.metafile.outputs[file].entryPoint === props.entryPoint` is reliable and won't be messed up by esbuild adding a slash or something like that?

Perhaps it can be simplified to?
```ts
const bundleOutputPath = Object.entries(result.metafile.outputs)
  .find(([file, output]) => output.entryPoint === props.entryPoint)
  ?.[0];
```

nickbalestra: the issue is that if you were to rely on props as before, but let say esbuild change it to .mjs, then you might not able to find the file, I would try to use bundle output information as much as possible vs rely on inputs and go chase it. it should be more robust like this, but yeah can totally simplify.

nickbalestra: Let me also test for if this works well with source maps (output files >1 )

sam-goodwin: Might be worth adding a test with that too in `function.test.ts`

PR #72, opened 2025-04-21
fix(cloudflare): support non-existing config files

depsimon:
Currently if you run alchemy in a CI environment, it throws an error because it tries to open & write to a non-existing file.

This PR gracefully handle that scenario.

### DIFF
--- a/alchemy/src/cloudflare/auth.ts
+++ b/alchemy/src/cloudflare/auth.ts
@@ -239,6 +239,8 @@ interface WranglerConfig {
   refresh_token?: string;
   expiration_time?: string;
   scopes?: string[];
+  /** exists is `false` if the config file doesn't exist, like in CI */
+  exists?: boolean;
   /** @deprecated - this field was only provided by the deprecated v1 `wrangler config` command. */
   api_token?: string;
 }
@@ -258,6 +260,8 @@ async function getRefreshedAuthConfig(): Promise<WranglerConfig> {
 }
 
 async function writeWranglerConfig(config: WranglerConfig) {
+  if (config.exists === false) return
+
   const TOML = await import("@iarna/toml");
   const configPath = await findWranglerConfig();
   config = {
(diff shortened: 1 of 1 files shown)


sam-goodwin:
Is this because you're using `WranglerJson` resource directly? In `ViteSite`, I solved this by creating a stub `wrangler.jsonc` if it doesn't exist.

see: https://github.com/sam-goodwin/alchemy/blob/main/alchemy/src/cloudflare/vite-site.ts#L61

I suspect you're using Tanstack? We ned to add an equivalent TanStack resource

depsimon:
I do use Tanstack but I'm not sure it's related to the project, I suspect you would have the same issue trying to running alchemy with a Vite project in a CI. It happens before the `WranglerJson` resource is handled.

I figured the issue was thrown because it was trying to load the config if there is no API Token or Key provided. It tries to load the `oauth_token` from the wrangler config (which you have if you are signed in with wrangler).

You can check that behavior in the [`getCloudflareAuthHeaders`](https://github.com/sam-goodwin/alchemy/blob/main/alchemy/src/cloudflare/auth.ts#L116) and more specifically if it reaches the [`getRefreshedAuthConfig`](https://github.com/sam-goodwin/alchemy/blob/main/alchemy/src/cloudflare/auth.ts#L140C28-L140C50) call which in turns call the config functions.

If that makes sense to you?

Current workaround is this in my CI:

```
      - name: Touch config
        run: |
          mkdir -p /home/runner/.config/.wrangler/config
          touch /home/runner/.config/.wrangler/config/default.toml

      - name: Build & Deploy
        env:
          AUTH_GITHUB_CLIENT_SECRET: ${{ secrets.AUTH_GITHUB_CLIENT_SECRET }}
          AUTH_GITHUB_CLIENT_ID: ${{ vars.AUTH_GITHUB_CLIENT_ID }}
          BETTER_AUTH_SECRET: ${{ secrets.BETTER_AUTH_SECRET }}
          CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
          SECRET_PASSPHRASE: ${{ secrets.SECRET_PASSPHRASE }}
        run: bun run ./alchemy.run --stage prod
```

sam-goodwin:
Oh! I was confused. I thought this was related to `wrangler.jsonc`, not the auth. Got it. This is good, thanks!

PR #73, opened 2025-04-21
Feature add invokemode to aws function

nickbalestra:
- Added support for setting `invokeMode` for aws function url. 
- default to `BUFFERED`

### DIFF
--- a/alchemy/test/aws/function.test.ts
+++ b/alchemy/test/aws/function.test.ts
@@ -1,5 +1,6 @@
 import {
   GetFunctionCommand,
+  GetFunctionUrlConfigCommand,
   InvokeCommand,
   LambdaClient,
   ResourceNotFoundException,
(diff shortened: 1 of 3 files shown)


alchemy/test/aws/function.test.ts
```diff
@@ -514,5 +514,158 @@ describe("AWS Resources", () => {
...
+
+        // Test function URL invocation (now in RESPONSE_STREAM mode)
+        const streamTestEvent = { test: "response-stream-mode" };
+        const streamResponse = await fetch(func.functionUrl!, {
+          method: "POST",
+          headers: {
+            "Content-Type": "application/json",
+          },
+          body: JSON.stringify(streamTestEvent),
+        });
```
sam-goodwin, line 650: should we be testing that it does actually stream? Is that possible?

nickbalestra: i think we could,
perhaps something along those lines? 
```
// Check transfer encoding
const isChunked = response.headers.get('transfer-encoding') === 'chunked';

// Check for readable stream body
const isStream = response.body instanceof ReadableStream;

// Consuming the tream 
const reader = response.body.getReader();
...
  ```

File: alchemy/test/aws/function.test.ts [hunk 2]
```diff
@@ -514,5 +514,158 @@ describe("AWS Resources", () => {
...
+
+        expect(streamResponse.status).toBe(200);
+        const streamResponseBody = await streamResponse.json();
+        expect(streamResponseBody.message).toBe("Hello from bundled handler!");
+        expect(streamResponseBody.event).toEqual(streamTestEvent);
```
nickbalestra: oh this make sense! 
@sam-goodwin fixing this should actually test for it to be streaming

alchemy/test/aws/function.test.ts
sam-goodwin: Ah yeah, the header is worth testing. Not sure consuming the stream is helpful since won't even a standard JSON response be streamable?

nickbalestra:
this now both pull configuration and check that the setup for InvokeMode is correct, and in that case properly parse chunks to see the response

PR #74, opened 2025-04-21
feat(cloudflare): support configuring Worker Asset Config

sam-goodwin:


alchemy/src/cloudflare/vite-site.ts
```diff
@@ -82,12 +82,25 @@ export async function ViteSite<B extends Bindings>(
...
       name: props.name ?? id,
       entrypoint: props.main,
+      assets: {
+        html_handling: "auto-trailing-slash",
+        not_found_handling: "single-page-application",
+        run_worker_first: false,
+      },
       script: props.main
         ? undefined
         : `
```
lockr7, line 106: do we want to do return new Response("Not Found2", { status: 404 }); here instead too?

sam-goodwin: Good catch! Yes, we should do that.

PR #90, opened 2025-04-24
Add docs context to README

janwilmake:
[sam-goodwin/alchemy context](https://uithub.com/sam-goodwin/alchemy/tree/main/alchemy-web/docs?accept=text%2Fhtml&maxTokens=10000000)

May be a good way to allow people to learn alchemy faster with LLMs!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # Alchemy
 
+[![sam-goodwin/alchemy context](https://badge.forgithub.com/sam-goodwin/alchemy/tree/main/alchemy-web/docs?accept=text%2Fhtml&maxTokens=10000000)](https://uithub.com/sam-goodwin/alchemy/tree/main/alchemy-web/docs?accept=text%2Fhtml&maxTokens=10000000)
+
 Alchemy is an embeddable, zero-dependency, TypeScript-native Infrastructure-as-Code (IaC) library for modeling Resources that are Created, Updated and Deleted automatically.
 
 Unlike similar tools like Pulumi, Terraform, and CloudFormation, Alchemy is implemented in pure ESM-native TypeScript code with zero dependencies.


sam-goodwin:
Is this copying the docs verbatim or is it doing LLM processing?

janwilmake:
This is purely streaming the docs from the `docs` folder without any further processing. It will always remain up to date, there's no caching.

If desired you can apply more/other filters and generate a new badge by clicking "Add to readme". Further LLM processing filters and features will be added in near future.

sam-goodwin:
I've noticed it's a bit slow, are you streaming via github's API? Do you cache? 

Are you worried about github API request limits?

janwilmake:
> I've noticed it's a bit slow, are you streaming via github's API? Do you cache?

no and no. caching is an optimisation that will come in the new version announced y'day (https://x.com/janwilmake/status/1915774892500300281) but may take more time. Happy to hear more about your needs. Is speed more important than freshness? Or do you want both?

> Are you worried about github API request limits?

No, it uses the zip archive. far from hitting limits.

sam-goodwin:
> Is speed more important than freshness

I can achieve both with a simple script in my repo that bundles all of the `.md` files into one and publishes it as a static asset when deploying my vitepress site.

janwilmake:
I'm sure you can, and please do if you think it's better. But imho, it's not only the static asset that matters. It's the interface that gives people (and agents alike) more control about choosing the context they need.

P.S.
```
curl https://uithub.com/sam-goodwin/alchemy/tree/main/alchemy-web/docs -o llms-full.txt
```

sam-goodwin:
I didn't notice the filtering options. Interesting.

sam-goodwin:
Can we disable line numbers? I don't think they are necessary?

sam-goodwin:
Where should I cut issues, is uithub open source?

this? https://github.com/janwilmake/uit

janwilmake:
yes please, that's the right place (it's the new version which is currently in active development, and will eventually replace the current uithub.com - now accessible through uuithub.com)

> Can we disable line numbers? I don't think they are necessary?

`?lines=false`

Also, see https://uithub.com/openapi.html

PR #92, opened 2025-04-24
lil fixes for nuxt example

FutureExcited:
forgot to add index.ts last time

### DIFF
--- a/alchemy-web/docs/guides/cloudflare-nuxt-pipeline.md
+++ b/alchemy-web/docs/guides/cloudflare-nuxt-pipeline.md
@@ -16,10 +16,10 @@ cd my-nuxt-app
 bun install
 ```
 
-Install alchemy and Cloudflare:
+Install alchemy, Cloudflare, and worker types:
 
 ```sh
-bun add alchemy cloudflare
+bun add alchemy cloudflare @cloudflare/workers-types
 ```
 
 ## Configure Nuxt for Cloudflare
@@ -119,10 +119,10 @@ await app.finalize();
 
 ## Infer Binding Types
 
-Create an `src/env.d.ts` file to support type hints for Cloudflare bindings:
+Create an `env.d.ts` file to support type hints for Cloudflare bindings:
 
 ```typescript
-// src/env.d.ts
+// env.d.ts
 /// <reference types="@cloudflare/workers-types" />
 
 import type { website } from './alchemy.run';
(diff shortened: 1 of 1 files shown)


alchemy-web/docs/guides/cloudflare-nuxt-pipeline.md
```diff
@@ -119,10 +119,10 @@ await app.finalize();
 
 ## Infer Binding Types
 
-Create an `src/env.d.ts` file to support type hints for Cloudflare bindings:
+Create an `env.d.ts` file to support type hints for Cloudflare bindings:
 
 ```typescript
-// src/env.d.ts
+// env.d.ts
 /// <reference types="@cloudflare/workers-types" />
```
sam-goodwin, line 126: Oops, I think this reference is wrong. 

Should be env.d.ts instead of worker-types. 

See: https://github.com/sam-goodwin/alchemy/blob/main/examples/cloudflare-vite/src/env.d.ts

alchemy-web/docs/guides/cloudflare-nuxt-pipeline.md
```diff
@@ -136,6 +136,34 @@ declare module 'cloudflare:workers' {
 }
 ```
 
+## Create Worker Entry Point
+
+Create an `index.ts` file in the root of your project to handle requests:
+
+```typescript
+// index.ts
+/// <reference types="@cloudflare/workers-types" />
```
sam-goodwin, line 145: We can move this to `types` in `tsconfig.json`

alchemy-web/docs/guides/cloudflare-nuxt-pipeline.md
```diff
@@ -136,6 +136,34 @@ declare module 'cloudflare:workers' {
...
+    environment: WorkerEnv,
+    ctx: ExecutionContext
+  ): Promise<Response> {
+    const url = new URL(request.url);
+
+    if (url.pathname.startsWith("/api/")) {
+      return nitroApp.fetch(request, environment, ctx);
+    }
+
+    return environment.ASSETS.fetch(request);
```
sam-goodwin, line 162: I don't think we need this conditional and asset fetch anymore. 

Cloudflare will serve the assets directly and our worker won't be called. You can just call nitroApp.fetch directly. 

Worth testing.

alchemy-web/docs/guides/cloudflare-nuxt-pipeline.md
sam-goodwin: It might also be worth checking if we can add this to types in tsconfig and have it apply globally.

PR #105, opened 2025-04-27
chore(docs): add vitepress-plugin-group-icons for ::: code-block

ericclemmons:
I'm a `pnpm` user primarily, and this also started showing places that `bun`'s built-in `.ts` support has slight differences from others.

This is the start for updating other parts of the docs to have per-package-manager code blocks...

### Before

image

### After

image

image


sam-goodwin:
Love it, thanks for setting this up. You're right, we should have docs for each package manager.

sam-goodwin:
Do you know if it remembers your choice across pages and site visits?

PR #107, opened 2025-04-27
WranglerJson returns `spec` for re-use

ericclemmons:
**EDIT**

Opted to move the `wrangler.jsonc` generation into user-code.  There's benefit to move `processBindings` out for re-use in user-land, but I feel like this is in _niche_ territory.

--- 

_Previous motivation below..._

Users can use Alchemy for creating Resources (e.g. `R2Bucket`, etc.) that can be used for generating a dynamic `wrangler.jsonc` using these Resources' IDs.

However, today `WranglerJson` requires a `worker`.

In my real-world application, this adds **22 seconds** to Alchemy's deploy time (otherwise < 1 second)!

This PR makes `worker` optional so users can generate a dynamic `wrangler.jsonc` and run `wrangler dev|deploy` as normal, but with IaC resources (e.g. `binding`)

### DIFF
--- a/alchemy/src/cloudflare/wrangler.json.ts
+++ b/alchemy/src/cloudflare/wrangler.json.ts
@@ -52,6 +52,11 @@ export interface WranglerJson
    * Path to the wrangler.json file
    */
   path: string;
+
+  /**
+   * `wrangler.json` spec
+   */
+  spec: WranglerJsonSpec;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


sam-goodwin:
Forgot to commit changes to `wranger.json.ts`?

ericclemmons:
> Forgot to commit changes to `wranger.json.ts`?

Nope, in draft while I do TDD 🙏

PR #120, opened 2025-04-30
Add alchemy/serde export

ericclemmons:
https://alchemy.run/docs/advanced/serde.html#using-the-serialization-system is invalid:

```js
import { serialize, deserialize } from "alchemy/serde";
```
> Cannot find module 'alchemy/serde' or its corresponding type declarations.ts(2307)


alchemy/package.json
```diff
@@ -23,6 +23,7 @@
     "./github": "./lib/github/index.js",
     "./neon": "./lib/neon/index.js",
     "./os": "./lib/os/index.js",
+    "./serde": "./lib/serde/index.js",
```
sam-goodwin: I think we can just add them to the root index

```ts
// alchemy/src/index.ts
export * from "./serde.js";
```

ericclemmons: Cool, I'll do that. I'll also update docs to match!

ericclemmons: Resolved via 0c123accf43ef83ae57c6de4b7b44b5334fa855f

alchemy-web/docs/guides/custom-state-store.md
```diff
@@ -87,7 +87,7 @@ Here's a skeleton implementation:
 
 ```typescript
 import type { Scope } from "alchemy/scope";
-import { deserialize, serialize } from "alchemy/serde";
+import { deserialize, serialize } from "alchemy";
 import type { State, StateStore } from "alchemy/state";
```
ericclemmons: I don't think `alchemy/state` is exported either?  Nor is `scope`

ericclemmons: yep, confirmed via:
```ts
import type { Scope, State, StateStore } from "alchemy";
```

ericclemmons: Resolved via 119722c3496c746cca12b49b222323f7e68c4461

PR #131, opened 2025-05-02
feat(core): replace Resource

sam-goodwin:
The `replace` operation allows a Resource to mark a resource as being replaced and enqueue deletion of the old resource.

The old resource will be deleted after `app.finalize()` is called to ensure the new resource has flushed through to downstream resources (e.g. a new Table's ARN has propagated through and it is now safe to delete the old Table)

```ts
const Replacable = Resource(
  "Replacable",
  async function (
    this: Context<any, any>,
    _id: string,
    props: {
      name: string;
    },
  ) {
      // destrory logic
      return this.destroy();
    } else if (this.phase === "update") {
      if (props.name !== this.output.name) {
        // detect a change in `update` that requires creation of a new resource and deletion of the old
        this.replace();
        // return the new resource
        return this({
          name: `${props.name}-${++i}`,
        });
      }
    }
    return this({
      name: `${props.name}-${i}`,
    });
  },
);```

### DIFF
--- a/alchemy-web/docs/concepts/phase.md
+++ b/alchemy-web/docs/concepts/phase.md
@@ -2,10 +2,7 @@
 order: 2
 ---
 
-
 # Phase
-
-## Phases
 
 An Alchemy app can run in one of three phases:
 1. `"up"` - resources should be created, updated and deleted as necessary.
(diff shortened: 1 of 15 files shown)


sam-goodwin:
Tricky problem with REPLACE - what happens if a replaced resource contains nested resources/scopes? Should we also DELETE the nested resources? Need to be careful to not delete resources created by the new (replacement) resource.

PR #132, opened 2025-05-02
feat: add AWS Cloud Control API support

### LINKED ISSUE #64: AWS Cloud Control API
Using the AWS Cloud Control API, we can get broad and stable coverage of AWS Resources "for free" (one implementation to support [these resources](https://docs.aws.amazon.com/cloudcontrolapi/latest/userguide/supported-resources.html)).

## Desired Experience

A user imports this to create resources:
```ts
import AWS from "alchemy/aws/control";

await AWS.Lambda.Function("my-function", {
   // properties
});
```

The `properties` are 1:1 parity with the corresponding CloudFormation resource, e.g. [AWS::Lambda::Function](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html).

## Implementation

Write a script that downloads the CloudFormation JSON spec from [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html) and generate types from the Resources and Data Types within

```ts
export declare namespace AWS {
  namespace Lambda {
    // props, Resource interface, Resource function signature per Resource
    interface FunctionProps { .. }
    interface Function extends Resource<"AWS::Lambda::Function"> { .. }
    declare Function(id: string, props: FunctionProps): Promise<Function>
  }
  namespace EC2 { .. }
}
```

> [!CAUTION]
> Make sure these are type-only (`export declare namespace` instead of export namespace)

Implement a generic function, `createResourceType`, that constructs (and caches) an AWS Resource's lifecycle handler using the [AWS Cloud Control API](https://docs.aws.amazon.com/cloudcontrolapi/latest/userguide/what-is-cloudcontrolapi.html):
```ts
const awsResources = {}
export function createResourceType(serviceName: string, resourceName: string) {
  const typeName = `AWS::${serviceName}::${resourceName}`;
  return awsResources[typeName] ??= Resource(
    typeName, 
    async function(this, id: string, props: any) {
      let attributes: any;
        await client.send(new DeleteResourceCommand({ .. })
        return this.destroy();
      } else if (this.phase === "create") {
        attributes = await client.send(new CreateResourceCommand({ .. })
      } else {
        attributes = await client.send(new UpdateResourceCommand({ .. })
      }
      return this(attributes);
    }
  );
}
```

Implement a generic client using a `Proxy` to construct the Resource FQN (e.g. `AWS::Lambda::Function) and proxy requests to the generic resource created with `createResourceType`:
```ts
const GenericCloudControlResource = new Proxy(undefined, {
  get(_, serviceName: string) {
    return new Proxy(undefined, {
      get(_, resourceName: string) {
        return createResourceType(serviceName, resourceName);
      }
    }
  }
} 
```

Export the Proxy and cast to the generated `AWS` type:

```ts
export default GenericCloudControlResource as AWS;
```

Add an export to `alchemy/package.json`
```json
"exports": {
  "./aws/control": "./lib/aws/control/index.js"
}
```

## Deletion Problem

Alchemy relies on a global dictionary of Resource Type -> Resource Handler to perform deletions. That is why you see these funky imports at the top of my `alchemy.run.ts`:

https://github.com/sam-goodwin/alchemy/blob/f7d9929d52f8c0845859275773461687490fccea/alchemy.run.ts#L2-L12

If we do a type-only proxy solution, it is optimal for bundle size, but its lazy nature means when a resources becomes orphaned, the resource handler will not be registered, causing deletions to fail.

### Solution 1 

Instead of generating types only, we could change the `namespace`s to be physical and walk over them in `alchemy/aws/control/index.ts` to register all the calls upfront.

Then, a naked import will register all handlers for all services.
```ts
import "alchemy/aws/control";
```

> [!CAUTION]
> This might be memory heavy since there are hundreds/thousands of resources.

### Solution 2

We extend Alchemy's core library to support registering a deletion handler based on a wildcard

```ts
registerDeletionHandler("AWS::*", (context, id, props) => { .. })
```

That way we can keep the optimal bundle size and memory impact.

I lean towards Solution 2

naorpeled:
NOTE: this is still WIP

Closes #64


sam-goodwin:
So excited! Looking great

alchemy/src/aws/control/resource.ts
```diff
@@ -0,0 +1,328 @@
...
+            console.error(
+              `Error deleting resource ${resource.identifier}:`,
+              error
+            );
+          }
+        }
+      });
+
+      // Wait for all deletions to complete
+      await Promise.allSettled(deletionPromises);
```
sam-goodwin: Did this throw of any fail?

alchemy/src/aws/control/resource.ts
```diff
@@ -0,0 +1,328 @@
...
+    } catch (error: unknown) {
+      const cloudError = error as CloudControlError;
+      if (cloudError.code === "ResourceNotFoundException") {
+        // No resources found, not an error
+        console.log(`No resources found matching pattern ${pattern}`);
+      } else {
+        console.error(`Error listing resources for pattern ${pattern}:`, error);
```
sam-goodwin: Should we throw? What errors are swallowed here ?

alchemy/src/aws/control/resource.ts
```diff
@@ -0,0 +1,328 @@
...
+      const client = createCloudControlClient();
+
+      if (this.phase === "delete") {
+        try {
+          if (this.output?.id) {
+            await client.deleteResource(typeName, this.output.id);
+          }
+        } catch (error) {
+          // Log but don't throw on deletion errors
+          console.error(`Error deleting resource ${typeName}:`, error);
```
sam-goodwin: I think we only want to swallow errors that indicate the resource is already deleted. If we have a transient failure, deletion should fail and the user should retry.

alchemy/src/aws/control/resource.ts
```diff
@@ -0,0 +1,328 @@
...
+ *       }
+ *     ],
+ *     ProvisionedThroughput: {
+ *       ReadCapacityUnits: 5,
+ *       WriteCapacityUnits: 5
+ *     }
+ *   }
+ * });
+ */
+export const CloudControlResource = Resource(
```
sam-goodwin, line 201: Is this a duplicate/redudant?

alchemy/src/aws/control/resource.ts
```diff
@@ -0,0 +1,230 @@
...
+
+        const resource = await client.getResource(
+          props.typeName,
+          error.progressEvent.Identifier!,
+        );
+
+        response = await client.updateResource(
+          props.typeName,
+          error.progressEvent.Identifier!,
+          // TODO(sam): some of these properties may be "output" properties
```
naorpeled: @sam-goodwin not sure about this one actually.
Picking only the properties from the desired state wouldn't work right?

alchemy/test/aws/control/proxy.test.ts
```diff
@@ -0,0 +1,166 @@
...
+        ProvisionedThroughput: {
+          ReadCapacityUnits: 1,
+          WriteCapacityUnits: 1,
+        },
+      });
+
+      expect(table.TableName).toEqual(tableId);
+      expect(table.KeySchema).toHaveLength(1);
+      // @ts-expect-error TODO: fix this
+      expect(table.KeySchema[0].AttributeName).toEqual("id");
```
naorpeled: @sam-goodwin should we replace `{}` with `Record<string, unknown>`?

image

alchemy/src/aws/control/client.ts
```diff
@@ -0,0 +1,385 @@
...
+        // console.warn(
+        //   `Error polling operation ${operationId} (attempt ${attempts + 1}):`,
+        //   error,
+        // );
+
+        await new Promise((resolve) =>
+          setTimeout(resolve, Math.min(delay * 2, this.maxPollingDelay))
+        );
+        delay = Math.min(delay * 2, this.maxPollingDelay);
+        attempts++;
```
naorpeled: @sam-goodwin 
do you think the following would be good enough?

I'm not sure how to force any of those errors to properly test this

```suggestion
	if (RETRYABLE_ERRORS.includes(error.message)) {
	        await new Promise((resolve) =>
	          setTimeout(resolve, Math.min(delay * 2, this.maxPollingDelay))
	        );
	        delay = Math.min(delay * 2, this.maxPollingDelay);
	        attempts++;
        }
```

alchemy/test/aws/control/proxy.test.ts
```diff
@@ -0,0 +1,166 @@
...
+import AWS from "../../../src/aws/control";
+import { destroy } from "../../../src/destroy.js";
+import { BRANCH_PREFIX } from "../../util.js";
+// must import this or else alchemy.test won't exist
+import "../../../src/test/bun.js";
+
+const test = alchemy.test(import.meta);
+
+describe("AWS Cloud Control Proxy", () => {
+  const testId = `${BRANCH_PREFIX}-test-bucket-${Date.now()}`;
```
sam-goodwin: Do not use Date.now(). Each run should be deterministic and idempotent. This will create sporadic buckets. 

Instead, make sure the test's `finally` destroys the test scope. 

We should also add an adopt: boolean property to a CC resource that can be used to adopt the bucket if it already exists on first creation.

naorpeled: The creation failed although the last bucket with the same name was deleted

I'll try and see if I can reproduce that tomorrow.

Regarding the adopt boolean, sounds good to me

alchemy/test/aws/control/proxy.test.ts
sam-goodwin: I noticed that AssumeRolePolicyDocument is also `{}`. `Record<string, unknown>` would be better but I also wonder if that type is not properly documented

sam-goodwin: Ah I see, it's a `Json` primitive type.
```json
"AssumeRolePolicyDocument": {
  "Documentation": "http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-role.html#cfn-iam-role-assumerolepolicydocument",
  "UpdateType": "Mutable",
  "Required": true,
  "PrimitiveType": "Json"
},
```

alchemy/src/aws/control/client.ts
sam-goodwin: Not sure about max attempts? I'd expect the API to eventually fail

sam-goodwin:
I think we're missing the output attributes. Extending `ResourceProps` brings in the input properties, but each resource also has output attributes.

image

PR #144, opened 2025-05-07
Filter orphans without output

ericclemmons:
When working with custom Resources, I found that there can be situations where state wasn't correctly persisted or _something_, causing this error:

```console
90 |         if (!this.isErrored) {
91 |             // TODO: need to detect if it is in error
92 |             const resourceIds = await this.state.list();
93 |             const aliveIds = new Set(this.resources.keys());
94 |             const orphanIds = Array.from(resourceIds.filter((id) => !aliveIds.has(id)));
95 |             const orphans = await Promise.all(orphanIds.map(async (id) => (await this.state.get(id)).output));
                                                                                                       ^
TypeError: undefined is not an object (evaluating '(await this.state.get(id)).output')
      at <anonymous> (/Users/eric/Projects/ericclemmons/bmarks/packages/infra/node_modules/alchemy/lib/scope.js:95:99)
```

When I locally patched this to use `?.output`, the error went away & the app worked as expected.

### DIFF
--- a/alchemy/src/scope.ts
+++ b/alchemy/src/scope.ts
@@ -3,7 +3,7 @@ import type { Phase } from "./alchemy.js";
 import { destroy } from "./destroy.js";
 import { FileSystemStateStore } from "./fs/file-system-state-store.js";
 import type { PendingResource, ResourceID } from "./resource.js";
-import type { StateStore, StateStoreType } from "./state.js";
+import type { State, StateStore, StateStoreType } from "./state.js";
 
 const scopeStorage = new AsyncLocalStorage<Scope>();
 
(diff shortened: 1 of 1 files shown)


sam-goodwin:
Thanks for reporting this. I have also ran into it. I thought it was caused by a bug that was in R2 Rest State Store that I recently fixed in https://github.com/sam-goodwin/alchemy/pull/138

Can you confirm you're on latest version? This really should never be undefined and I don't think it's safe to ignore it. Better to FATAL here since that was what caught the bug.

ericclemmons:
I was on `0.15.2`, but I'll try again on `0.15.4`

alchemy/src/scope.ts
```diff
@@ -130,9 +130,12 @@ export class Scope {
       const orphanIds = Array.from(
         resourceIds.filter((id) => !aliveIds.has(id)),
       );
-      const orphans = await Promise.all(
-        orphanIds.map(async (id) => (await this.state.get(id))!.output),
-      );
+      const orphans = (
+        await Promise.all(
+          orphanIds.map(async (id) => (await this.state.get(id))?.output),
```
ericclemmons, line 135: > Better to FATAL here since that was what caught the bug.

In which case, the `!` makes it seem safe whereas you'd probably prefer something like:

```suggestion
          orphanIds.map(async (id) => {
            const orphan = await this.state.get(id);
            
            if (!oprhan) {
              throw new Error(`Unexpected error fetching orphan ${id}`);
            }
          
            return orgphan.output;
          }),
```

ericclemmons:
Closing since `0.15.4` seems to work?  (But I may have also made changes that worked around it too?)

alchemy/src/scope.ts
sam-goodwin: Fair point! That would be more informative.

PR #152, opened 2025-05-07
fix(cloudflare): do not apply new_class_migration to a Workflow

sam-goodwin:
Workflows should not have class migrations


nickbalestra:
Thanks for this!

PR #161, opened 2025-05-10
feat(aws): support lambda layers in Function

blytheaw:
Expose layers option for Lambda functions

### DIFF
--- a/alchemy/src/aws/function.ts
+++ b/alchemy/src/aws/function.ts
@@ -133,6 +133,11 @@ export interface FunctionProps {
       maxAge?: number;
     };
   };
+
+  /**
+   * Lambda layers for the function. Use the layer ARN
+   */
+  layers?: string[];
 }
 
 /**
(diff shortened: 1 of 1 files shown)


alchemy/src/aws/function.ts
```diff
@@ -133,6 +133,11 @@ export interface FunctionProps {
       maxAge?: number;
     };
   };
+
+  /**
+   * Lambda layers for the function. Use the layer ARN
+   */
+  layers?: string[];
```
sam-goodwin, line 140: Should we add some tests? Is the idea to just add the minimum to support a Layer ARN for now, rather than code to also build layers? I know Pulumi has a way to construct them from assets, but that would be a larger task: https://www.pulumi.com/registry/packages/aws/api-docs/lambda/layerversion/

THis seems ok for now, but we should also add a `"@example"` to the `Function` docse

PR #164, opened 2025-05-11
Add UpstashRedis to alchemy/upstash

ericclemmons:
50% created via https://alchemy.run/docs/guides/custom-resources.html & https://upstash.com/docs/devops/developer-api/redis/create_database_global.

50% hand-crafted & tested in my local project.

I couldn't run these tests myself because
Screenshot 2025-05-10 at 11 41 12 PM
:


alchemy/src/upstash/redis.ts
```diff
@@ -0,0 +1,402 @@
...
+
+  /**
+   * Type of the database in terms of pricing model
+   */
+  databaseType: string;
+
+  /**
+   * Region where database is hosted
+   */
+  region: "global";
```
sam-goodwin, line 81: Is there no other region?

alchemy/src/upstash/redis.ts
```diff
@@ -0,0 +1,402 @@
...
+ * });
+ */
+export const UpstashRedis = Resource(
+  "upstash::Redis",
+  async function (
+    this: Context<UpstashRedis>,
+    id: string,
+    props: UpstashRedisProps,
+  ): Promise<UpstashRedis> {
+    const api = new UpstashApi();
```
sam-goodwin: The api client should be configurable by props - right now this depends on the environment variable. We should accept them as optional Secrets too.

alchemy/src/upstash/redis.ts
ericclemmons: Nope. They deprecated the other options.

https://upstash.com/docs/devops/developer-api/redis/create_database_global

alchemy/src/upstash/redis.ts
ericclemmons: Resolved via 6b02ddcc790a2d514d19711c14d870a8f31d83ac & fab08f35927f012be3e58cc1d60a5e105a7f03f6.

(I didn't think that `email` is a secret...?)

sam-goodwin: Email is probably fine plain text

sam-goodwin (approved):

alchemy/src/upstash/api.ts
```diff
@@ -1,4 +1,4 @@
-import type { Secret } from "../secret";
+import type { Secret } from "../secret.js";
```
ericclemmons, line 1: Unrelated to this PR, but I'd like to understand this better.

There's been weirdness in the erasable-types + node discussion, extensionless-imports, etc.

Is there a doc or some guidance I can learn from?

alchemy/src/upstash/redis.ts
```diff
@@ -2,6 +2,8 @@ import { alchemy } from "../alchemy.js";
 import type { Context } from "../context.js";
 import { Resource } from "../resource.js";
 import type { Secret } from "../secret.js";
+import { UpstashApi } from "./api.js";
```
ericclemmons, line 5: This seems to be a common-enough abstraction to warrant some guidance in `.cursorrules` 👀

ericclemmons:
@sam-goodwin Looks great!  I'll remove my `bun link` and test this immediately once merged.

Nothing blocking this PR – just more learnings for me!

alchemy/src/upstash/redis.ts
sam-goodwin: I think it is in there. Probably not smart enough to remember to put it in its own file.

It's also something we can share across all implementations. The only thing that changes is the auth part

PR #177, opened 2025-05-13
fix: move core Resource properties to Symbols

sam-goodwin:
We currently store properties like `ID`, `Scope`, `Kind` on the Resource's Output which can conflict with user-defined properties.

Usually we don't conflict because most properties are lower-case, but this is not always true - e.g. for the Control API https://github.com/sam-goodwin/alchemy/pull/132

Instead, now that `serde.ts` can serialize symbols, we move these to unique symbols to ensure they are totally isolated.

State is auto-migrated to the new format:
image

### DIFF
--- a/alchemy/src/alchemy.ts
+++ b/alchemy/src/alchemy.ts
@@ -3,7 +3,14 @@ import path from "node:path";
 
 import { destroy, DestroyedSignal } from "./destroy.js";
 import { env } from "./env.js";
-import type { PendingResource } from "./resource.js";
+import {
+  ResourceFQN,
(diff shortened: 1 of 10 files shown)


ericclemmons:
Heads up – `ResourceID` still shows up in autocomplete, but not an actual property on the returned state:
Screenshot 2025-05-14 at 11 20 34 AM
Screenshot 2025-05-14 at 11 20 42 AM

sam-goodwin:
Hm, I don't understand @ericclemmons 

It shows up in auto-complete but then type errors?

PR #178, opened 2025-05-13
fix(cloudflare): use Worker name for Worker to Worker binding

sam-goodwin:
There was a bug where we were not using `Worker.name` for the Worker to Worker binding.


nickbalestra:
Thanks so much Sam for being super responsive on this

nickbalestra (approved):
Oh, that was a quick fix

PR #180, opened 2025-05-14
feat(os): memoize exec from file patterns

john-royal:
## Changes

1. Enhanced `ExecProps.memoize` to support file-based memoization:
   - Previously: `memoize?: boolean`
   - Now: `memoize?: boolean | { patterns: string[] }`
   - Patterns can be an array of file names or globs, relative to the given working directory if provided or `process.cwd()` otherwise (uses `glob` from `fs/promises`, compatible with Bun and Node.js >= 22)
2. Added tests verifying the new file pattern functionality:
   - Commands are memoized when files haven't changed
   - Commands are re-executed when files are modified

## Example

```typescript
const result = await Exec("build", {
  command: "npm run build",
  memoize: { patterns: ["src/**/*.ts", "package.json"] }
});
```

### DIFF
--- a/alchemy/src/os/exec.ts
+++ b/alchemy/src/os/exec.ts
@@ -301,27 +301,27 @@ export async function exec(
 }
 
 async function hashInputs(cwd: string, patterns: string[]) {
-  const hashes: { file: string; hash: string }[] = [];
+  const hashes = new Map<string, string>();
 
   await Promise.all(
     patterns.flatMap(async (pattern) => {
       const files = await Array.fromAsync(glob(pattern, { cwd }));
       return Promise.all(
         files.map(async (file: string) => {
-          const content = await readFile(join(cwd, file));
-          hashes.push({
-            file,
-            hash: createHash("sha256").update(content).digest("hex"),
-          });
+          const path = join(cwd, file);
+          const content = await readFile(path);
+          hashes.set(path, createHash("sha256").update(content).digest("hex"));
         }),
       );
     }),
   );
 
-  hashes.sort((a, b) => a.file.localeCompare(b.file));
+  const sortedHashes = Array.from(hashes.entries()).sort((a, b) =>
+    a[0].localeCompare(b[0]),
+  );
 
   const finalHash = createHash("sha256");
-  for (const { hash } of hashes) {
+  for (const [, hash] of sortedHashes) {
     finalHash.update(hash);
   }
   return finalHash.digest("hex");
--- a/alchemy/src/os/exec.ts
+++ b/alchemy/src/os/exec.ts
@@ -1,6 +1,5 @@
 import { spawn, type SpawnOptions } from "node:child_process";
 import { createHash } from "node:crypto";
-import { glob, readFile } from "node:fs/promises";
 import { join } from "node:path";
 import type { Context } from "../context.js";
 import { Resource } from "../resource.js";
@@ -301,6 +300,8 @@ export async function exec(
 }
 
 async function hashInputs(cwd: string, patterns: string[]) {
+  const { glob, readFile } = await import("node:fs/promises");
+
   const hashes = new Map<string, string>();
 
   await Promise.all(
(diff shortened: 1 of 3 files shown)


alchemy/src/os/exec.ts
```diff
@@ -68,6 +77,11 @@ export interface Exec extends Resource<"os::Exec">, ExecProps {
    * Whether the command has completed execution
    */
   completed: boolean;
+
+  /**
+   * Hash of the command inputs
+   */
+  hash?: string;
```
sam-goodwin, line 96: I don't think this is being used is it?

alchemy/src/os/exec.ts
```diff
@@ -242,3 +266,14 @@ export async function exec(
...
 }
+
+async function hashInputs(cwd: string, patterns: string[]) {
+  const hash = createHash("sha256");
+  for (const pattern of patterns) {
+    for await (const file of glob(pattern, { cwd })) {
+      const content = await readFile(join(cwd, file), "utf-8");
+      hash.update(content);
+    }
+  }
```
sam-goodwin: This looks sequential. Should we instead use `await Promise.all` to compute the hashes of all files in parallel, then order by path and compute a single hash from the remaining hashes? That would be faster and maybe more deterministic?

alchemy/src/os/exec.ts
john-royal: It is on line 184:

```typescript
 (typeof props.memoize === "boolean" ||
        (hash && this.output?.hash === hash))
```

sam-goodwin: Oh i see, this is for a user to pass a hash? Not sure how i would use it?

john-royal: It's not included as an arg, but it's part of the output so that way we can reference it to know if something changes.

First we compute it if patterns are provided and we use it to know when not to re-run the command:

```ts
    const hash =
      typeof props.memoize === "object"
        ? await hashInputs(props.cwd ?? process.cwd(), props.memoize.patterns)
        : undefined;

    if (
      this.phase === "update" &&
      props.memoize &&
      this.output?.command === props.command &&
      (typeof props.memoize === "boolean" ||
        (hash && this.output?.hash === hash))
    ) {
      // If memoize is enabled and the command hasn't changed, return the existing output
      return this.output;
    }
```

Then it's returned as part of the output:

```ts
    // Return the execution result
    return this({
      id,
      command: props.command,
      cwd: props.cwd,
      env: props.env,
      memoize: props.memoize,
      inheritStdio,
      exitCode,
      stdout,
      stderr,
      executedAt: Date.now(),
      completed: true,
      hash,
    });
```

I'm new to Alchemy so feel free to let me know if this is the wrong approach!

alchemy/src/os/exec.ts
john-royal: Good catch! In my experience computing sequentially seems deterministic (obviously raw `Promise.all` isn't, learned that the hard way 😅), but you're right that computing in parallel would be faster. Just pushed an update to do just that.

john-royal: Also side note: the type signature for `glob` ostensibly accepts an array or a string, but Node.js throws if you actually pass it an array (Bun doesn't seem to mind). That's why I processed each glob separately.

sam-goodwin: That's interesting, thanks for being mindful about runtime compatibility!

sam-goodwin: If the globs overlap, could the same file end up being in the files list multiple times? Should we make sure to de-dupe?

john-royal:
Speaking of runtime compatibility... investigating why this is happening now lol

alchemy/src/os/exec.ts
sam-goodwin: Oh I may have misunderstood, I though that `hash` property was on the `ExecProps` not the `Exec` output. 

You got it right.

PR #187, opened 2025-05-16
Add cloudflare workers analytics engine binding

ostenbom:
Adds the ability to bind to cloudflare worker analytics engine datasets

https://developers.cloudflare.com/analytics/analytics-engine/

Odd kind of binding, since there is no iteraction with cloudflare's API.

- New datasets get created when they get used
- Old datasets die when the data in them gets 3 months old
- No ability to delete data or datasets

### DIFF
--- a/alchemy/src/cloudflare/analytics-engine.ts
+++ b/alchemy/src/cloudflare/analytics-engine.ts
@@ -1,0 +1,35 @@
+import type { Context } from "../context.js";
+import { Resource } from "../resource.js";
+import type { CloudflareApiOptions } from "./api.js";
+
+/**
+ * Properties for creating an Analytics Engine binding
+ */
+export interface AnalyticsEngineProps extends CloudflareApiOptions {
+  /**
+   * The name of the dataset to bind to
+   */
+  dataset: string;
+}
+
+export interface AnalyticsEngine
+  extends Resource<"cloudflare::AnalyticsEngineBinding">,
+    AnalyticsEngineProps {
+  type: "analytics_engine";
+}
+
+export const AnalyticsEngine = Resource(
+  "cloudflare::AnalyticsEngine",
+  async function (
+    this: Context<AnalyticsEngine>,
+    id: string,
+    props: AnalyticsEngineProps,
+  ): Promise<AnalyticsEngine> {
+    // Analytics engine datasets are not created or deleted by the api
+    // they are created when used, and removed when the data becomes 3 months old
+    return this({
+      type: "analytics_engine",
+      dataset: props.dataset,
+    });
+  },
+);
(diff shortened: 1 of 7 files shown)


alchemy/src/cloudflare/analytics-engine.ts
```diff
@@ -0,0 +1,35 @@
...
+  dataset: string;
+}
+
+export interface AnalyticsEngine
+  extends Resource<"cloudflare::AnalyticsEngineBinding">,
+    AnalyticsEngineProps {
+  type: "analytics_engine";
+}
+
+export const AnalyticsEngine = Resource(
```
sam-goodwin: Can you add docs with examples like we have in the other resources? I find cursor does a really good job of this by the way.

sam-goodwin (commented):
Thanks for taking the initiative on this! Been meaning to take this service for a spin. 

Can you add some docs and a test to make sure it works? I am trying to have at least one test that invokes a worker with the binding (an e2e test).

alchemy/src/cloudflare/analytics-engine.ts
sam-goodwin: Maybe this doesn't warrant being a resource since there's no way to create, update or delete a dataset?

Maybe follow the same pattern as we do for `DurableObjectNamespace`, `Workflow` and `BrowserRendering` where it's just a class (data structure) that can be bound to.

We can also attach the type of data the table accepts and infer that in `bound.ts`

```ts
export type DatasetRecord = Record<string, any>;

export class Dataset<Data extends DatasetRecord = DatasetRecord> {
  public readonly type = "analytics_engine" as const;
  constructor(public readonly dataset: string) {}
}
```

Then, in `bound.ts`, map `Dataset<D>` to the worker runtime binding type.

ostenbom: Updated to be a class kind in [30e6c20](https://github.com/sam-goodwin/alchemy/pull/187/commits/30e6c20f937a31641f9462fcf4a90cb490176e00) - that was a good idea!

Wasn't sure how to add it to a test suite though - any pointers?

ostenbom: Managed an e2e test in [997579e](https://github.com/sam-goodwin/alchemy/pull/187/commits/997579e48a944b7d568f871a4734abe8ded9f786)

sam-goodwin: That looks good - it should follow the same convention as the other worker tests. Create a worker inline with the binding and `url: true`, then execute fetch against it and check it actually works.

ostenbom: 👍  I think that's what the test does now!

PR #194, opened 2025-05-19
fix(cloudflare): open up policy and scope types for AccountApiToken

sam-goodwin:


alchemy/src/cloudflare/account-api-token.ts
```diff
@@ -50,12 +50,15 @@ export interface TokenPolicy {
...
+  | "com.cloudflare.edge.r2.bucket"
+  | `com.cloudflare.edge.r2.bucket.${string}`
   | `com.cloudflare.api.account`
   | `com.cloudflare.api.account.${string}`
   // | `com.cloudflare.api.account.zone`
   | `com.cloudflare.api.account.zone.${string}`
   // | `com.cloudflare.api.user`
-  | `com.cloudflare.api.user.${string}`;
+  | `com.cloudflare.api.user.${string}`
+  | (string & {});
```
eastlondoner, line 61: Magic Runes spotted! `(string & {})` classic trick, educated Murzin on it

PR #195, opened 2025-05-19
Add Cloudflare AccountId resource

eastlondoner:
## Summary
- Implements a new AccountId resource that retrieves and provides Cloudflare account IDs for use with other Cloudflare resources
- Adds ability to programmatically access Cloudflare account IDs
- Includes test coverage ensuring the resource works correctly

## Test plan
- Run `bun test test/cloudflare/account-id.test.ts` to verify the AccountId resource correctly retrieves account IDs

🤖 Generated


sam-goodwin (commented):
Does it need to be a resource? Could just be a function.

sam-goodwin (approved):

PR #196, opened 2025-05-19
FIX for cwd/wrangler.props

nickbalestra:
- Add support for using cwd when running script
- Better check through all the possible options for props.wrangler when building filename and path for temporary wrangler file
- Added compatibility_flags for temporary wrangler if avail in the props, to avoid vite complaining when doing the build

### DIFF
--- a/alchemy/src/cloudflare/website.ts
+++ b/alchemy/src/cloudflare/website.ts
@@ -61,7 +61,7 @@ export type Website<B extends Bindings> = B extends { ASSETS: any }
 
 export async function Website<B extends Bindings>(
   id: string,
-  props: WebsiteProps<B>,
+  props: WebsiteProps<B>
 ): Promise<Website<B>> {
   if (props.bindings?.ASSETS) {
     throw new Error("ASSETS binding is reserved for internal use");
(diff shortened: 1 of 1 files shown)


alchemy/src/cloudflare/website.ts
```diff
@@ -70,20 +70,41 @@ export async function Website<B extends Bindings>(
...
-    const wranglerPath =
-      fileName && path.relative(cwd, path.join(cwd, fileName));
-    const wranglerMain =
-      typeof props.wrangler === "object"
-        ? (props.wrangler.main ?? props.main)
-        : props.main;
+
+    let fileName;
+    let wranglerPath;
+    let wranglerMain = path.join(cwd, props.main ?? "");
```
sam-goodwin: Is `?? ""` right? In what situation would it be empty?

alchemy/src/cloudflare/website.ts
```diff
@@ -70,20 +70,41 @@ export async function Website<B extends Bindings>(
...
+    // Building fileName and wranglerPath for all the different supporteded cases of props.wrangler
+    // Taking into consideration also cwd
+    if (typeof props.wrangler === "boolean") {
+      fileName = "wrangler.jsonc";
+      wranglerPath = cwd;
+    }
+    if (typeof props.wrangler === "string") {
+      if (path.basename(props.wrangler) === props.wrangler) {
+        fileName = props.wrangler;
+        wranglerPath = cwd;
```
sam-goodwin: What's this case achieving?

alchemy/src/cloudflare/website.ts
```diff
@@ -70,20 +70,41 @@ export async function Website<B extends Bindings>(
...
+      } else {
+        fileName = path.basename(props.wrangler);
+        wranglerPath = path.dirname(props.wrangler);
+      }
+    }
+    if (typeof props.wrangler === "object" && props.wrangler != null) {
+      if (props.wrangler.main != null) {
```
sam-goodwin: comparing with `null` is a red flag. Users don't usually provide `null`, should this be `undefined`?

sam-goodwin: Also, you should always use `===` or `!==`, not `==` or `!=`

sam-goodwin (commented):
Having a hard time understanding what the desired behavior is? Can you add some comments with examples to each of the examples to make it clearer?

Am also wondering if we should add tests just for the file resolution because there are so many cases?

nickbalestra:
Hey @sam-goodwin 
> Having a hard time understanding what the desired behavior is? Can you add some comments with examples to each of the examples to make it clearer? 

That was exactly my same questions when I reviewed the original code rotfl:
```
const fileName =
      typeof props.wrangler === "boolean"
        ? "wrangler.jsonc"
        : typeof props.wrangler === "string"
          ? props.wrangler
          : (props.wrangler?.path ?? "wrangler.jsonc");
    const wranglerPath =
      fileName && path.relative(cwd, path.join(cwd, fileName));
    const wranglerMain =
      typeof props.wrangler === "object"
        ? (props.wrangler.main ?? props.main)
        : props.main;
 ```
 What are the original desired supported usecases you had in mind? there are no example, docs or tests for this. The code is really hard to read with nested ternaries plus shortcircuiting expressions nested within it lol. 
 
 Currently there are few issues in here:
 - props.cwd not used to exec props.command
 - compatibility_flags not being added to temp wrangler can cause vite to complain 
 - path of wrangler hard o set for situations where you not using the happy path. I wasn't able to find a way with current implementation to set specific output path for wrangler and also specify its main that was working for my setup, it always ended up with a wrangler.json written in the wrong place lol

alchemy/src/cloudflare/website.ts
nickbalestra: props.main is optional

alchemy/src/cloudflare/website.ts
nickbalestra: this check that props.wrangler passed is just the filename with no path: ie 'wrangler.json', in that case that is the fileName, and its path should be cwd

nickbalestra:
I think that matter as a first PR would be to get merged two things:
- props.cwd not used to exec props.command
- compatibility_flags not being added to temp wrangler can cause vite to complain

let me rebase it so that we can get those in

nickbalestra:
@sam-goodwin I've updated and reduced scope of this pr, this should be now a pretty quick merge, if we want/need we can open a dedicated pr/issue with a more complete story about path, but this should be good for now

PR #209, opened 2025-05-23
feat: add dispatch namespace support to worker

xar:
Add support for deploying user workers https://developers.cloudflare.com/cloudflare-for-platforms/workers-for-platforms/get-started/user-workers/


sam-goodwin:
Thanks for putting this together!

Perhaps we should add a test for deploying and invoking a dispatch namespace worker. 

Any constraints to be aware of? Like unsupported bindings?

xar:
I have added test with skipIf as I'm not sure If you / other contributors have access to workers for platforms. Let me please know if it should be done in different way.

As per documentation every binding type should be supported :) 
https://developers.cloudflare.com/cloudflare-for-platforms/workers-for-platforms/get-started/user-workers/#bindings

I'm gonna to use this now pretty extensively so we will see if there will be any issues

alchemy/test/cloudflare/worker.test.ts
```diff
@@ -2113,4 +2113,47 @@ describe("Worker Resource", () => {
       await assertWorkerDoesNotExist(workerName);
     }
   }, 60000); // Increase timeout for Worker operations
+
+  test.skipIf(true)("create, update, and delete worker (ESM format) in dispatch namespace", async (scope) => {
```
sam-goodwin: let's skipIf `!process.env.DISABLE_WFP_TEST` so a contributor can opt-in. I'll set it for CI since it's important to test this.

sam-goodwin: I've subscribed to WFP in alchemy account so we can enable it

In retrospect, I think it should be opt-out instead of opt-in. Contributors tend to just run their tests in isolation and then let CI run broad tests.

xar:
I have added `!!process.env.DISABLE_WFP_TEST`, but maybe I should also add support for creating actual dispatch namespace and dynamic worker. https://developers.cloudflare.com/cloudflare-for-platforms/workers-for-platforms/get-started/configuration/#4-create-a-dynamic-dispatch-worker  to have it fully complete

wdyt?

sam-goodwin:
> I have added test with skipIf as I'm not sure If you / other contributors have access to workers for platforms

Other contributors don't have to run the WFP tests. They can run individual tests and then submit PR to CI. I have access to WFP and so does CI.

> maybe I should also add support for creating actual dispatch namespace and dynamic worker. https://developers.cloudflare.com/cloudflare-for-platforms/workers-for-platforms/get-started/configuration/#4-create-a-dynamic-dispatch-worker to have it fully complete

I see, I was mistakenly assuming that just deploying a worker to WFP did this.

Looks like it also needs a Resource to be generated from their CRUD API: https://developers.cloudflare.com/api/resources/workers_for_platforms/subresources/dispatch/subresources/namespaces/methods/create/

Then, support binding it to a Worker and specifying it as the Dispatch Namespace a Worker belongs to:

```ts
const myNamespace = await DispatchNamespace("my-namespace", {
  namespace: "my-namespace", // <- default to the id ^
});

// this is the Worker that will dispatch requests
const dispatcher = await Worker("my-dispatcher", {
  bindings: {
    NAMESPACE: myNamespace
  }
});

// this is a worker in the namespace
const myWorker = await Worker("my-worker", {
  // does adding this implicitly mean the Worker is a WFP Worker?
  dispatchNamespace: myNamespace
});
```

sam-goodwin:
Closing in favor of https://github.com/sam-goodwin/alchemy/pull/242

PR #217, opened 2025-05-28
do-state-store

FutureExcited:
gotta switch to entrypoint worker deployment instead of a placeholder. adding tests too


FutureExcited:
* **#217** <a href="https://app.graphite.dev/github/pr/sam-goodwin/alchemy/217?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sam-goodwin/alchemy/217?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

alchemy/package.json
```diff
@@ -62,6 +62,7 @@
     "aws4fetch": "^1.0.20",
     "cloudflare": "^4.0.0",
     "diff": "^7.0.0",
+    "dofs": "^0.0.1-rc.2",
```
sam-goodwin: hellz yeah

sam-goodwin: not sure if `^` will work here for a `rc`. The dofs creator should probably drop that. During `0.x` it's fine to just treat it as `0.major.minor`

alchemy/src/cloudflare/do-state-store/worker.ts
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  async fetch(request: Request): Promise<Response> {
+    try {
+      const url = new URL(request.url);
+      const operation = request.headers.get("X-Operation");
+      
+      if (!operation) {
+        return new Response("Missing X-Operation header", { status: 400 });
+      }
```
sam-goodwin: what is `X-Operation` and why are we using a header instead of url/query string?

alchemy/src/cloudflare/do-state-store/worker.ts
```diff
@@ -0,0 +1,231 @@
...
+  private fs: Fs;
+
+  constructor(state: DurableObjectState, env: any) {
+    super(state, env);
+    this.fs = new Fs(state, env, {
+      chunkSize: 256 * 1024, // 256KB chunks for good performance
+    });
+  }
+
+  async fetch(request: Request): Promise<Response> {
```
sam-goodwin: we will need to validate access. Am thinking that we should use a bearer token that is configured.

```ts
await alchemy("my-app", {
  stateStore: (scope) => new DOFSStateStore({
    apiKey: await alchemy.secret.env.DOFS_API_KEY
  })
})
```

Then, the `DOFSStateStore` will auto-create the Worker and add this as a Secret binding.

sam-goodwin (changes requested):
Looking great, thanks for putting this together. I think we'll want to add some tests and also make sure to enable auto-creation and auth.

FutureExcited:
> Looking great, thanks for putting this together. I think we'll want to add some tests and also make sure to enable auto-creation and auth.

yeah boiii working on auto-creation rn. thanks for all the feedback, all is clear

alchemy/src/cloudflare/do-state-store/dofs-worker.ts
```diff
@@ -0,0 +1,308 @@
+/**
+ * DOFS State Store Worker Script
+ * 
+ * This is the Cloudflare Worker script that gets auto-deployed
+ * to provide the DOFS (Durable Object File System) functionality.
+ */
+
+export const DOFS_WORKER_SCRIPT = `
+import { DurableObject } from "cloudflare:workers";
+import { Fs } from "dofs";
```
sam-goodwin: why encode as a string? you can use a .js file?

alchemy/src/cloudflare/do-state-store/example.ts
```diff
@@ -0,0 +1,170 @@
...
+
+async function main() {
+  console.log("🚀 Starting DOFS State Store Auto-Deployment Example");
+  console.log("=".repeat(60));
+
+  try {
+    console.log("🔧 Initializing Alchemy with DOFS state store...");
+    
+    // Create Alchemy app with auto-deploying DOFS state store
+    const app = await alchemy("dofs-test-app", {
```
sam-goodwin: use top level awair instead of wrapping in async function main

alchemy/test/cloudflare/do-state-store.test.ts
```diff
@@ -0,0 +1,251 @@
...
+import { DOFSStateStore } from "../../src/cloudflare/do-state-store/index.js";
+import { destroy } from "../../src/destroy.js";
+import { BRANCH_PREFIX } from "../util.js";
+
+// must import this or else alchemy.test won't exist
+import "../../src/test/bun.js";
+
+const test = alchemy.test(import.meta, {
+  prefix: BRANCH_PREFIX,
+});
```
sam-goodwin, line 12: A great way to test this would be to use it.

```ts
const test = alchemy.test(import.meta, {
  prefix: BRANCH_PREFIX,
  stateStore: (scope) => new DOFSStateStore(scope, {
      autoDeploy: false,
      url: "https://test-worker.example.workers.dev",
      workerName: `${BRANCH_PREFIX}-test-worker`,
      basePath: "/test-alchemy",
    })
});
```

alchemy/test/cloudflare/do-state-store.test.ts
```diff
@@ -0,0 +1,251 @@
...
+
+describe("DOFS State Store", () => {
+  test("create and configure state store", async (scope) => {
+    const stateStore = new DOFSStateStore(scope, {
+      autoDeploy: false,
+      url: "https://test-worker.example.workers.dev",
```
sam-goodwin, line 18: i don't quite understand this url? Where is this hosted?

alchemy/src/cloudflare/do-state-store/store.ts
```diff
@@ -0,0 +1,507 @@
...
+        return;
+      }
+
+      // Auto-deploy worker if enabled and no URL provided
+      if (this.autoDeploy) {
+        try {
+          console.log("🔧 Auto-deployment enabled, setting up API client...");
+          
+          // Create Cloudflare API client
+          this.api = await createCloudflareApi();
```
sam-goodwin, line 122: this is not using the optional api key passed in by the options

alchemy/src/cloudflare/do-state-store/store.ts
```diff
@@ -0,0 +1,507 @@
...
+      this.worker = await this.scope.run(async () => {
+        return Worker(this.workerName, {
+          name: this.workerName,
+          entrypoint: path.join(__dirname, "dofs-worker.ts"), // Point to the actual TypeScript file
+          bindings,
+          url: this.workerUrl,
+          compatibilityDate: "2024-09-23", // Required for nodejs_compat
+          compatibilityFlags: ["nodejs_compat"], // Required for dofs
+        });
+      });
```
sam-goodwin, line 249: We should not use alchemy for this since it is stateful.

We should follow the pattern that we use in R2 Rest State store where we create a Worker with a known name. For R2, we create a bucket with name `alchemy-state`.

We should create a Worker named `alchemy-state`

alchemy/src/cloudflare/do-state-store/store.ts
```diff
@@ -0,0 +1,507 @@
...
+    // Optionally clean up the directory
+    try {
+      await this.fetch("/rmdir", {
+        method: "POST",
+        body: JSON.stringify({ path: this.fullPath, recursive: true }),
+      });
+    } catch (error) {
+      // Ignore cleanup errors
+    }
+  }
```
sam-goodwin, line 271: We should not delete the data on deinit

sam-goodwin:
I tried using the DOFSStateStore in place of R2RestStateStore in our examples and it just hung:

```
🚀 Starting DOFS state store initialization...
🔧 Auto-deployment enabled, setting up API client...
✅ API client created for account: 7bc8e300d3a2fc8be08a9215c36be2d6
🔍 Checking if worker 'alchemy-dofs-state-store' already exists...
🏗️  Worker not found, deploying new worker...
🏗️  Deploying DOFS worker with bundling...
⚠️  Init already in progress, waiting...
```

I think we should simplify. Don't use alchemy. Just call the API directly to deploy a Worker with the known name if it doesn't already exist.

You can use the DownloadWOrker API to check if the worker exists by name: https://developers.cloudflare.com/api/resources/workers/subresources/scripts/methods/get/

alchemy/src/cloudflare/index.ts
```diff
@@ -38,3 +38,4 @@ export * from "./worker.ts";
 export { Workflow } from "./workflow.ts";
 export * from "./wrangler.json.ts";
 export * from "./zone.ts";
+
```
sam-goodwin, line 41: we should export DOFSStateStore

alchemy/src/cloudflare/do-state-store/example.ts
```diff
@@ -0,0 +1,146 @@
...
+console.log("=".repeat(60));
+
+try {
+  console.log("🔧 Initializing Alchemy with DOFS state store...");
+  
+  // Create Alchemy app with auto-deploying DOFS state store
+  const app = await alchemy("dofs-test-app", {
+    stage: "dev", 
+    phase: "up",
+
```
sam-goodwin, line 25: Ultumately the goal is to move all the tests to DOFSStateStore, so we don't need this example. Use the current examples and move them to it. Good way to test actual usage too.

john-royal:
Replacing with #266

PR #224, opened 2025-05-29
chore: this should be only `dist` and not `dist/client` for vite

Rohithgilla12:
https://x.com/gillarohith/status/1928030984852652068


sam-goodwin:
Hmm, I've changed this a few times. It must be different depending on config. Will have to look 

Got a similar discussion going on with react router.

https://github.com/sam-goodwin/alchemy/pull/225

PR #226, opened 2025-05-29
chore: move to ts imports for less confusing imports

BlankParticle:
instead of using `.js` imports which is confusing, we can use `.ts` imports which is more explicit.
tsc takes care of rewriting the imports to `.js` when publishing to npm.

> Merge #227 first to test out the npm package from this PR without having to merge this PR first


alchemy/tsconfig.json
```diff
@@ -1,16 +1,17 @@
...
   "extends": "../tsconfig.base.json",
-  "include": ["src/**/*"],
+  "include": ["src/**/*.ts"],
   "compilerOptions": {
     "composite": true,
     "noEmit": false,
     "outDir": "./lib",
     "rootDir": "./src",
     "sourceMap": true,
-    "declarationMap": true,
```
sam-goodwin: Why not declaration map?

alchemy/tsconfig.json
```diff
@@ -1,16 +1,17 @@
...
     "module": "Preserve",
     "moduleResolution": "Bundler",
     "target": "ESNext",
-    "allowJs": true
+    "allowJs": true,
+    "allowImportingTsExtensions": true,
+    "rewriteRelativeImportExtensions": true
```
sam-goodwin, line 16: I think `biome.json` will also need to be updated since I was using that to check for correct suffix. It's configured as `.js`

alchemy/tsconfig.json
BlankParticle: its for generating sourcemaps for d.ts files,
i think I disabled it for some testing and forgot to add back, do we need it?

sam-goodwin: We might not be making use of it properly right now. I don't know if we need it necessarily, but the outcome we want is:
1. users of `alchemy` get stack traces that point to the original `.ts`
2. jump to definition should bring the user to the `.ts` file (this may require packaging the `.ts` source in NPM too)

sam-goodwin:
Tested the new package and it worked

PR #228, opened 2025-05-29
feat: resource for orange-js

zebp:
Adds a resource for the [OrangeJS framework](https://github.com/orange-framework/orange-js) which is a small abstraction over `Website` and an example based off the `create-orange` template.

## Sample

```ts
export const website = await Orange("orange-website");
```


examples/cloudflare-orange/app/env.d.ts
```diff
@@ -0,0 +1,16 @@
+/// <reference types="./env.d.ts" />
```
sam-goodwin: i've noticed that `.d.ts` doesn't type check and have been un-winding this. Better to rename to `.ts`

I am also not sure if the `/// reference` helps vs putting this in the tsconfig.json
```ts
{
  "compilerOptions": {
    "types": ["./app/env.ts"]
  }
}
```

examples/cloudflare-orange/app/env.d.ts
```diff
@@ -0,0 +1,16 @@
...
+
+import type { website } from "../alchemy.run.js";
+
+export type CloudFlareEnv = typeof website.Env;
+
+// Necessary for the `env` property of loaders/actions to be typed
+declare global {
+  export type Env = CloudFlareEnv;
+}
```
sam-goodwin: Yeah, i like this pattern too. Make it global.

sam-goodwin (changes requested):
This is so great.

Do you want to add a getting started guide to alchemy-web/guides/orange.md

You can link out to your project website and walk people through setting up a new Orange project with Alchemy

examples/cloudflare-orange/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "version": "1.0.0",
+  "type": "module",
+  "main": "index.js",
+  "scripts": {
+    "dev": "vite",
+    "build": "vite build",
+    "preview": "vite build && wrangler dev",
+    "deploy": "bun run --env-file ../../.env ./alchemy.run.ts",
+    "destroy": "bun run --env-file ../../.env ./alchemy.run.ts --destroy",
+    "types": "orange types --no-wrangler"
```
sam-goodwin, line 12: What does this do?

examples/cloudflare-orange/app/env.d.ts
zebp: As long as the file is checked in some way (via types or include) it should work for declaring the types, I did this slightly different for the Orange example just to keep the tsconfig as similar to the Orange template but it still works the same.

examples/cloudflare-orange/package.json
zebp: `orange types --no-wrangler` generates types for each route so you can have type safe parameters and such.

sam-goodwin:
I resolved the merge conflict and rebased

sam-goodwin:
Looks like a credential problem in the tests, will fix

alchemy/src/cloudflare/orange.ts
```diff
@@ -0,0 +1,54 @@
...
+  return Website(id, {
+    ...props,
+    command: props?.command ?? "bun run build",
+    // The bundler main and the Wrangler main must differ for the Cloudflare vite plugin
+    // to bundle the application correctly, and alchemy must run our bundler on the Vite
+    // output since we don't resolve vite imports.
+    wrangler: props?.wrangler ?? {
+      main: "app/entry.server.ts",
+    },
+    main: props?.main ?? "dist/ssr/entry.server.js",
```
sam-goodwin, line 57: Should we also set noBundle: true and have it upload all files in dist/ssr as es modules instead of re-bundling?

examples/cloudflare-nuxt-pipeline/env.d.ts
```diff
@@ -3,6 +3,10 @@ import type { worker } from "./alchemy.run.js";
 
 export type WorkerEnv = typeof worker.Env;
```
sam-goodwin, line 4: Just above this line it should be import alchemy.run.ts instead of .js

examples/cloudflare-orange/alchemy.run.ts
```diff
@@ -0,0 +1,27 @@
...
+  phase: process.argv.includes("--destroy") ? "destroy" : "up",
+  stateStore:
+    process.env.ALCHEMY_STATE_STORE === "cloudflare"
+      ? (scope) => new R2RestStateStore(scope)
+      : undefined,
+});
+
+export const website = await Orange(
+  `cloudflare-orange-website${BRANCH_PREFIX}`,
+);
```
sam-goodwin: Have been preferring the following pattern:

```ts
await Orange("website", {
  name: `cloudflare-orange-website${BRANCH_PREFIX}`
})
```

examples/cloudflare-nuxt-pipeline/env.d.ts
zebp: Fixed in the Orange env.ts, do you want me to apply this to all of the envs and similarly switch them to env.ts like mentioned earlier?

sam-goodwin: There's a bit of a bind with `env.ts` vs `env.d.ts`. 

`env.d.ts` can be included in tsconfig's `types`, `env.ts` cannot 

`env.ts` is type-checked, `env.d.ts` is not, so it is easy to make mistakes.

Not sure what to say on that front, but we should make sure to use `.ts` in imports everywhere.

PR #229, opened 2025-05-30
chore: consistent workspace:* and "ship typescript"

sam-goodwin:


package.json
```diff
@@ -20,13 +20,7 @@
...
     "alchemy-web",
-    "examples/aws-app",
-    "examples/cloudflare-nuxt-pipeline",
-    "examples/cloudflare-react-router",
-    "examples/cloudflare-redwood",
-    "examples/cloudflare-vite",
-    "examples/cloudflare-worker-bootstrap",
-    "examples/cloudflare-worker"
+    "examples/*"
   ],
```
sam-goodwin, line 24: I should omit tanstack from this since it doesn't work in monorepo right now, or i should mayeb try pnpm instead

PR #230, opened 2025-05-30
feat(cloudflare): dev mode

sam-goodwin:


sam-goodwin:
@claude can you create a fork of this branch and submit a pull request that merges into sam/miniflare base branch that:
1. uses esbuild "context" to watch the bundle (when using esbuild to bundle a Worker with `entrypoint: string` defined and have it trigger an auto-update of the miniflare worker. This is for dev mode. It's basically hot reloading of the worker.
2. use the vite programmable API to launch the vite projects when scope.dev is true in Website? The goal is to use vite's API to start a server that hot reloads and shares the same miniflare state as the singleton miniflare instance in alchemy

PR #253, opened 2025-05-31
feat: add scriptName to workflow bindings

jonastemplestein:
Drive-by PR from codex - we are blocked on this so would be awesome if you could add and check it actually works <3


sam-goodwin:
Closing in favor of https://github.com/sam-goodwin/alchemy/pull/254

PR #264, opened 2025-06-02
fix: include types.d.ts file in lib

tvanhens:
This isn't a complete fix yet but it at least improves the current experience. The crux of the issue is that `.d.ts` files aren't used by the compiler to produce the library outputs. As a result the generated `types.d.ts` file for CloudControl is not in the `lib` directory which breaks the typing and `any` gets used as a fallback.

This uses `cpx` to copy the file into the lib outputs so that the types will work as expected. However, this is not an ideal fix as there are still typing errors which fail silently. If you change the file from `types.d.ts` to `types.ts` you see these errors explicitly but doing so breaks compilation since these errors need to be resolved to compile.

I will follow up with a more complete fix.

### DIFF
--- a/alchemy/package.json
+++ b/alchemy/package.json
@@ -11,6 +11,7 @@
   },
   "scripts": {
     "build": "rm -rf ./*.tsbuildinfo && rm -rf ./lib && tsc -b",
+    "postbuild": "cpx 'src/**/*/types.d.ts' lib",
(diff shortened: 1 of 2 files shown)


sam-goodwin:
Should we just rename it from `.d.ts` to `.ts`?

tvanhens:
> Should we just rename it from `.d.ts` to `.ts`?

Yeah working on a followup for this along with fixes for the errors that surface when you do that.

PR #266, opened 2025-06-03
feat(cloudflare): do state store

john-royal:
Probably needs some cleanup before we merge but looks like tests are passing!


alchemy/src/cloudflare/do-state-store/worker.ts
```diff
@@ -0,0 +1,229 @@
...
+      return new Response("Unauthorized", { status: 401 });
+    }
+    if (request.url.endsWith("/status")) {
+      return new Response("OK");
+    }
+    if (request.method !== "POST") {
+      return new Response("Method not allowed", { status: 405 });
+    }
+    const id = env.DOFS_STATE_STORE.idFromName("default");
+    const stub = env.DOFS_STATE_STORE.get(id);
```
sam-goodwin: This should be determined by app-name/stage

alchemy/test/cloudflare/queue-consumer.test.ts
```diff
@@ -12,6 +13,7 @@ import "../../src/test/bun.js";
 
 const test = alchemy.test(import.meta, {
   prefix: BRANCH_PREFIX,
+  stateStore: (scope) => new DOStateStore(scope),
```
sam-goodwin: you can change this in `alchemy/test/bun.ts` instead of doing it for every test

john-royal: Noob mistake my bad!

alchemy/src/cloudflare/do-state-store/store.ts
```diff
@@ -0,0 +1,293 @@
...
+
+  private async createClient(): Promise<DOFetchClient> {
+    const workerName = this.options.worker?.name ?? "alchemy-state";
+    const token = this.options.worker?.token ?? process.env.ALCHEMY_STATE_TOKEN;
+    if (!token) {
+      throw new Error(
+        "DOStateStore requires an API key. Please set options.worker.token or the ALCHEMY_STATE_TOKEN environment variable.",
+      );
+    }
+    if (this.options.worker?.url) {
```
sam-goodwin: What do we do if it is not defined?

alchemy/src/cloudflare/do-state-store/store.ts
```diff
@@ -0,0 +1,293 @@
...
+      getWorkerScriptMetadata(api, workerName),
+    ]);
+    const client = new DOFetchClient(
+      `https://${workerName}.${subdomain}.workers.dev`,
+      token,
+      this.prefix,
+    );
+    if (worker && !this.options.worker?.force) {
+      return client;
+    }
```
sam-goodwin: what does force mean?

alchemy/src/cloudflare/do-state-store/worker.ts
sam-goodwin: Or at least I was thinking as such. If we have all apps point at a single durable object, we're limited to 100 TPS across all apps.

But it would greatly simplify the problem of building a dashboard across all apps/stages.

That could be achieved with an `app` DO that maintains a list of all stages though. Food for thought.

alchemy/src/cloudflare/do-state-store/store.ts
john-royal: Force replace the worker if it already exists, mainly for debugging. I'll add docs+comments to make this clearer.

I'm also thinking we can use the tag for versioning so if we want to update the worker, we can do that. In those cases, you wouldn't need to force.

alchemy/src/cloudflare/do-state-store/store.ts
john-royal: If the URL isn't defined we fill it in automatically. This is just so you can force it to use an existing worker at a specific rather than creating one. Open to feedback on this API design or lack thereof!

alchemy/src/cloudflare/do-state-store/store.ts
sam-goodwin: Makes a lot of sense. Maybe use the version of alchemy or a hash of the bundle. Store the bundle version as a tag on the worker and use that to determine if it should be updated.

alchemy/src/cloudflare/do-state-store/worker.ts
john-royal: I think it makes sense to shard by app/stage. The app DO is a good idea but we can add that in later.

john-royal:
Implemented the changes you requested. Feel free to take look and we can re-run the tests when CF is in the mood!

PR #267, opened 2025-06-03
Export analytics engine dataset binding from cloudflare

ostenbom:
.... oops, forgot this in the last PR.


ostenbom:
The new PR packages are 👌  - could properly test it in my project easily!

sam-goodwin:
> The new PR packages are 👌 - could properly test it in my project easily!

Yeah that is a super nice feature!

PR #269, opened 2025-06-03
fix(cloudflare): implement source maps support equivalent to wrangler's upload_source_maps

### LINKED ISSUE #147: Equivalent of wrangler's `upload_source_maps`
Look into what `upload_source_maps` and determine if we need to do anything special

saint0x:
Fixes #147

Switch to inline source maps to avoid Cloudflare API 5xx errors. External source maps require 3-part multipart uploads which Workers API rejects.

Changes:
- sourcemap: 'external' → 'inline' in esbuild  
- Remove external source map upload logic
- Add source map testing


alchemy/src/cloudflare/bundle/bundle-worker.ts
```diff
@@ -73,16 +106,21 @@ export async function bundleWorkerScript<B extends Bindings>(
   }
 
   try {
-    const bundle = await Bundle("bundle", {
+    const entryPointBaseName = path.basename(main, path.extname(main));
+    const outfileName = `${entryPointBaseName}.js`;
```
sam-goodwin, line 171: Does this mean we always write a file now? Is it safe to write this file there? We were previously in a state of in-memory only, can we keep that?

alchemy/src/cloudflare/bundle/bundle-worker.ts
```diff
@@ -73,16 +106,21 @@ export async function bundleWorkerScript<B extends Bindings>(
...
       entryPoint: main,
-      format: props.format === "cjs" ? "cjs" : "esm", // Use the specified format or default to ESM
+      outfile: outfileName,
+      format: props.format === "cjs" ? "cjs" : "esm",
       target: "esnext",
       platform: "node",
-      minify: false,
+      minify: props.bundle?.minify ?? false,
       ...(props.bundle || {}),
+      sourcemap: props.sourceMaps ? 'inline' : undefined,
```
sam-goodwin: looks like the user could already pass `bundle: { sourcemap: "inline" }`? Should we support that instead of adding an additonal prop?

alchemy/src/cloudflare/bundle/bundle-worker.ts
```diff
@@ -105,14 +143,28 @@ export async function bundleWorkerScript<B extends Bindings>(
...
     }
-    if (bundle.path) {
-      return await fs.readFile(bundle.path, "utf-8");
+
+    let scriptName = "_worker.js";
+    if (bundleResult.path) {
+      scriptName = path.basename(bundleResult.path);
+      if (scriptName.endsWith(".map")) {
+        scriptName = scriptName.slice(0, -4);
+      }
```
sam-goodwin: Not sure what this is for?

alchemy/src/cloudflare/worker-metadata.ts
```diff
@@ -2,14 +2,14 @@ import path from "node:path";
...
 import { slugify } from "../util/slugify.ts";
 import {
-  Self,
-  type Bindings,
-  type WorkerBindingDurableObjectNamespace,
-  type WorkerBindingSpec,
+    Self,
+    type Bindings,
+    type WorkerBindingDurableObjectNamespace,
+    type WorkerBindingSpec,
```
sam-goodwin: make sure to format code with `bun check --fix` from the root and use biome in vscode/your IDE

alchemy/src/cloudflare/bundle/bundle-worker.ts
```diff
@@ -17,13 +17,46 @@ export type NoBundleResult = {
...
+     */
+    content: string;
+  };
+}
+
+/**
+ * The output of the worker script bundling process.
+ * Can be a single bundled script with an optional source map, or multiple files if noBundle is true.
+ */
+export type WorkerScriptOutput = NoBundleResult | BundledWorkerScript;
```
sam-goodwin: I think we also need source maps for the NobundleResult

"no bundle" means upload multiple files found in the `path.basename(entrypoint)`. Maybe we need to include `.map` source files when scanning that directory. Depending on how they produced the bundle originally (e.g. `vite build`), inlines should work out of the box.

sam-goodwin: Maybe thinking in terms of producing a directory for the worker with all the files (source maps and all) and then pushing everything through the NoBundle code path would be a nice way to go about it?

alchemy/src/cloudflare/bundle/bundle-worker.ts
```diff
@@ -17,13 +17,46 @@ export type NoBundleResult = {
   [fileName: string]: Buffer;
 };
 
+/**
+ * Represents the output of a bundled worker script, including optional source map.
+ */
+export interface BundledWorkerScript {
```
sam-goodwin, line 60: I wonder if we can consolidate all "bundles" into a single interface.

alchemy/src/cloudflare/worker.ts
```diff
@@ -1008,37 +1017,53 @@ export async function deleteWorker<B extends Bindings>(
...
-      const scriptName =
-        scriptMetadata.main_module ?? scriptMetadata.body_part!;
-
       // Create FormData for the upload
       const formData = new FormData();
 
-      function addFile(fileName: string, content: Buffer | string) {
-        const contentType = getContentType(fileName) ?? "application/null";
+      function addFile(fileName: string, content: Buffer | string, typeOverride?: string) {
+        const contentType = typeOverride ?? getContentType(fileName) ?? "application/octet-stream";
```
sam-goodwin: Should use `application/null` as the default. Cloudflare barfs if you don't

alchemy/src/cloudflare/worker.ts
```diff
@@ -1008,37 +1017,53 @@ export async function deleteWorker<B extends Bindings>(
...
+        }
+
+        addFile(scriptBundleInput.scriptName, scriptBundleInput.scriptContent);
+
+        // Note: Source maps are now inline (embedded in the JavaScript content)
+        // No need to upload them as separate files
+      } else { // It's NoBundleResult (multiple files, typically from noBundle=true)
+        // Original logic for NoBundleResult
+        const mainEntryFile = scriptMetadata.main_module ?? scriptMetadata.body_part!;
+        for (const [fileName, content] of Object.entries(scriptBundleInput)) {
```
sam-goodwin: This seems really complicated. I'd expect it to remain mostly the same with now support for a `.map` file

alchemy/src/esbuild/bundle.ts
```diff
@@ -91,8 +89,35 @@ export interface Bundle<P extends BundleProps = BundleProps>
...
+    content: string;
+  };
 }
 
+/**
+ * Output returned after bundle creation/update
+ */
+export interface Bundle<P extends BundleProps = BundleProps>
+  extends Resource<"esbuild::Bundle">,
+    Omit<P, 'entryPoint' | 'outdir' | 'outfile' | 'format' | 'target' | 'minify' | 'sourcemap' | 'external' | 'platform'>, // Omit esbuild.BuildOptions to avoid conflict
```
sam-goodwin: This is excessive, why?

alchemy/src/esbuild/bundle.ts
```diff
@@ -136,70 +161,181 @@ export const Bundle = Resource(
...
+            mainBundlePath = path.resolve(outputPath);
+            mainBundleContent = await fs.readFile(mainBundlePath, "utf-8");
+            mainBundleHash = crypto.createHash("sha256").update(mainBundleContent).digest("hex");
+            // Check for an associated source map file
+            const potentialMapPath = mainBundlePath + ".map";
+            if (result.metafile.outputs[potentialMapPath.substring(path.resolve(props.absWorkingDir ?? process.cwd()).length +1 )] || result.metafile.outputs[potentialMapPath.substring(process.cwd().length +1)]) {
+               // Need to check both relative to absWorkingDir and cwd for robustness if metafile paths are inconsistent
+              const mapKey = Object.keys(result.metafile.outputs).find(p => path.resolve(props.absWorkingDir ?? process.cwd(), p) === potentialMapPath);
+              if (mapKey) {
+                sourceMapOutput = {
```
sam-goodwin: This looks a little out of control? Can we take a step back try and simplify? 

Source maps are either inline or separate files.

We need to support `bundle` and `noBundle` use-cases.

Maybe it's best to change bundling to generate a directory (or even a virtual in memory only if esbuild supports it) and then generalize the way we upload all of those files. 

Most of that is established already with `NoBundleResult`.

I am not sure what all this extra hashing is for

sam-goodwin (changes requested):

saint0x:
Updates: 
In-memory only (write: false)
Use existing bundle.sourcemap pattern
Remove script name manipulation
Fix formatting/content-types
Add noBundle source map support
Unify bundle interfaces
Simplify upload logic
Remove excessive Bundle interface complexity
Eliminate complex path/hash parsing

sam-goodwin:
Thank you so much for digging into this. I was able to achieve with a simpler change, so will be closing this for now.

PR #271, opened 2025-06-03
feat(cloudflare): add SvelteKit on Cloudflare Workers guide using Alchemy

acoyfellow:
Adds a full example for deploying SvelteKit apps to Cloudflare Workers with Alchemy


sam-goodwin:
Boom! And now we have SvelteKit 🔥

alchemy/src/cloudflare/sveltekit.ts
```diff
@@ -0,0 +1,66 @@
...
+export async function SvelteKit<B extends Bindings>(
+  id: string,
+  props?: Partial<SvelteKitProps<B>>,
+): Promise<SvelteKit<B>> {
+  return Website(id, {
+    ...props,
+    // Default build command for SvelteKit
+    command: props?.command ?? "bun run build",
+    // FIXED: Use the correct entry point that SvelteKit adapter generates
+    main: props?.main ?? "./.svelte-kit/cloudflare/_worker.js",
```
sam-goodwin: Does svelte work with wrangler at all? What's the local dev story? We've been having to have adifferent `Website.main` and `Website.wrangler.main` to deal with this weirdness

alchemy/src/cloudflare/sveltekit.ts
```diff
@@ -0,0 +1,66 @@
...
+    // FIXED: Use the cloudflare directory which contains all static assets
+    assets: props?.assets ?? "./.svelte-kit/cloudflare",
+    // SvelteKit with Cloudflare adapter needs nodejs_compat
+    compatibilityFlags: ["nodejs_compat", ...(props?.compatibilityFlags ?? [])],
+    // Set compatibility date for nodejs_compat (must be >= 2024-09-23)
+    compatibilityDate: props?.compatibilityDate ?? "2024-09-23",
```
sam-goodwin: I think you can safely leave this undefined - our default is much older and it's probably better to use a 2025 date.

maybe just validate it's `>=2024-09-03` if not `undefined`

alchemy/src/cloudflare/sveltekit.ts
```diff
@@ -0,0 +1,66 @@
...
+ *     SESSIONS: sessions
+ *   }
+ * });
+ *
+ * @example
+ * // Deploy with custom build command and assets directory
+ * const customSvelteApp = await SvelteKit("custom-svelte", {
+ *   command: "npm run build:cloudflare",
+ *   assets: "./static"
+ * });
```
sam-goodwin: Probably need an example of how to also deploy a worker with API server.

alchemy/src/cloudflare/sveltekit.ts
sam-goodwin: What if the user wants to provide a worker `./src/worker.ts` for the server. 

what creates that `./.svelte-kit/cloudflare/_worker.js`?

sam-goodwin: Try and use `path.join(".svelte-kit/cloudflare/_worker.js")`

examples/cloudflare-sveltekit/README.md
```diff
@@ -0,0 +1,85 @@
...
+   bun install
+   ```
+   This will install dependencies for the core library and all examples, including this one.
+
+2. **Cloudflare Credentials:** Ensure you have your Cloudflare API token and Account ID configured as environment variables:
+   ```bash
+   export CLOUDFLARE_API_TOKEN="your_api_token"
+   export CLOUDFLARE_ACCOUNT_ID="your_account_id"
+   ```
+   You can also place these in a `.env` file in the repository root.
```
sam-goodwin, line 26: Link to the Cloudflare auth page https://alchemy.run/docs/guides/cloudflare-auth.html

examples/cloudflare-sveltekit/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+In your SvelteKit application, access the Cloudflare resources via `platform.env`:
+
+```typescript
+// In a +page.server.ts file
+export async function load({ platform }) {
+  const kvData = await platform?.env?.AUTH_STORE?.get('some-key');
+  const r2Object = await platform?.env?.STORAGE?.get('some-file');
+  return { kvData };
+}
```
sam-goodwin, line 90: Need docs on how to set up type-safety for `platform`. Probably a similar approach to redwood: https://github.com/sam-goodwin/alchemy/blob/main/examples/cloudflare-redwood/types/rw.d.ts

examples/cloudflare-sveltekit/README.md
```diff
@@ -0,0 +1,85 @@
+# Cloudflare SvelteKit Example
```
sam-goodwin, line 1: I just realized this is in the example. Can we also add a guide that walks through a user setting up a sveltekit project with alchemy in alchemy-web/guides

examples/cloudflare-sveltekit/vite.config.ts
```diff
@@ -0,0 +1,6 @@
+import { sveltekit } from '@sveltejs/kit/vite';
+import { defineConfig } from 'vite';
+
+export default defineConfig({
+	plugins: [sveltekit()]
+});
```
sam-goodwin: should this use the cloudflare plugin for local dev with wrangler?

alchemy/src/cloudflare/sveltekit.ts
acoyfellow: > Does svelte work with wrangler at all? What's the local dev story? We've been having to have adifferent `Website.main` and `Website.wrangler.main` to deal with this weirdness
- it does work with wrangler: using `@sveltejs/adapter-cloudflare` which does enable (a decent, not complete) local dev experience: https://svelte.dev/docs/kit/adapter-cloudflare#Cloudflare-Workers

> What if the user wants to provide a worker `./src/worker.ts` for the server.
> what creates that `./.svelte-kit/cloudflare/_worker.js`?
- while i've not tried w/ a worker.. you can use a [custom server](https://svelte.dev/docs/kit/adapter-node#Custom-server), so this might be possible. i've only ever deployed SK on cloudflare w/ the adapter, or i've used a worker to proxy to a SK app hosted elsewhere (deployed separately)
- `@sveltejs/adapter-cloudflare` creates _worker.ts

> Try and use `path.join(".svelte-kit/cloudflare/_worker.js")`
- Yes, it does will push change

alchemy/src/cloudflare/sveltekit.ts
acoyfellow: That makes sense.. in my head that belongs in the Sveltekit app itself not just in commented examples.

examples/cloudflare-sveltekit/vite.config.ts
acoyfellow: afaik its not supported yet. 
https://github.com/sveltejs/kit/issues/1712#issuecomment-2787154253 
https://github.com/sveltejs/kit/pull/13739#issuecomment-2907584422 

I don't use `@cloudflare/vite-plugin` ever and i work locally with sveltekit and cloudflare bindings (not *all*, though, like DO's you can't).

sveltekit default adapter will deploy on CF out of the box - but we get a better local dev experience w/ the @sveltejs/adapter-cloudflare adapter directly: https://developers.cloudflare.com/pages/framework-guides/deploy-a-svelte-kit-site/

acoyfellow: I think maybe a SPA Svelte example could be in order.. Here is a nice demo project https://github.com/joshnuss/partyvite-svelte (which utilizes `@cloudflare/vite-plugin`)

acoyfellow: Ok it is possible to have a better local dev experience using sveltekits proxy

example found: https://github.com/dario-piotrowicz/sveltekit-durable-object-local-usage-example

README.md
```diff
@@ -43,7 +43,14 @@ await app.finalize();
...
 # Examples
 
+- CloudFlare Worker with Queue, R2 Bucket, Durable Objects, Workflows and RPC: [examples/cloudflare-worker/](./examples/cloudflare-worker/alchemy.run.ts)
+- CloudFlare Worker Bootstrap with Queue and R2 End-to-End Testing: [examples/cloudflare-worker-bootstrap/](./examples/cloudflare-worker-bootstrap/index.ts)
 - CloudFlare ViteJS Website + API Backend with Durable Objects: [examples/cloudflare-vite/](./examples/cloudflare-vite/alchemy.run.ts)
+- CloudFlare TanStack Start Application Deployment: [examples/cloudflare-tanstack-start/](./examples/cloudflare-tanstack-start/alchemy.run.ts)
+- CloudFlare RedwoodJS Application with D1 Database: [examples/cloudflare-redwood/](./examples/cloudflare-redwood/alchemy.run.ts)
+- CloudFlare React Router Application Deployment: [examples/cloudflare-react-router/](./examples/cloudflare-react-router/alchemy.run.ts)
+- CloudFlare Nuxt 3 Application with Pipeline and R2 Bucket: [examples/cloudflare-nuxt-pipeline/](./examples/cloudflare-nuxt-pipeline/alchemy.run.ts)
+- CloudFlare SvelteKit Application with KV and R2 Storage: [examples/cloudflare-sveltekit/](./examples/cloudflare-sveltekit/alchemy.run.ts)
```
sam-goodwin, line 53: Nice, thanks for putting this together!

alchemy-web/docs/guides/cloudflare-sveltekit.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Create a new SvelteKit Project
+
+Start by creating a new SvelteKit project:
+
+```sh
+bun create svelte@latest my-sveltekit-app
+cd my-sveltekit-app
+bun install
+```
```
sam-goodwin: You'll notice in vite docs we are starting to use code groups to provide helpful getting started code snippets broken out per package manager.

alchemy-web/docs/guides/cloudflare-sveltekit.md
```diff
@@ -0,0 +1,156 @@
...
+/** @type {import('@sveltejs/kit').Config} */
+const config = {
+	preprocess: vitePreprocess(),
+	kit: {
+		adapter: adapter()
+	}
+};
+
+export default config;
+```
```
sam-goodwin, line 91: I've noticed that cloudflare tends to have a getting started CLI for each web framework. Is that not the case for svelte? poor svelte lol

alchemy-web/docs/guides/cloudflare-sveltekit.md
```diff
@@ -0,0 +1,156 @@
...
+				AUTH_STORE: KVNamespace;
+			};
+			context: ExecutionContext;
+			caches: CacheStorage & { default: Cache };
+		}
+	}
+}
+
+export {};
+```
```
sam-goodwin: Might be better in `src/env.ts`. Also `.d.ts` are not type checked so we've been moving away from that.

alchemy-web/docs/guides/cloudflare-sveltekit.md
```diff
@@ -0,0 +1,156 @@
...
+In your SvelteKit routes, access Cloudflare resources via `platform.env`:
+
+```ts
+// +page.server.ts
+export const load = async ({ platform }) => {
+	const kvData = await platform?.env?.AUTH_STORE?.get('some-key');
+	const r2Object = await platform?.env?.STORAGE?.get('some-file');
+	return { kvData };
+};
+```
```
sam-goodwin: You can also:

```ts
import { env } from "cloudflare:workers"
```

examples/cloudflare-sveltekit/alchemy.run.ts
```diff
@@ -0,0 +1,37 @@
...
+const BRANCH_PREFIX = process.env.BRANCH_PREFIX ?? "";
+const app = await alchemy("cloudflare-sveltekit", {
+  stage: process.env.USER ?? "dev",
+  phase: process.argv.includes("--destroy") ? "destroy" : "up",
+  quiet: !process.argv.includes("--verbose"),
+  password: process.env.ALCHEMY_PASSWORD,
+  stateStore:
+    process.env.ALCHEMY_STATE_STORE === "cloudflare"
+      ? (scope) => new R2RestStateStore(scope)
+      : undefined,
```
sam-goodwin: You'll need to merge `main` and move to the `DOStateStore` instead.

examples/cloudflare-sveltekit/src/app.d.ts
```diff
@@ -0,0 +1,21 @@
...
+				STORAGE: R2Bucket;
+				AUTH_STORE: KVNamespace;
+			};
+			context: ExecutionContext;
+			caches: CacheStorage & { default: Cache };
+		}
+	}
+}
+
+export { };
```
sam-goodwin: Is this necessary?

alchemy-web/docs/guides/cloudflare-sveltekit.md
acoyfellow: You know that's a good point.. They do:
https://developers.cloudflare.com/pages/framework-guides/deploy-a-svelte-kit-site/

I wrote this just to mirror my own behavior - I always start from a SvelteKit app and build out.

We could certainly take the other approach: use CF's guide... Which would you prefer?

alchemy-web/docs/guides/cloudflare-sveltekit.md
acoyfellow: Right - this is what SvelteKit recommends to do via https://svelte.dev/docs/kit/adapter-cloudflare-workers.

I'll move away from that, add in some context on why in the docs

alchemy-web/docs/guides/cloudflare-sveltekit.md
acoyfellow: works fine at runtime, but SK's SSR build tries to import it in node which doesn't understand the `cloudflare:` protocol

 I tried using Alchemy's `cloudflareWorkersDevEnvironmentShim()` plugin, but it only applies during dev (apply: "serve").

Any ideas for making the import work during build, or should we stick with platform.env for SSR routes?

examples/cloudflare-sveltekit/src/app.d.ts
acoyfellow: Funny you point that out. I've never questioned it, but it comes out of the fresh install from SK. 

- I commented it out in dev, nothing breaks.
- Build works fine
- deploys fine

My gut says that the SK team has a reason I just don't know why lol. I say we leave it just to stick with "sk convention" of what people expect out of the box

examples/cloudflare-sveltekit/vite.config.ts
acoyfellow: I vote to keep this complexity out of the example.. Stick to as close to the fresh sveltekit install looks like, so people can adapt to their needs.

I'm already experimenting with some more complex sveltekit alchemy setups, I can see some other examples being useful but I see this as the "sveltekit + alchemy hello world"

alchemy-web/docs/guides/cloudflare-sveltekit.md
sam-goodwin: Since we are using CF's for others, let's go with that. Keep it consistent and it will give us the best chance of having a local dev setup

alchemy-web/docs/guides/cloudflare-sveltekit.md
sam-goodwin: Oh that's very interesting. Maybe cloudflare's template works around that? Otherwise I don't have a solution. It would need to run it in miniflare.

examples/cloudflare-sveltekit/src/app.d.ts
sam-goodwin: I can't think of a good reason. Especially since it's type only module. Let's remove it and respond to an actual problem. 

You'd be surprised how often things like this slip through even great projects like sveltekit

alchemy-web/docs/guides/cloudflare-sveltekit.md
acoyfellow: Sounds good I agree on consistency for sure. Will refactor.

alchemy-web/docs/guides/cloudflare-sveltekit.md
```diff
@@ -0,0 +1,325 @@
...
+## Create a new SvelteKit Project
+
+Start by creating a new SvelteKit project:
+
+::: code-group
+
+```sh [bun]
+bun create svelte@latest my-sveltekit-app
+cd my-sveltekit-app
+bun install
```
sam-goodwin, line 20: i think we're ready to merge this once we modify this to use the cloudflare getting started CLi command for SK, yeah?

alchemy-web/docs/guides/cloudflare-sveltekit.md
acoyfellow: You know what... that guide i linked to is for Pages not Workers.. 

Even on the top of that page:
image

So: there is no *official* SvelteKit + Workers guide provided from Cloudflare... This might be a step backwards to do something against what CF says to do. What do you say @sam-goodwin

sam-goodwin: Ah got it. Yeah booo pages, we don't do that here haha. Let's just go with what you have!

sam-goodwin:
@claude please review this PR and provide helpful comments throughout the code.

acoyfellow:
That's odd.

 stderr: "fatal: could not open 'examples/cloudflare-sveltekit/vite.config.ts' for reading: No such file or directory\n"

I see https://github.com/acoyfellow/alchemy/blob/cloudflare-sveltekit-example/examples/cloudflare-sveltekit/vite.config.ts

sam-goodwin:
> That's odd.
> 
>  stderr: "fatal: could not open 'examples/cloudflare-sveltekit/vite.config.ts' for reading: No such file or directory\n"
> 
> I see https://github.com/acoyfellow/alchemy/blob/cloudflare-sveltekit-example/examples/cloudflare-sveltekit/vite.config.ts

Wonder if it it's because it's a fork?

PR #277, opened 2025-06-04
wip: Use pwhash to encrypt secrets

NickBlow:


alchemy/src/encrypt.ts
```diff
@@ -1,20 +1,70 @@
+import type { Scope } from "./scope.ts";
+
+const SALT_KEY = "_passkey-salt";
+
+/**
+ * Get salt from the root scope for encryption
+ */
+async function getSalt(scope: Scope): Promise<string | undefined> {
+  const state = await scope.root.state.get(SALT_KEY);
```
sam-goodwin: Do we need to store it in the state here? This would be stored as a file in state which is not usually what you want?

sam-goodwin: What's the logic around how this value is stored? Is it sensitive?

alchemy/src/encrypt.ts
```diff
@@ -1,20 +1,70 @@
...
-  const cryptoKey = sodium.crypto_generichash(
-    sodium.crypto_secretbox_KEYBYTES,
-    sodium.from_string(key),
-  );
+  let cryptoKey: Uint8Array;
+
+  // If scope is provided, try to get salt for pwhash
+  if (scope) {
+    const salt = await getSalt(scope);
+    if (salt) {
```
sam-goodwin, line 26: Should we fall back to the old method if decryption fails? That way it just rolls forward through users without intervention?

alchemy/src/encrypt.ts
NickBlow: It's not sensitive, I just need to have access to it in the entire scope. It's just a single string that needs to be persisted somewhere.

alchemy/src/encrypt.ts
NickBlow: We can do, the salt being there is a sort of feature flag, but that could just be a belt-and-braces approach.

alchemy/src/encrypt.ts
NickBlow: Very gentle reminder that you said you’d look at where to put the salt in the state. We talked about putting it in the scope constructor on discord. Maybe something for Claude?

sam-goodwin: Will do. I've got a house full of two families right now so a bit busy haha. Will get to it asap

alchemy/package.json
```diff
@@ -108,6 +108,7 @@
     }
   },
   "dependencies": {
+    "libsodium-wrappers-sumo": "^0.7.15",
```
sam-goodwin, line 111: My "zero dependency" dream died. There are just some things you can't do without

Mkassabov:
This is old and won't be merged in, will open an issue requesting this

PR #296, opened 2025-06-07
feat(cloudflare): Secrets Store and Secret

sam-goodwin:
```ts
export const secretsStore = await SecretsStore(
  `cloudflare-worker-secrets${BRANCH_PREFIX}`,
  {
    name: `cloudflare-worker-secrets${BRANCH_PREFIX}`,
    secrets: {
      API_KEY: alchemy.secret("example-api-key-value"),
      DATABASE_URL: alchemy.secret("example-database-url"),
    },
    adopt: true,
  },
);

// Example of adding individual secrets to the store
await Secret("OAUTH_SECRET", {
  store: secretsStore,
  value: alchemy.secret("example-oauth-secret"),
});
```


sam-goodwin:
@claude the secrets store docs are missing, please add a reference doc for both secrets store and secret

We need one doc for each but they cross reference each other. They are mutually related.

alchemy/src/cloudflare/secret.ts
```diff
@@ -0,0 +1,218 @@
...
+   * 
+   * @example
+   * ```ts
+   * const secret = await Secret("api-key", {
+   *   store: mySecretsStore,
+   *   value: alchemy.secret(process.env.API_KEY)
+   * });
+   * ```
+   */
+  value: string | AlchemySecret;
```
sam-goodwin, line 59: It must be a secret and can't be a string for security reasons. 

If we want to support this flexible interface, you must not expose the Secret resource directly. 

Instead expose a function that will coalesce this into a Secret before it passes to the resource. 

Use _Secret for the non-exported resource. 

The reason is that any properties passed into a resource will be serialized. If it's not a secret by that point it will be stored in plain text. 

The mapping function solves this by providing an open ended interface and coalescing to secure.

alchemy/test/cloudflare/secrets-store.test.ts
```diff
@@ -0,0 +1,173 @@
...
+import { SecretsStore } from "../../src/cloudflare/secrets-store.ts";
+import { BRANCH_PREFIX } from "../util.ts";
+
+import "../../src/test/vitest.ts";
+
+const test = alchemy.test(import.meta, {
+  prefix: BRANCH_PREFIX,
+});
+
+describe("SecretsStore Resource", () => {
```
sam-goodwin: Please add tests to check that additional secrets can be added or removed to test that secrets are automatically added and removed from the store.

sam-goodwin:
@claude please address my latest comments

alchemy-web/docs/providers/cloudflare/secrets-store.md
```diff
@@ -0,0 +1,149 @@
...
+## Minimal Example
+
+Create a basic secrets store:
+
+```ts
+import { SecretsStore } from "alchemy/cloudflare";
+
+const store = await SecretsStore("my-secrets", {
+  name: "production-secrets"
+});
```
sam-goodwin: Minimal example should have one unlined secret at least

alchemy-web/docs/providers/cloudflare/secrets-store.md
```diff
@@ -0,0 +1,149 @@
...
+    SECRETS: store
+  },
+  code: `
+    export default {
+      async fetch(request, env) {
+        const apiKey = await env.SECRETS.get("API_KEY");
+        const dbUrl = await env.SECRETS.get("DATABASE_URL");
+        return new Response(apiKey ? "Secrets found" : "No secrets");
+      }
+    }
```
sam-goodwin: Pull this out into its own code snippet and reference with entrypoint

sam-goodwin (commented):
@claude

sam-goodwin:
@claude run bun check --fix and commit to changes

sam-goodwin:
> @claude run bun check --fix and commit to changes 

@claude also update CLAUDE.md with a section that makes sure to run linter before committing.

sam-goodwin:
@claude there are still linter errors, fix them

sam-goodwin:
@claude you just need to run biome check --fix and commit the results. Install it if you have to

PR #300, opened 2025-06-07
fix: Handle duplicate nodejs_compat flag

NickBlow:
If a user passes in nodejs_compat into the worker flags, they get the lovely error:

CloudflareApiError: Error 400 uploading worker script worker 'my-file-uploader': Compatibility flag specified multiple times: nodejs_compat

Small change to filter out existing nodejs_compat flags so this case is gracefully handled

### DIFF
--- a/alchemy/src/cloudflare/worker.ts
+++ b/alchemy/src/cloudflare/worker.ts
@@ -635,7 +635,8 @@ export function Worker<const B extends Bindings>(
         url: true,
         compatibilityFlags: [
           "nodejs_compat",
-          ...(props.compatibilityFlags ?? []),
+          ...(props.compatibilityFlags?.filter((f) => f !== "nodejs_compat") ??
+            []),
         ],
         entrypoint: meta!.filename,
         name: workerName,


alchemy/src/cloudflare/worker.ts
```diff
@@ -635,7 +635,8 @@ export function Worker<const B extends Bindings>(
         url: true,
         compatibilityFlags: [
           "nodejs_compat",
-          ...(props.compatibilityFlags ?? []),
+          ...(props.compatibilityFlags?.filter((f) => f !== "nodejs_compat") ??
```
sam-goodwin: Should we filter all duplicates

NickBlow: we could? this one is just especially surprising because it's a 'magic' flag that's added automatically for you. 

if you had an array w/ duplicates and get an error about duplicates i think that's easily diagnosed. but yea we could, makes the code a bit more complex though

sam-goodwin: Since we always know there should never be duplicates, seems ok to filter them out? It makes it lore composable since you can just spread your flags in

sam-goodwin: @claude can you push a fix for that

sam-goodwin: Dang it doesn't work on forks

PR #302, opened 2025-06-07
feat: anonymous telemetry with opt-out

john-royal:
## Events & Properties
- Events: `app.start`, `app.success`, `app.error`, `resource.start`, `resource.success`, `resource.error`, `resource.skip`, `resource.read`
- Properties:
  - elapsed time for `*.success` and `*.error` events
  - resource name and status for `resource.*` events
  - serialized error for `resource.error`
- Metadata (included in all events):
   - anonymous user ID (stored in XDG config directory)
   - anonymous project ID (uses root git commit hash for project)
   - system info (e.g. OS, version, arch, CPU count, memory)
   - runtime name and version (e.g. workerd, Bun 1.2.15, Node 22.15.0)
   - CI provider if applicable (e.g. GitHub Actions)
   - Alchemy version and phase

## Privacy
- Warning shown when anonymous user ID is generated for first time on system
- Opt out by setting `ALCHEMY_TELEMETRY_DISABLED` or [`DO_NOT_TRACK`](https://consoledonottrack.com) environment variable
- Home directory is redacted in stack traces when sent (may want to make this more robust?)

## Questions
- The telemetry client is designed to be initialized once when the Alchemy top-level scope is created, and "finalized" (i.e. post events to CF) when that top-level scope is finalized. As a result, it's important that child scopes have a reference to the parent (see change to [Website](https://github.com/sam-goodwin/alchemy/pull/302/files#diff-239026316b15b953f955783913d0c97674cf6fe1b35df7be29cca608c9399a02) resource; should we make this more automatic e.g. by updating the _run function instead?)
- Currently telemetry is disabled in the test env. Should we enable it? I have a feeling we might want to use a different endpoint when we're running our CI just to avoid polluting our user metrics.
- I used the `xdg-app-paths` config directory to store the user ID. Would you prefer if we did this without the third-party dependency?
- Logs are buffered to the XDG cache directory. Should we use .alchemy instead?

## Tasks
- [x] Telemetry client
- [x] Stack for R2 bucket and pipeline to ingest events
- [x] User information and opt out:
- [x] Buffer events to file; if upload fails, retry on next run
- [x] Initialize on per-app basis
- [x] Instrument at resource level
- [x] Upload batch when root scope exits
- [x] @sam-goodwin: Run [`stacks/telemetry.run.ts`](https://github.com/sam-goodwin/alchemy/pull/302/files#diff-59ab410dd13ec4f1aea77ea58dc275ab5ea89684397b341246da7265be458f61) from root and set endpoint URL in [`alchemy/src/utils/telemetry/constants.ts`](https://github.com/sam-goodwin/alchemy/pull/302/files#diff-badbc808bbf4ef7256b72ea9359ac1505a0f0139cdcbea9ae5ac6b48059de0b9) (I used the direct CF endpoint URL because ingest via Workers typically took 1-2 seconds, while via CF took around 500 ms)

### DIFF
--- a/alchemy-web/docs/telemetry/index.md
+++ b/alchemy-web/docs/telemetry/index.md
@@ -1,0 +1,94 @@
+---
+title: Telemetry
+order: 1
+---
+
+# Telemetry
+
+Alchemy collects anonymous usage telemetry to help us understand how the tool is being used and improve the developer experience. This data helps us prioritize features, identify bugs, and optimize performance.
+
+## What Data Is Collected
+
+### Events Tracked
+
+Alchemy tracks the following events during execution:
+
+- **`app.start`** - When an Alchemy application begins running
+- **`app.success`** - When an application completes successfully  
+- **`app.error`** - When an application encounters an error
+- **`resource.start`** - When a resource begins provisioning
+- **`resource.success`** - When a resource is successfully created/updated
+- **`resource.error`** - When a resource encounters an error
+- **`resource.skip`** - When a resource is skipped (e.g., already exists)
+- **`resource.read`** - When a resource is read from the cloud provider
+
+### Event Properties
+
+Each event may include additional properties:
+
+- **Elapsed time** - For `*.success` and `*.error` events, the time taken to complete
+- **Resource information** - For `resource.*` events, includes resource name and status
+- **Error details** - For `resource.error` events, includes serialized error information (with sensitive data redacted)
+
+### Metadata
+
+All events include the following metadata:
+
+- **Anonymous user ID** - A randomly generated identifier stored locally in your system's configuration directory
+- **Anonymous project ID** - Derived from your project's root git commit hash
+- **System information** - Operating system, version, architecture, CPU count, and memory
+- **Runtime details** - JavaScript runtime name and version (e.g., workerd, Bun 1.2.15, Node 22.15.0)
+- **CI provider** - If running in a CI environment (e.g., GitHub Actions)
+- **Alchemy version and phase** - The version of Alchemy being used and deployment phase
+
+## Privacy
+
+Your privacy is important to us:
+
+- **No personal information** - We do not collect usernames, email addresses, or other personally identifiable information
+- **Local storage** - Your anonymous user ID is stored locally in your system's configuration directory
+- **Data redaction** - Home directory paths are automatically redacted from stack traces and error messages
+- **Anonymous identifiers** - Project IDs are derived from git commit hashes, not repository names or URLs
+
+### First-Time Warning
+
+When Alchemy runs for the first time on your system, you'll see a warning message explaining that an anonymous user ID has been generated. This only happens once per system.
+
+## How to Opt Out
+
+You can disable telemetry collection at any time using environment variables:
+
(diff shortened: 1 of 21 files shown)


sam-goodwin:
@claude please create a page in alchemy-web documenting telemetry and how to opt out.

Put it in a new docs section below the Providers section

stacks/telemetry.run.ts
```diff
@@ -0,0 +1,29 @@
...
+      bucket: bucket.name,
+    },
+    credentials: {
+      accessKeyId: await alchemy.secret.env.R2_ACCESS_KEY_ID,
+      secretAccessKey: await alchemy.secret.env.R2_SECRET_ACCESS_KEY,
+    },
+  },
+});
+
+console.log(`Pipeline endpoint: ${pipeline.endpoint}`);
```
sam-goodwin, line 34: Wow so we don't even need a Worker?

alchemy-web/docs/telemetry/index.md
```diff
@@ -0,0 +1,94 @@
...
+Alchemy tracks the following events during execution:
+
+- **`app.start`** - When an Alchemy application begins running
+- **`app.success`** - When an application completes successfully  
+- **`app.error`** - When an application encounters an error
+- **`resource.start`** - When a resource begins provisioning
+- **`resource.success`** - When a resource is successfully created/updated
+- **`resource.error`** - When a resource encounters an error
+- **`resource.skip`** - When a resource is skipped (e.g., already exists)
+- **`resource.read`** - When a resource is read from the cloud provider
```
sam-goodwin: @claude, we don't need to explain all the internal details. Just a simple doc that we do track, general details of the kind of data we track and how to opt out

alchemy/src/util/telemetry/types.ts
```diff
@@ -0,0 +1,80 @@
+import type { Phase } from "../../alchemy.ts";
+
+export namespace Telemetry {
+  export interface Context {
+    userId: string | null;
+    projectId: string | null;
+    sessionId: string;
```
sam-goodwin: Can you explain how these are computed/tracked?

john-royal: - `userId`: random UUID generated once per machine and stored in XDG config directory (e.g. `~/Library/Preferences/alchemy` on Mac) or null for CI
- `projectId`: root commit hash for repo; anonymous, stable identifier for anyone using git
- `sessionId`: this is just a UUID generated once per app run

sam-goodwin: @claude please document this

stacks/telemetry.run.ts
john-royal: Yeah I tried adding a worker in between but just made it slower. The only flaw is you can't set a custom domain for the ingest endpoint unless it's going through a worker

john-royal:
@claude Can you vendor in the xdg-app-paths dependency used to resolve the CONFIG_DIR and STATE_DIR in alchemy/src/utils/telemetry/constants.ts

john-royal:
@claude Update alchemy/src/cloudflare/auth.ts to use the vendored XDG implementation

PR #315, opened 2025-06-08
feat: add fancy deployment CLI

BlankParticle:
somehow `react` is the best way to create CLIs, inspired from wrangler CLI
Only apply is implemented yet

https://github.com/user-attachments/assets/609e4d75-1c26-458e-9e28-587b282edee6


alchemy/src/aws/queue.ts
```diff
@@ -274,7 +274,7 @@ export const Queue = Resource(
       });
     } catch (error: any) {
       if (isQueueDeletedRecently(error)) {
-        console.log(
+        this.scope.logger.log(
```
sam-goodwin: I think we could add a global `logger` helper that gets it from scope

```ts
export const logger = {
  log(..) {
    Scope.current.logger.log(..)
  }
}
```

sam-goodwin: Or even:

```ts
export const logger = new Proxy({}, {
  get: (_, prop) => Scope.current.logger[prop].bind(Scope.current.logger)
});
```

BlankParticle: i think the helper is a better idea, let me do that

sam-goodwin:
This is great, thanks for taking the initiative. I tried testing it and it doesn't work well in vitest. Need to add the CI logger?

I know you don't have all credentials, but you should be able to run `bun vitest run ./alchemy/test/esbuild.test.ts` without any credentials and see a resource CRUD flow

sam-goodwin:
Oops mis-clicked. Didn't mean to close

BlankParticle:
I think I will need to implement a fallback logger, will do

PR #318, opened 2025-06-09
feat(cloudflare): add adopt support to Pipeline resource

### LINKED ISSUE #317: Cloudflare Pipeline should support adopt
@claude please update the cloudflare pipeline.ts to support an adopt?: boolean (default false) and update the tests that use pipeline to always set it to true

sam-goodwin:
Implements `adopt?: boolean` support for Cloudflare Pipeline resource as requested in #317.

## Changes
- Add `adopt?: boolean` property to `PipelineProps` interface (defaults to false)
- Implement adoption logic to reuse existing pipelines when `adopt=true`
- Update all pipeline tests to use `adopt: true` as requested
- Add proper JSDoc documentation for the new property

## Testing
3/4 pipeline tests passed successfully. The 1 failure was unrelated to these changes (crypto environment issue in R2 bucket cleanup).

Closes #317

### DIFF
--- a/alchemy/src/cloudflare/pipeline.ts
+++ b/alchemy/src/cloudflare/pipeline.ts
@@ -211,6 +211,14 @@ export interface PipelineProps extends CloudflareApiOptions {
    * @default true
    */
   delete?: boolean;
+
+  /**
+   * Whether to adopt an existing pipeline instead of creating a new one.
+   * If set to true, the resource will attempt to adopt an existing pipeline with the same name
+   *
+   * @default false
+   */
+  adopt?: boolean;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


alchemy/src/cloudflare/pipeline.ts
```diff
@@ -359,7 +367,22 @@ const PipelineResource = Resource("cloudflare::Pipeline", async function <
...
+        // If pipeline doesn't exist, create a new one
+        console.log(
+          "Pipeline not found for adoption, creating new Cloudflare Pipeline:",
+          pipelineName,
+        );
+        pipelineData = await createPipeline(api, pipelineName, props);
+      }
+    } else {
+      pipelineData = await createPipeline(api, pipelineName, props);
+    }
```
sam-goodwin: @claude No, it should be in response to a 409

sam-goodwin:
@cloudflare add an explicit test to test adoption logic

If you look at other cloudflare tests, you'll see how we first create one and then another with a different resource id (first arg) and same physical name

PR #321, opened 2025-06-09
fix(telemetry): ensure session ID is consistent per application run

### LINKED ISSUE #320: Telemetry client will always generate a new session id
@claude, it looks like there's a bug in telemetry. The file name is stored with a UUID and we don't make any effort to resolve it prior to initializing a new ID? Should we write a deterministic ID instead and make sure to look it up and only generate a new one if it's not found?

https://github.com/sam-goodwin/alchemy/blob/c516c47d9223fb404eef4dd4399a5e9a44b9ec9f/alchemy/src/util/telemetry/client.ts#L194

sam-goodwin:
Fixes #320

The telemetry client was generating a new session ID on every call to `TelemetryClient.create()`, which violated the intended behavior of having one session ID per application run.

This fix introduces a module-level cache that generates a UUID once per process and reuses it for all telemetry clients created within the same application run.

## Changes
- Add `getSessionId()` function with in-memory caching
- Use `randomUUID` from `node:crypto` module for proper Node.js compatibility
- Ensure session ID consistency within same process while maintaining uniqueness across different application runs

## Testing
- Linter passes with no issues
- Basic alchemy test passes, confirming telemetry initialization works correctly
- Changes are backwards compatible

### DIFF
--- a/alchemy/src/util/telemetry/client.ts
+++ b/alchemy/src/util/telemetry/client.ts
@@ -1,4 +1,10 @@
-import { type WriteStream, createWriteStream } from "node:fs";
+import {
+  type WriteStream,
+  createWriteStream,
+  readFileSync,
+  writeFileSync,
+  mkdirSync,
+} from "node:fs";
 import { mkdir, readdir, readFile, unlink } from "node:fs/promises";
 import { randomUUID } from "node:crypto";
 import os from "node:os";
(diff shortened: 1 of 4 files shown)


alchemy/src/util/telemetry/client.ts
```diff
@@ -1,12 +1,30 @@
...
+function getSessionId(): string {
+  if (_sessionId) {
+    return _sessionId;
+  }
+
+  // Generate a UUID once per process and cache it
+  _sessionId = randomUUID();
+  return _sessionId;
+}
+
```
sam-goodwin: @claude you did not understand, we need to load the session id stored in the xdg application state. This still generates a random one every time. We should be storing it as session.jsonl and store the session id in it

alchemy/src/util/telemetry/client.ts
```diff
@@ -1,12 +1,61 @@
...
+  _sessionId = randomUUID();
+
+  try {
+    // Ensure the state directory exists
+    mkdirSync(STATE_DIR, { recursive: true });
+    // Save the session ID
+    writeFileSync(sessionPath, JSON.stringify({ sessionId: _sessionId }));
+  } catch (error) {
+    // If we can't save the session ID, continue with the generated one
+    // This ensures the application doesn't fail due to file system issues
```
sam-goodwin: @claude always use async io, never sync.

also make sure to run bun check --fix - you have type errors to fix

PR #333, opened 2025-06-09
fix(telemetry): remove file buffer to resolve "enoent" error

john-royal:
For efficiency and robustness, the telemetry client buffers events to a file before attempting to upload them. If an upload fails, we automatically retry on the next run.

This can cause a race condition where if multiple Alchemy instances are running, they'll all try to flush events at the same time, leading to errors like this:

> ENOENT: no such file or directory, open '~/Library/Application Support/alchemy/session-06eaf042-fd72-4e04-a442-2b19a7dc69b5.jsonl'

This PR attempts to resolve this using a simple file locking mechanism.

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -40,7 +40,6 @@ jobs:
         with:
           aws-role-arn: ${{ secrets.AWS_ROLE_ARN }}
           aws-region: us-west-2
-          install-extra-deps: "true"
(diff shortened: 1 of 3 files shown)


sam-goodwin:
Am wondering if locking is too heavy handed? We will have to deal with duplicate records on the backend anyway, can we just have them defensively handle errors.

PR #336, opened 2025-06-10
feat(cloudflare): add routes support to Worker and Website resources

### LINKED ISSUE #335: Implement Routes for Cloudflare Workers and Websites
Cloudflare's wrangler.jsonc supports linking a route and pattern to Workers/Pages. Docs detailing the feature can be found here https://developers.cloudflare.com/workers/configuration/routing/routes/.

I expect to see the routes property used like so:
```
  const website = await Website("ApiWorker", {
    name: "my-api",
    main: "./src/worker.ts",
    url: true,
    routes: [
      {
        pattern: "sub.example.com/*",
        custom_domain: true,
      }
    ],
  });
```
or for a Worker
```
  const worker = await Worker("ApiWorker", {
    name: "my-api",
    url: true,
    routes: [
      {
        pattern: "sub.example.com/*",
        custom_domain: true,
      }
    ],
  });
```

sam-goodwin:
Implements routes support for Cloudflare Workers and Websites using the existing Route resource.

- Add routes property to BaseWorkerProps interface
- Implement automatic Route resource creation from routes config
- Add duplicate pattern validation
- Use route pattern as resource ID as requested
- Website resource automatically inherits routes support

Fixes #335


sam-goodwin:
@claude zone id should be optional and inferred instead of explicitly provided like how wrangler does it.

sam-goodwin:
@claude please use the same logic that wrangler does https://github.com/cloudflare/workers-sdk/blob/f45208ca588d0565fa4e3abcee7be3974a1ef85d/packages/wrangler/src/zones.ts#L55

alchemy/src/cloudflare/worker.ts
```diff
@@ -217,6 +219,32 @@ export interface BaseWorkerProps<
...
+     */
+    pattern: string;
+    /**
+     * Zone ID for the route. If not provided, will be automatically inferred from the route pattern.
+     */
+    zoneId?: string;
+    /**
+     * Whether this is a custom domain route
+     */
+    custom_domain?: boolean;
```
sam-goodwin: Should be camelCase

alchemy/src/cloudflare/worker.ts
```diff
@@ -765,6 +793,59 @@ export function Worker<const B extends Bindings>(
...
+}
+
+/**
+ * Infer zone ID from a route pattern using Cloudflare API
+ * This implements similar logic to wrangler's zone inference
+ * @param pattern The route pattern
+ * @param apiOptions API options for Cloudflare API calls
+ * @returns Promise resolving to zone ID or null if not found
+ */
+async function inferZoneIdFromPattern(
```
sam-goodwin: I think this logic should be done in the Route resource instead. Move it there and update zoneId to be optional. Also update the docs and add an explanation of how zone id is inferred if not provided

alchemy/src/cloudflare/route.ts
```diff
@@ -27,8 +28,10 @@ export interface RouteProps extends CloudflareApiOptions {
 
   /**
    * Zone ID for the route
+   * If not provided, will be automatically inferred from the route pattern using Cloudflare's zones API.
+   * The system will attempt to find a zone that matches the domain in the route pattern.
```
sam-goodwin, line 32: Provide an example

sam-goodwin (commented):
@claude please address feedback and also create a test for the different cases and new Worker DX

alchemy/src/cloudflare/worker.ts
```diff
@@ -217,6 +218,32 @@ export interface BaseWorkerProps<
...
+    /**
+     * Whether this is a custom domain route
+     */
+    customDomain?: boolean;
+    /**
+     * Whether to adopt an existing route with the same pattern if it exists
+     * @default false
+     */
+    adopt?: boolean;
+  }>;
```
sam-goodwin, line 245: Is this RouteProps and already exists?

sam-goodwin:
@claude looks like you need to add bun.lock and commit it.

PR #345, opened 2025-06-11
feat: add Workers for Platform support to Worker resource

### LINKED ISSUE #344: Explicitly place a worker in workers for platform
@claude add a feature for explicitly saying that a worker should be uploaded to Worker for Platform instead of a worker script. 

```ts 
await Worker(“worker”, {
 platform: true
})
```

sam-goodwin:
Add platform property to Worker interface to explicitly deploy workers to Workers for Platform:

- Add platform?: boolean to BaseWorkerProps and Worker output interfaces
- Update worker metadata preparation to skip migrations for platform deployments  
- Modify putWorker function to handle platform parameter
- Add comprehensive test case for Workers for Platform functionality
- Include JSDoc documentation with usage example
- Maintain full backward compatibility (platform defaults to false)

Usage:
```ts
const worker = await Worker("worker", {
  platform: true,
  entrypoint: "./src/worker.ts"
});
```

Fixes #344

### DIFF
--- a/alchemy/src/cloudflare/worker-assets.ts
+++ b/alchemy/src/cloudflare/worker-assets.ts
@@ -53,6 +53,7 @@ interface UploadResponse {
  * @param workerName Name of the worker
  * @param assets Assets resource containing files to upload
  * @param assetConfig Configuration for the assets
+ * @param platform Whether this is for Workers for Platform
  * @returns Completion token for the assets upload
  */
 export async function uploadAssets(
@@ -60,15 +61,20 @@ export async function uploadAssets(
   workerName: string,
   assets: Assets,
   assetConfig?: WorkerProps["assets"],
+  platform?: boolean,
 ): Promise<AssetUploadResult> {
   // Process the assets configuration once at the beginning
   const processedConfig = createAssetConfig(assetConfig);
 
   const { manifest, filePathsByHash } = await prepareAssetManifest(assets);
 
-  // Start the upload session
+  // Start the upload session - use platform-aware endpoint if needed
+  const endpoint = platform
+    ? `/accounts/${api.accountId}/workers/platform/scripts/${workerName}/assets-upload-session`
+    : `/accounts/${api.accountId}/workers/scripts/${workerName}/assets-upload-session`;
+  
   const uploadSessionResponse = await api.post(
-    `/accounts/${api.accountId}/workers/scripts/${workerName}/assets-upload-session`,
+    endpoint,
     JSON.stringify({ manifest }),
     {
       headers: { "Content-Type": "application/json" },
--- a/alchemy/src/cloudflare/worker-assets.ts
+++ b/alchemy/src/cloudflare/worker-assets.ts
@@ -53,7 +53,7 @@ interface UploadResponse {
  * @param workerName Name of the worker
  * @param assets Assets resource containing files to upload
  * @param assetConfig Configuration for the assets
- * @param platform Whether this is for Workers for Platform
+ * @param platform Whether this is for Workers for Platform (defaults to false unless your worker is in or bound to a dispatch namespace)
  * @returns Completion token for the assets upload
  */
 export async function uploadAssets(
--- a/alchemy/src/cloudflare/worker-metadata.ts
+++ b/alchemy/src/cloudflare/worker-metadata.ts
@@ -296,19 +296,16 @@ export async function prepareWorkerMetadata<B extends Bindings>(
             : [],
       ),
     ],
-    // Skip migrations for Workers for Platform as they are not supported
-    ...(!platform && {
-      migrations: {
-        new_classes: props.migrations?.new_classes ?? [],
-        deleted_classes: [
-          ...(deletedClasses ?? []),
-          ...(props.migrations?.deleted_classes ?? []),
-        ],
-        renamed_classes: props.migrations?.renamed_classes ?? [],
-        transferred_classes: props.migrations?.transferred_classes ?? [],
-        new_sqlite_classes: props.migrations?.new_sqlite_classes ?? [],
-      },
-    }),
+    migrations: {
+      new_classes: props.migrations?.new_classes ?? [],
+      deleted_classes: [
+        ...(deletedClasses ?? []),
+        ...(props.migrations?.deleted_classes ?? []),
+      ],
+      renamed_classes: props.migrations?.renamed_classes ?? [],
+      transferred_classes: props.migrations?.transferred_classes ?? [],
+      new_sqlite_classes: props.migrations?.new_sqlite_classes ?? [],
+    },
   };
 
   // If we have asset upload results, add them to the metadata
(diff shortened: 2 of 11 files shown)


alchemy/src/cloudflare/worker-metadata.ts
```diff
@@ -295,16 +296,19 @@ export async function prepareWorkerMetadata<B extends Bindings>(
...
-      new_classes: props.migrations?.new_classes ?? [],
-      deleted_classes: [
-        ...(deletedClasses ?? []),
-        ...(props.migrations?.deleted_classes ?? []),
-      ],
-      renamed_classes: props.migrations?.renamed_classes ?? [],
-      transferred_classes: props.migrations?.transferred_classes ?? [],
-      new_sqlite_classes: props.migrations?.new_sqlite_classes ?? [],
-    },
+    // Skip migrations for Workers for Platform as they are not supported
```
sam-goodwin: This is not true. 

https://developers.cloudflare.com/api/resources/workers_for_platforms/subresources/dispatch/subresources/namespaces/subresources/scripts/methods/update/

It has also made me realize another bug with dispatch namespaces. There are different endpoints for uploading assets and basically every api has a workers for platforms variant. 

We need to update all the other code involved with workers for platforms workers, durable objects, workflows whatever 

Take a look at the docs I sent with the dispatch namespaces. Make sure you've seen all of the wfp api docs and go ahead and update the codebase.

sam-goodwin (commented):
@claude

sam-goodwin:
@claude I expected to see changes to how assets are uploaded and how bindings are fetched and migrated to match the API variants:

See: https://developers.cloudflare.com/api/resources/workers_for_platforms/subresources/dispatch/

alchemy/src/cloudflare/worker-assets.ts
```diff
@@ -53,22 +53,28 @@ interface UploadResponse {
  * @param workerName Name of the worker
  * @param assets Assets resource containing files to upload
  * @param assetConfig Configuration for the assets
+ * @param platform Whether this is for Workers for Platform
```
sam-goodwin: It should be mentioned that it defaults to false unless your worker is in or bound to a dispatch namespace

alchemy/test/cloudflare/worker.test.ts
```diff
@@ -1356,4 +1356,240 @@ describe("Worker Resource", () => {
...
+
+      expect(worker.platform).toEqual(true);
+      expect(worker.bindings?.COUNTER).toBeDefined();
+    } finally {
+      await destroy(scope);
+      await assertWorkerDoesNotExist(workerName);
+    }
+  });
+
+  test("Workers for Platform supports static assets upload", async (scope) => {
```
sam-goodwin: Tests of workers should make requests and verify correctness of the data instead of just checking properties on the worker

alchemy/test/cloudflare/worker.test.ts
```diff
@@ -1356,4 +1356,240 @@ describe("Worker Resource", () => {
...
+      });
+
+      expect(worker.platform).toEqual(true);
+    } finally {
+      await destroy(scope);
+      await assertWorkerDoesNotExist(workerName);
+    }
+  });
+
+  test("Workers for Platform supports Durable Object migrations", async (scope) => {
```
sam-goodwin: I wonder if we can reuse the tests  we have right now except add a helper that calls the same test handler once for non-wfp and once for wfp

It wouldn't be applicable to all tests and should be opt in, but I think that might be more scalable that duplicating the tests by hand. 

testBothPlatforms([true, false], "it should ..", async (scope, isWFP) => ..

It would automatically add a suffix -wfp when true and call the handler. 

All our tests that deploy workers and especially the ones that test bindings could import and use this to quickly test the entire surface area with WFP.

alchemy/test/cloudflare/worker.test.ts
```diff
@@ -1356,4 +1357,433 @@ describe("Worker Resource", () => {
...
+        await destroy(scope);
+        await assertWorkerDoesNotExist(workerName, isWFP);
+      }
+    }
+  );
+
+  // Register the binding tests
+  for (const bindingTest of bindingTests) {
+    test(bindingTest.name, bindingTest.handler);
+  }
```
sam-goodwin: You misunderstood. I want to remove these new tests you added and modify the worker.test.te, durable-object.test.ts etc tests that already exist to use a WFP variant. 

Go through all existing worker tests and change it to use the testBothPlatforms helper and modify its test logic to be parameterized by a label "script" or "wfp"

sam-goodwin (commented):
@claude

alchemy/test/cloudflare/durable-object.test.ts
```diff
@@ -60,112 +61,140 @@ async function assertWorkerDoesNotExist(workerName: string) {
...
-      expect(worker.name).toEqual(workerName);
-      expect(worker.bindings).toBeDefined();
-    } finally {
-      await destroy(scope);
-      await assertWorkerDoesNotExist(workerName);
+  const doBindingTests = testBothPlatforms(
+    [false, true],
+    "create and delete worker with Durable Object binding",
+    async (scope, isWFP: boolean) => {
+      const workerName = `${BRANCH_PREFIX}-test-worker-do-binding-${isWFP ? 'wfp' : 'std'}-1`;
```
sam-goodwin: It would be better if you passed in platform: "vanilla" | "wfp" so the tests didn't need to always have a ternary and could splice it into the string verbatim.

alchemy/test/util.ts
```diff
@@ -33,3 +33,24 @@ function sanitizeForAwsResourceName(str: string): string {
...
+  platforms: Array<"vanilla" | "wfp">,
+  testName: string,
+  testHandler: (scope: any, platform: "vanilla" | "wfp", ...args: T) => Promise<void>,
+): Array<{ name: string; handler: (scope: any, ...args: T) => Promise<void> }> {
+  return platforms.map((platform) => ({
+    name: platform === "wfp" ? `${testName} (${platform})` : testName,
+    handler: async (scope: any, ...args: T) => {
+      return testHandler(scope, platform, ...args);
+    },
+  }));
```
sam-goodwin: I didn't realize this was a generic function. Rename testBothPlatforms to testVariants 

Capture the literal string types from the array passed in. Use T[number] for the type of the variant arg. 

This function is entirely general purpose. 

In fact it could be test.variant on the alchemy/src/test/vitest.ts

alchemy/src/test/vitest.ts
```diff
@@ -79,6 +79,19 @@ type test = {
...
   skipIf(condition: boolean): test;
 
+  /**
+   * Run the same test for multiple variants
+   * @param variants Array of variant values to test
+   * @param testName Base name for the test (will be suffixed with variant for non-first items)
+   * @param testHandler Function that receives scope and variant value
+   * @returns Array of test functions that can be called
+   */
+  variant<T extends readonly string[]>(
```
sam-goodwin: Should be 
```
const T extends readonly string[]
```
That was you don't need to explicitly cast as const when using

alchemy/test/cloudflare/durable-object.test.ts
```diff
@@ -60,112 +65,140 @@ async function assertWorkerDoesNotExist(workerName: string) {
...
-      worker = await Worker(workerName, {
-        name: workerName,
-        script: durableObjectWorkerScript,
-        format: "esm",
-        bindings: {
-          COUNTER: counterNamespace,
-        },
-      });
+  const doBindingTests = test.variant(
+    ["vanilla", "wfp"] as const,
```
sam-goodwin: Remove the as const from this and other usages of test.variant

alchemy/src/test/vitest.ts
sam-goodwin: Also rename to variants

sam-goodwin (commented):
@claude

PR #352, opened 2025-06-11
feat: implement delayed scope finalization with LIFO order

### LINKED ISSUE #348: Replace Resource
We need to add support for replacing a resource - called a REPLACEMENT (or REPLACE operation).

Resources can't always be updated. Some properties, when changed, can only be fulfilled by creating a new Resource and deleting the old one.

I already took a stab at implementing this but I think I took it on in too big of a chunk and now it's very far behind main: https://github.com/sam-goodwin/alchemy/pull/131. You should make sure to analyze the changes in this PR and use it to inform implementation.

To achieve this, let's break it down into the following steps and work on them one PR at a time:
- [ ] Scope finalization logic needs to be changed. Right now, a scope finalizes as soon as the scope function returns. E.g. a resource lifecycle operation completes and then finalize is called. When replacing a resource, we need to delay ALL scope finalizers until the final `app.finalize()`
- [ ] We need to update State to store a record of a Replaced resource and a way to guarantee idempotency in the case where alchemy crashes or partially fails. I suspect it's best to write to the same file, but it could be tricky to get this right. It's basically a list of replaced resource per scope that need to be differentiated from live resources and are then attempted to be cleaned up during scope finalization. When finalizing a scope, look up replaced resources, delete them. Then delete any orphans in the scope.

sam-goodwin:
Implements step 1 of the resource replacement feature by delaying scope finalization until root app.finalize() with proper LIFO execution order.

Closes #348

### DIFF
--- a/alchemy/src/scope.ts
+++ b/alchemy/src/scope.ts
@@ -72,12 +72,6 @@ export class Scope {
   private startedAt = performance.now();
 
   private deferred: (() => Promise<any>)[] = [];
-  
-  /**
-   * Collection of scope finalizers to be executed in LIFO order.
-   * Only used by the root scope to accumulate finalizers from all child scopes.
-   */
-  private scopeFinalizers: Array<() => Promise<void>> = [];
 
   constructor(options: ScopeOptions) {
     this.appName = options.appName;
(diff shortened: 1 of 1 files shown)


alchemy/src/scope.ts
```diff
@@ -199,9 +207,38 @@ export class Scope {
...
+    
+    // If this is not the root scope, register this scope's finalizer with the root
+    // to be executed later in LIFO order
+    if (this.parent !== undefined) {
+      const root = this.root;
+      root.scopeFinalizers.push(async () => {
+        await this.executeFinalization();
+      });
+      return;
+    }
```
sam-goodwin: You have done it in a very hacky way. The goal wasn't to override finalize to work this way. The goal is to call finalize in the right order.

sam-goodwin: All you need to do is have the root finalize call kick off the backwards pass

sam-goodwin (commented):
@claude

PR #366, opened 2025-06-12
feat: update logging format in cli.ts to match vite

sam-goodwin:
Logging prefixes in `alchemy/src/apply.ts` and `alchemy/src/destroy.ts` were converted to lowercase for improved consistency and readability.

*   `SETUP` was changed to `setup`.
*   `SKIPPED` was changed to `skipped`.
*   `CREATING` and `UPDATING` were changed to `creating` and `updating` respectively.
*   `CREATED` and `UPDATED` were changed to `created` and `updated`.
*   `DELETING` was changed to `deleting` and its color updated to `redBright`.
*   `DELETED` was changed to `deleted`.

In `alchemy/src/util/cli.ts`, a `getTimestamp()` function was introduced to format the current time as `HH:MM:SS`. The `task` logger was updated to prepend this timestamp, colored gray, to all log messages. This aligns the logging output with common development tools like Vite, providing immediate temporal context and enhancing visual clarity. The prefix padding logic was also adjusted to accommodate the new timestamp.

### DIFF
--- a/alchemy/src/apply.ts
+++ b/alchemy/src/apply.ts
@@ -42,7 +42,7 @@ async function _apply<Out extends Resource>(
   const start = performance.now();
   try {
     logger.task(resource[ResourceFQN], {
-      prefix: "SETUP",
+      prefix: "setup",
       prefixColor: "cyanBright",
       resource: formatFQN(resource[ResourceFQN]),
       message: "Setting up Resource...",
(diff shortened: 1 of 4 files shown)


sam-goodwin:
@claude actually remove the timestamp prefix, find where this log is and remove it:

[SETUP]     cloudflare-redwood-website > assets Setting up Resource...

There is no such phase as SETUP

PR #371, opened 2025-06-12
fix: move @cloudflare/workers-types to dependencies

sam-goodwin:
Fixes implicit dependency issue in `bound.ts` where @cloudflare/workers-types was only in devDependencies, causing types to become 'any' for consumers.

Now consumers of the alchemy package will automatically get the required Cloudflare types when they install it.

### DIFF
--- a/alchemy/package.json
+++ b/alchemy/package.json
@@ -108,6 +108,7 @@
     }
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


sam-goodwin:
@claude you also need to import it in bound.ts

Issue #369, opened 2025-06-12
alchemy create CLI


sam-goodwin:
@claude implement an alchemy CLI for creating and initializing projects.

Put it in ./alchemy/src/cli.ts

It should:
1. Detect the package manager that it is run with (e.g. npx, bunx, pnpm, yarn) and adjust the template accordingly (e.g. install dependencies with that package manager)
2. Rely on the underlying init scripts already established. E.g. for a simpel typescript app with bun, use `bun init -y`. 
3. Create `alchemy.run.ts` with the appropriate alchemy resources and initialization. Make sure it includes the <reference>

E..g something like this:
```ts
/// <reference types="@types/node" />

import alchemy from "alchemy";
import { Worker } from "alchemy/cloudflare";

const app = await alchemy("<project-name>")

export const worker = await Worker("aegis-worker", {
  entrypoint: "./src/worker.ts",
});

console.log(worker.url);
```
4. Expose templates for "typescript" and each of the web frameworks documented in ./alchemy-web/docs/guides and seen in ./examples
5. You'll notice the guides run cloudflare cli to initialize the project. Make sure to use that and then layer on top edits to files or addition of new files.
6. Update package.json to expose the CLI so that `bunx alchemy` will run this CLI.
7. For publishing, run esbuild on the `cli.ts` entrypoint to create a dependency-less bundle and make sure that .js script is used when executing the alchemy CLI. Any dependencies you need for the TUI or create scripts should only be dev dependencies. (e.g. ink, ink-spinner, react, etc.)

PR #376, opened 2025-06-12
feat: implement alchemy create CLI

sam-goodwin:
Implements the alchemy CLI for creating and initializing projects as requested in issue #369.

## Features
- Package manager detection (bun, npm, pnpm, yarn)
- Interactive prompts for project setup
- Template system for TypeScript, Vite, Astro, React Router, SvelteKit
- Integration with existing Cloudflare CLI workflows
- Dependency-less bundling with esbuild
- Proper alchemy.run.ts generation


src/cli.ts
```diff
@@ -0,0 +1,640 @@
...
+  execCommand(commands.init, projectPath);
+  
+  // Install dependencies
+  execCommand(`${commands.addDev} alchemy @cloudflare/workers-types @types/node typescript`, projectPath);
+  
+  // Create basic project structure
+  mkdirSync(join(projectPath, "src"), { recursive: true });
+  
+  // Create worker.ts
+  writeFileSync(join(projectPath, "src", "worker.ts"), `export default {
```
sam-goodwin: Don't use synchronous io. Always async.

sam-goodwin (commented):
@claude implement test in alchemy/test/create.test.ts that runs each of the templates in a clean directory and ensures they build and deploy. 

Also make sure that there is a way to provide all answers as cli args so non-interactive is possible.

alchemy/test/create.test.ts
```diff
@@ -0,0 +1,50 @@
...
+  });
+
+  test("CLI validates template name", () => {
+    expect(() => {
+      execSync("bun ../../src/cli.ts --name=testproject --template=invalid-template --yes", {
+        encoding: "utf-8",
+        cwd: path.join(process.cwd(), "alchemy", "test")
+      });
+    }).toThrow();
+  });
```
sam-goodwin: None of tests are useful. 

You actually need to loop over a set of cli argument variations, run the cli with those arguments, cd into the directly and run deploy and then destroy. Similar to smoke.test.ts

```
const variants = {
  "astro": "--template astro"
}
```

For each variant in variants, it(variant name, ..)

sam-goodwin (commented):
@claude please add the proper tests

PR #385, opened 2025-06-14
feat(cloudflare): add SecretKey binding support for Cloudflare Workers

### LINKED ISSUE #384: SecretKey Binding
@claude see https://developers.cloudflare.com/api/resources/workers/subresources/scripts/methods/update/

Please create a SecretKey class that is just a data class with a type property "secret_key"

Add support to bind it in worker-metadata and bound.ts

Secret key binding interface 
SecretKey = {
algorithm: unknown
Algorithm-specific key parameters. Learn more.
format: "raw" | "pkcs8" | "spki" | 1 more...
Data format of the key. Learn more.
name: string
A JavaScript variable name for the binding.
type: "secret_key"
The kind of resource that the binding provides.
usages: Array<"encrypt" | "decrypt" | "sign" | 5 more...>
Allowed operations with the key. Learn more.
key_base64: stringOPTIONAL
Base64-encoded key data. Required if format is "raw", "pkcs8", or "spki".
key_jwk: unknownOPTIONAL
Key data in JSON Web Key format. Required if format is "jwk".
}

sam-goodwin:
Closes #384
```typescript
import { SecretKey } from "alchemy/cloudflare";
import { secret } from "alchemy";

// Generate an RSA key pair
const keyPair = await crypto.subtle.generateKey(
  {
    name: "RSA-PSS",
    modulusLength: 2048,
    publicExponent: new Uint8Array([1, 0, 1]),
    hash: "SHA-256",
  },
  true, // extractable
  ["sign", "verify"]
);

// Export the private key as JWK
const privateKeyJwk = await crypto.subtle.exportKey("jwk", keyPair.privateKey);

const signingKey = new SecretKey({
  algorithm: { name: "RSA-PSS", hash: "SHA-256" },
  format: "jwk",
  usages: ["sign"],
  key_jwk: secret(privateKeyJwk),
});

  entrypoint: "./worker.ts",
  bindings: {
    SIGNING_KEY: signingKey,
  },
});
```

**worker.ts**

```javascript
export default {
  async fetch(request, env) {
    const data = new TextEncoder().encode("Hello, World!");

    const signature = await crypto.subtle.sign(
      { name: "RSA-PSS", saltLength: 32 },
      env.SIGNING_KEY,
      data
    );

    return Response.json({
      signed: Array.from(new Uint8Array(signature)),
    });
  },
};
```

### DIFF
--- a/alchemy/src/cloudflare/secret-key.ts
+++ b/alchemy/src/cloudflare/secret-key.ts
@@ -2,10 +2,80 @@
  * SecretKey binding for Cloudflare Workers
  *
  * A data class that represents a secret key binding with cryptographic properties.
- * Based on the Cloudflare Workers API specification for secret keys.
+ * Keys are securely stored and exposed as CryptoKey objects for use with the Web Crypto API.
+ *
+ * @example
+ * // Create an RSA signing key from a JWK format:
+ * const keyPair = await crypto.subtle.generateKey(
+ *   {
+ *     name: "RSA-PSS",
+ *     modulusLength: 2048,
+ *     publicExponent: new Uint8Array([1, 0, 1]),
+ *     hash: "SHA-256",
+ *   },
+ *   true,
+ *   ["sign", "verify"]
+ * );
+ *
+ * const privateKeyJwk = await crypto.subtle.exportKey("jwk", keyPair.privateKey);
+ *
+ * const signingKey = new SecretKey({
+ *   algorithm: { name: "RSA-PSS", hash: "SHA-256" },
+ *   format: "jwk",
+ *   usages: ["sign"],
+ *   key_jwk: secret(privateKeyJwk),
+ * });
+ *
+ * @example
+ * // Create an AES encryption key from raw bytes:
+ * const aesKey = await crypto.subtle.generateKey(
+ *   { name: "AES-GCM", length: 256 },
+ *   true,
+ *   ["encrypt", "decrypt"]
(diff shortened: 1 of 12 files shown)


alchemy/src/cloudflare/secret-key.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+   * Base64-encoded key data. Required if format is "raw", "pkcs8", or "spki"
+   */
+  key_base64?: string;
+
+  /**
+   * Key data in JSON Web Key format. Required if format is "jwk"
+   */
+  key_jwk?: unknown;
```
sam-goodwin: These should be Secret (alchemy core Secret) since they are sensitive.

alchemy/src/cloudflare/secret-key.ts
```diff
@@ -0,0 +1,100 @@
...
+  | "unwrapKey";
+
+/**
+ * SecretKey binding properties
+ */
+export interface SecretKeyProps {
+  /**
+   * Algorithm-specific key parameters
+   */
+  algorithm: unknown;
```
sam-goodwin, line 105: Can we look at https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/importKey#algorithm and define a union type of different types. 

Use a rune `| (unknown @ {}}` if we need to keep it open while also keeping the auto complete type hints.

PR #393, opened 2025-06-16
feat: allow binding individual secrets

tvanhens:


alchemy/src/cloudflare/secret.ts
```diff
@@ -209,24 +215,98 @@ export async function insertSecret(
...
+        `Error updating secret '${secretName}': ${updateErrorMessage}`,
+      );
+    }
+  } else {
+    // Some other error occurred during creation
+    const createErrorMessage =
+      createErrorData.errors?.[0]?.message || createResponse.statusText;
+    throw new Error(
+      `Error creating secret '${secretName}': ${createErrorMessage}`,
+    );
```
sam-goodwin, line 276: Call handleApiError instead of doing this manually each time.

sam-goodwin:
@claude please open another PR that deletes secrets-store.test.ts and re-writes secret.test.ts to adopt the default secrets store with delete: false (instead of creating them because cloudflare only allows one per account). Just test binding a single secret resource and return it in the worker response to test end to end. Keep it simple.

PR #395, opened 2025-06-16
feat(cloudflare): auto-create default secrets store

sam-goodwin:
```ts
import { Secret } from "alchemy/cloudflare";

const mySecret = await Secret("my-secret", {
  value: alchemy.secret(process.env.MY_SECRET),
});
```

> [!TIP]
> This will auto-create a Secrets Store, `default_secrets_store` if one does not exist. Cloudflare's UI does the same.

Then bind the Secret to your Worker:

```ts
export const worker = await Worker("worker", {
  bindings: {
    MY_SECRET: mySecret,
  },
});
l;
```

And use it at runtime:

```ts
import type { worker } from "../alchemy.run.ts";

export default {
  async fetch(request, env: typeof worker.Env) {
    const secret = await env.MY_SECRET.get();

    // ..
  },
};
```


alchemy/test/cloudflare/secret.test.ts
```diff
@@ -1,190 +1,97 @@
...
 
       expect(secretsStore).toBeTruthy();
 
-      // Create a secret in the store
-      secretResource = await Secret(testId, {
+      // Create a test secret in the default store
+      testSecret = await Secret(testId, {
         store: secretsStore,
         value: secret("test-secret-value"),
+        delete: false, // Don't delete the secret when destroyed
```
sam-goodwin: Secret should be delted

alchemy/test/cloudflare/secret.test.ts
```diff
@@ -1,190 +1,97 @@
...
+                });
+              }
+            }
+          };
+        `,
+        format: "esm",
+        url: true,
+        bindings: {
+          SECRETS: secretsStore,
+        },
```
sam-goodwin: Don't bind the secrets store, bind the secret

alchemy/test/cloudflare/secret.test.ts
```diff
@@ -1,190 +1,97 @@
...
+      expect(testSecret.value.unencrypted).toEqual("test-secret-value");
+
+      // Create a worker that binds to the secrets store and returns the secret value
+      worker = await Worker(workerName, {
+        name: workerName,
+        script: `
+          export default {
+            async fetch(request, env, ctx) {
+              try {
+                const secretValue = await env.SECRETS.get("${testId}");
```
sam-goodwin: It should be `await env.TEST_SECRET.get()`

PR #398, opened 2025-06-16
fix: fix secret binding type

tvanhens:


alchemy/src/cloudflare/bound.ts
```diff
@@ -44,9 +44,9 @@ export type Bound<T extends Binding> = T extends _DurableObjectNamespace<
             : T extends _Hyperdrive
               ? Hyperdrive
               : T extends Secret
-                ? string
+                ? SecretsStoreSecret
```
sam-goodwin: @claude This should still be `string`, only CloudflareSecret needs to change

PR #403, opened 2025-06-17
feat(aws): rewrite AWS resources to use aws4fetch + Effect

### LINKED ISSUE #401: Re-write alchemy/src/aws resources to use aws4fetch instead of @aws-sdk/
@claude please go through the resources in alchemy/src/aws (but not alchemy/src/aws/control/*) and change them to use aws4fetch. Take a look at aws/control/* for an example of how we use aws4fetch already.

Wrap aws4fetch in Effect so we can have type-safe errors over raw aws4fetch: https://effect.website/docs/error-management/expected-errors/

sam-goodwin:
Converts AWS resources in `alchemy/src/aws` to use `aws4fetch` instead of `@aws-sdk/`, following the patterns used in the control directory.

## Changes

- Add Effect peer dependency for type-safe error handling
- Create new AWS client wrapper using aws4fetch with Effect
- Convert S3 bucket resource to use aws4fetch instead of @aws-sdk/client-s3
- Convert SQS queue resource to use aws4fetch instead of @aws-sdk/client-sqs
- Convert SSM parameter resource to use aws4fetch instead of @aws-sdk/client-ssm
- Convert account ID utility to use aws4fetch instead of @aws-sdk/client-sts

All conversions maintain identical interfaces and functionality while removing heavy AWS SDK dependencies in favor of lightweight aws4fetch with Effect-based error handling and retry logic.

Fixes #401

🤖 Generated with [Claude Code](https://claude.ai/code)


alchemy/src/aws/bucket.ts
```diff
@@ -130,104 +119,85 @@ export interface Bucket extends Resource<"s3::Bucket">, BucketProps {
...
-            new DeleteBucketCommand({
-              Bucket: props.bucketName,
-            }),
-          ),
-        ),
-      );
+      await ignore(AwsResourceNotFoundError.name, async () => {
+        const deleteEffect = client.delete(`/${props.bucketName}`);
+        await Effect.runPromise(deleteEffect);
+      });
```
sam-goodwin, line 128: @claude Instead of spraying Effect.runPromise everywhere, instead write an effect-version of the whole lifecycle

async function (this: Context<Bucket>, _id: string, props: BucketProps) {

Call it `EffectResource` 

And then do everything in effect. e.g. await ignore(.) is not very effect-friendly. Would prefer to use Effect's declarative flow control features instead of implement our own.

sam-goodwin: @claude Instead of spraying Effect.runPromise everywhere, instead write an effect-version of the whole lifecycle

async function (this: Context, _id: string, props: BucketProps) {

Call it EffectResource

And then do everything in effect. e.g. await ignore(.) is not very effect-friendly. Would prefer to use Effect's declarative flow control features instead of implement our own.

alchemy/src/aws/client.ts
```diff
@@ -0,0 +1,282 @@
...
+      return new AwsAccessDeniedError(message, response, data);
+    }
+    if (response.status === 429 || errorCode.includes("Throttling")) {
+      return new AwsThrottleError(message, response, data);
+    }
+    if (response.status === 400 || errorCode.includes("ValidationException")) {
+      return new AwsValidationError(message, response, data);
+    }
+
+    return new AwsError(message, response, data);
```
sam-goodwin: All errors should have a reference to the error code string

alchemy/src/aws/queue.ts
```diff
@@ -1,16 +1,31 @@
...
+    props: P,
+  ) => Effect.Effect<T, any>,
+) {
+  return Resource(
+    type,
+    async function (this: Context<T>, id: string, props: P): Promise<T> {
+      return Effect.runPromise(effectHandler(this, id, props));
+    },
+  );
+}
```
sam-goodwin: Share this same wrapper across all the resources instead of duplicating

alchemy/src/aws/queue.ts
```diff
@@ -136,211 +151,208 @@ export interface Queue extends Resource<"sqs::Queue">, QueueProps {
...
-            new GetQueueUrlCommand({
-              QueueName: queueName,
-            }),
-          ),
-        );
+  (context, _id, props) =>
+    Effect.gen(function* () {
+      const client = yield* Effect.promise(() =>
+        createAwsClient({ service: "sqs" }),
+      );
```
sam-goodwin: Would prefer the AWS client to be implement with native effect instead of wrapping here

alchemy/src/aws/queue.ts
```diff
@@ -136,211 +151,208 @@ export interface Queue extends Resource<"sqs::Queue">, QueueProps {
...
-        await retry(() =>
-          client.send(
-            new DeleteQueueCommand({
-              QueueUrl: urlResponse.QueueUrl,
-            }),
-          ),
+      // Validate that FIFO queues have .fifo suffix
+      if (props.fifo && !queueName.endsWith(".fifo")) {
+        yield* Effect.fail(
+          new Error("FIFO queue names must end with .fifo suffix"),
```
sam-goodwin: We should expose specific error classes for each of the cases

sam-goodwin (commented):
@claude

alchemy/src/aws/account-id.ts
```diff
@@ -1,15 +1,29 @@
...
 
 export type AccountId = string & {
   readonly __brand: "AccountId";
 };
 
 /**
- * Helper to get the current AWS account ID
+ * Helper to get the current AWS account ID using Effect-based API
+ */
+export function AccountId(): Effect.Effect<AccountId, any> {
```
sam-goodwin: Error should not be any - make sure all errors are type safe

alchemy/src/aws/bucket.ts
```diff
@@ -127,124 +114,111 @@ export interface Bucket extends Resource<"s3::Bucket">, BucketProps {
  *   }
  * });
  */
-export const Bucket = Resource(
+export const Bucket = EffectResource<Bucket, BucketProps>(
   "s3::Bucket",
-  async function (this: Context<Bucket>, _id: string, props: BucketProps) {
-    const client = new S3Client({});
+  (context, _id, props) =>
+    Effect.gen(function* () {
```
sam-goodwin: Why can't we just make this a function*

The EffectResouece can call Effect.gen

alchemy/src/aws/bucket.ts
```diff
@@ -127,124 +114,111 @@ export interface Bucket extends Resource<"s3::Bucket">, BucketProps {
  *   }
  * });
  */
-export const Bucket = Resource(
+export const Bucket = EffectResource<Bucket, BucketProps>(
   "s3::Bucket",
-  async function (this: Context<Bucket>, _id: string, props: BucketProps) {
-    const client = new S3Client({});
+  (context, _id, props) =>
```
sam-goodwin: context should be `this` like Resource is.

sam-goodwin (commented):
@claude

alchemy/src/aws/bucket.ts
```diff
@@ -127,108 +114,89 @@ export interface Bucket extends Resource<"s3::Bucket">, BucketProps {
...
-            new GetBucketVersioningCommand({ Bucket: props.bucketName }),
-          ),
-        ),
-        retry(() =>
-          client.send(new GetBucketAclCommand({ Bucket: props.bucketName })),
-        ),
+      yield* Effect.all([
+        client.get(`/${props.bucketName}?location`),
+        client.get(`/${props.bucketName}?versioning`),
+        client.get(`/${props.bucketName}?acl`),
```
sam-goodwin: The docs suggest we might need a header?

GET /?location HTTP/1.1
Host: Bucket.s3.amazonaws.com

https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketLocation.html

alchemy/src/aws/bucket.ts
```diff
@@ -127,108 +114,89 @@ export interface Bucket extends Resource<"s3::Bucket">, BucketProps {
...
-        );
-        tags = Object.fromEntries(
-          taggingResponse.TagSet?.map(({ Key, Value }) => [Key, Value]) || [],
-        );
-      } catch (error: any) {
-        if (error.name !== "NoSuchTagSet") {
-          throw error;
+      const taggingResponse = yield* client
+        .get(`/${props.bucketName}?tagging`)
+        .pipe(Effect.catchAll(() => Effect.succeed(null)));
```
sam-goodwin: Why are we swallowing errors? We need to focus on how we handle errors

alchemy/src/aws/bucket.ts
```diff
@@ -240,8 +208,12 @@ export const Bucket = Resource(
...
       region,
       hostedZoneId: getHostedZoneId(region),
-      versioningEnabled: versioningResponse.Status === "Enabled",
-      acl: aclResponse.Grants?.[0]?.Permission?.toLowerCase(),
+      versioningEnabled:
+        (versioningResponse as any)?.VersioningConfiguration?.Status ===
+        "Enabled",
+      acl: (
+        aclResponse as any
+      )?.AccessControlPolicy?.AccessControlList?.Grant?.[0]?.Permission?.toLowerCase(),
```
sam-goodwin: Why cast to any? Types should be respected

alchemy/src/aws/effect-resource.ts
```diff
@@ -0,0 +1,39 @@
...
+    type,
+    async function (this: Context<T>, id: string, props: P): Promise<T> {
+      const result = await Effect.runPromise(
+        Effect.gen(effectHandler.bind(this, id, props)),
+      );
+
+      // Handle the delete case where effectHandler returns null
+      if (result === null) {
+        return this.destroy();
+      }
```
sam-goodwin: Why does null mean this? 

Is it because destroy throws? We should pass an Effect wrapper of Context then and have the effect yield it

sam-goodwin (commented):
@claude

sam-goodwin:
@claude you forgot to push the change. Please readdress feedback.

alchemy/src/aws/client.ts
```diff
@@ -0,0 +1,308 @@
...
+  }
+}
+
+export class AwsNetworkError extends AwsError {}
+export class AwsThrottleError extends AwsError {}
+export class AwsResourceNotFoundError extends AwsError {}
+export class AwsAccessDeniedError extends AwsError {}
+export class AwsValidationError extends AwsError {}
+export class AwsConflictError extends AwsError {}
+export class AwsInternalServerError extends AwsError {}
```
sam-goodwin: Can we use Data.TaggedError instead?

https://effect.website/docs/error-management/yieldable-errors/#datataggederror

Then use catchTag to handle errors.

alchemy/src/aws/bucket.ts
```diff
@@ -127,108 +143,98 @@ export interface Bucket extends Resource<"s3::Bucket">, BucketProps {
...
-          client.send(
-            new GetBucketTaggingCommand({ Bucket: props.bucketName }),
+      const taggingResponse = yield* client
+        .get<S3TaggingResponse>(`/${props.bucketName}?tagging`)
+        .pipe(
+          Effect.catchSome(
+            (err) =>
+              err instanceof AwsResourceNotFoundError
+                ? Effect.succeed(null) // Tags don't exist - OK
+                : Effect.fail(err), // Other errors should bubble up
```
sam-goodwin: Not sure why resource not found is succeeds or whether it would be thrown anyway?

Is ResourceNotFound misleading and it should just be NotFound?

AMar4enko:
Hi Sam. There is a PR open here https://github.com/floydspace/effect-aws/pull/114 adding SignatureV4 interop to effect HttpClient with 2 signature implementations - smithy and aws4fetch. With something similar in place you could stick to regular effect HttpClient with all the goodness that comes with it.

sam-goodwin:
Thanks for the tip, @AMar4enko 

Does it use the AWS SDK v3 underneath or is it raw fetch ?

I found an effect library that wraps AWS SDK v3 but was hoping to also drop the clunky dependencies.

AMar4enko:
@sam-goodwin well effect-aws covers multiple scenarios, primary being indeed to lift AWS SDK v3 primitives into Effect-friendly shape, and it's good ol' aws sdk under the blanket ofcourse.
My point about Signature 4 is to use Effect native HttpClient instead of aws4fetch client. I suppose you use it to send authenticated requests to AWS services using credentials, don't you?

P.S. I'll see if I can hack together a PR replacing aws4fetch with Effect HttpClient

sam-goodwin:
@AMar4enko if it's just a replacement for aws4fetch and the signing process without bringing in the entire SDK, that is precisely what i want.

Taking it one step further, I'd like to find a way to do a type-only representation of the entire AWS SDK like I did here: https://github.com/sam-goodwin/itty-aws. But for Effect (so error types also inferred). 

If that's not possible, AI-generated wrappers around services on top of Effect's HttpClient would be 👌

PR #405, opened 2025-06-17
feat(cli): complete cli overhaul with trpc-cli, zod, and clack/prompts

AmanVarshney01:


alchemy/package.json
```diff
@@ -114,6 +115,7 @@
   },
   "dependencies": {
     "@aws-sdk/credential-providers": "^3.0.0",
+    "@clack/prompts": "^0.11.0",
```
sam-goodwin: CLI dependencies should be in devDependencies and bundled in unless we are also using them in the core lib

alchemy/tsdown.config.ts
```diff
@@ -0,0 +1,13 @@
+import { defineConfig } from "tsdown";
+
+export default defineConfig({
+  entry: ["bin/alchemy.ts"],
+  format: ["esm"],
+  clean: false,
+  shims: true,
+  outDir: "bin",
+  external: ["jsonc-parser"],
```
sam-goodwin: I ran into a problem with that too, fixed it with:

```sh
#!/usr/bin/env sh

esbuild bin/alchemy.ts \
  --bundle \
  --platform=node \
  --target=node20 \
  --outfile=bin/alchemy.mjs \
  --format=esm \
  --external:node:* \
  --main-fields=module,main \
  --banner:js="import { createRequire as __createRequire } from 'node:module'; const require = __createRequire(import.meta.url);"
```

Specificlaly:
```sh
  --main-fields=module,main \
```

sam-goodwin: Are we sure we can't bundle it in?

sam-goodwin (commented):
Looking so good!

alchemy/tsdown.config.ts
AmanVarshney01: I am not sure how to fix :(

sam-goodwin:
```ts
resolve: {
  mainFields: ['module', 'main']
}

```
https://rolldown.rs/reference/config-options#resolve-mainfiles

alchemy/bin/templates/nuxt.ts
```diff
@@ -0,0 +1,76 @@
...
+</style>
+`,
+      );
+    } catch (error) {
+      const errorMsg = error instanceof Error ? error.message : String(error);
+      throw new Error(`Failed to update app.vue: ${errorMsg}`);
+    }
+  } catch (error) {
+    const errorMsg = error instanceof Error ? error.message : String(error);
+    throw new Error(`Nuxt template initialization failed: ${errorMsg}`);
```
sam-goodwin: We should set the cause to maintain the stack trace https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause

alchemy/bin/templates/redwood.ts
```diff
@@ -0,0 +1,45 @@
...
+`,
+      );
+    } catch (error) {
+      const errorMsg = error instanceof Error ? error.message : String(error);
+      throw new Error(`Failed to create redwood.toml: ${errorMsg}`);
+    }
+  } catch (error) {
+    const errorMsg = error instanceof Error ? error.message : String(error);
+    throw new Error(`Redwood template initialization failed: ${errorMsg}`);
+  }
```
sam-goodwin: Seeing this boilerplate everywhere. Anyway to simplify with a helper?

AmanVarshney01: `
export function throwWithContext(error: unknown, context: string): never {
  const errorMsg = error instanceof Error ? error.message : String(error);
  throw new Error(`${context}: ${errorMsg}`, {
    cause: error instanceof Error ? error : new Error(String(error)),
  });
}
`

added the helper function

alchemy/templates/typescript/alchemy.run.js
sam-goodwin, line 1: This file shouldn't be checked in

biome.json
```diff
@@ -20,6 +20,7 @@
       "./stacks/**/*.ts",
       "./examples/*/src/**/*.ts",
       "./examples/*/*.ts",
+      "!alchemy/templates/**",
```
sam-goodwin: why not include this? Good for formatting and linting?

PR #407, opened 2025-06-17
fix(cloudflare): fix SSR for astro by setting html_handling=none

sam-goodwin:


just-be-dev:
https://github.com/sam-goodwin/alchemy/pull/407/files#diff-a3c5367780f9574f32d89e9a1be82390995077711c524529541466c91ab075bdL80

Slight update here. You probably want to use `path.join` on all the path parts.

PR #417, opened 2025-06-19
Support replacing resources

Mkassabov:
## What this does
- supports replacing resources
- based on [PR#131](https://github.com/sam-goodwin/alchemy/pull/131)


alchemy/src/apply.ts
```diff
@@ -179,6 +181,22 @@ async function _apply<Out extends Resource>(
...
+          // For now, we just disallow it.
+          throw new Error(
+            `Resource "${resource[ResourceFQN]}" has children and cannot be replaced.`,
+          );
+        }
+        if (state.replace !== undefined) {
+          throw new Error(
+            `Resource "${resource[ResourceFQN]}" has pending replaced resource that must be deleted first.`,
+          );
+        }
```
sam-goodwin: Why do we need this? Replace happens during scope finalization. This should just be idempotent and re-run.

alchemy/src/apply.ts
```diff
@@ -212,6 +230,14 @@ async function _apply<Out extends Resource>(
       output,
       props,
       // deps: [...deps],
+      // TOOD(sam): what happens if we fail to set the state here?
+      // next pass, we've lost track of the replacement resource
+      replace: isReplaced
```
sam-goodwin: That's true. Should `this.replace` be async and `await this.replace()` guarantees persistence?

alchemy/src/destroy.ts
```diff
@@ -18,6 +20,14 @@ export class DestroyedSignal extends Error {}
...
   quiet?: boolean;
   strategy?: "sequential" | "parallel";
+  /**
+   * This destroy operation is for a replaced resource
+   *
+   * The state of the resource should not be unpersisted.
+   */
+  replace?: {
+    props: ResourceProps | undefined;
+  };
```
sam-goodwin: You will also need `output` (the output attributes of the old resource) which is required when you call delete

alchemy/src/destroy.ts
```diff
@@ -41,14 +51,29 @@ export async function destroy<Type extends string>(
...
+
+      const resources = (await load(scope.resources.keys())).map(
+        (resource) => ({
+          seq: resource.seq,
+          state: resource,
+          type: "orphan",
+        }),
+      );
+
+      await destroyAll(resources, options);
```
sam-goodwin: Can you explain why this needed to change? Looks out of place.

alchemy/src/destroy.ts
```diff
@@ -164,15 +204,58 @@ export async function destroy<Type extends string>(
...
+  )[],
   options?: DestroyOptions,
+  force?: boolean,
 ) {
   if (options?.strategy !== "parallel") {
-    const sorted = resources.sort((a, b) => b[ResourceSeq] - a[ResourceSeq]);
-    for (const resource of sorted) {
-      await destroy(resource, options);
+    for (const node of sequence) {
+      if (node instanceof Scope) {
```
sam-goodwin: use `isScope` instead since `instanceof` breaks regularly

alchemy/src/scope.ts
```diff
@@ -223,17 +240,63 @@ export class Scope {
...
+            .filter(
+              (r) => r.replace !== undefined && (this.parent == null || force),
+            )
+            .map((state) => ({
+              type: "replaced",
+              seq: state.seq,
+              state,
+            })),
+        ),
+        // TODO(sam): what if an orphan also has replace !== undefined? We need to delete it too (2 deletions).
```
sam-goodwin: That sounds right.

alchemy/src/apply.ts
Mkassabov: children check is still there, pending replaced has been fixed

alchemy/src/destroy.ts
Mkassabov: nah, its gone, I don't want to think about it

alchemy/src/scope.ts
Mkassabov: done

alchemy/src/destroy.ts
Mkassabov: can you elaborate?

alchemy/src/scope.ts
```diff
@@ -179,6 +193,69 @@ export class Scope {
...
+        return await this.parent.state.set(this.scopeName, scopeState);
+      }
+    }
+    //todo(michael): ugly hack to support state in the root
+    const rootState =
+      (await this.state.get(this.chain.join("-"))) ?? this.createRootState();
+    delete rootState.data[key];
+    await this.state.set(this.chain.join("-"), rootState);
+  }
+
```
Mkassabov: @sam-goodwin 
2 things about this
1. I renamed `get` and `delete` to `getScope` and `deleteResource` that way we could use `get` and `delete` for state, in parity with how we deal with state in the resources. is that okay, or do we want to call these soemthing like `getState`?
2. any ideas how to handle state for the root scope, not a big fan of the current solution

alchemy/src/scope.ts
```diff
@@ -250,6 +338,26 @@ export class Scope {
...
   }
 
+  public async destroyPendingDeletions() {
+    const pendingDeletions =
+      (await this.get<PendingDeletions>("pendingDeletions")) ?? [];
+    if (pendingDeletions) {
+      for (const { resource, oldProps } of pendingDeletions) {
+        //todo(michael): ugly hack due to the way scope is serialized
+        const realResource = this.resources.get(resource[ResourceID])!;
+        resource[ResourceScope] = realResource?.[ResourceScope] ?? this;
```
Mkassabov, line 375: @sam-goodwin 
is there something better I can do here? we don't serialize scope, and we technically work with these in a different scope than where they were created.

the result is that we can't properly restore scope; so we have this hack to pull the correct scope from the new resource and reassign it.

is there a reason we don't serialize scope?

alchemy/src/scope.ts
sam-goodwin: > I renamed get and delete to getScope and deleteResource that way we could use get and delete for state, in parity with how we deal with state in the resources. is that okay, or do we want to call these soemthing like getState?

Seems reasonable to me. I've seen one person make the mistake of thinking it's an arbitrary key-value store, so this is an improvement.

> Any ideas how to handle state for the root scope, not a big fan of the current solution
Oh good question ... i had not considered that. First thing that comes to mind is write a special file with characters that are illegal in a resource name (`:` i think is illegal, for example)

Mkassabov: `:` is not valid in names on windows
already had to change the test runner to get it to run

sam-goodwin: Oh by the way, root scope should never have state, since there is ALWAYS a stage scope underneath

Mkassabov: got it, will remove. wasn't sure

Mkassabov: done :)

sam-goodwin:
Some tests are failing:
```sh
 FAIL  alchemy/test/cloudflare/worker.test.ts > Worker Resource > create, update, and delete worker using entrypoint file
Error: Resource "samgoodwin-worker.test.ts/Worker Resource/create, update, and delete worker using entrypoint file/samgoodwin-test-worker-entrypoint" has children and cannot be replaced.
 ❯ Scope.<anonymous> alchemy/src/apply.ts:203:19
    201|             // ... that then conflict with the replaced resource's children and are either deleted or …
    202|             // For now, we just disallow it.
    203|             throw new Error(
       |                   ^
    204|               `Resource "${resource[ResourceFQN]}" has children and cannot be replaced.`,
    205|             );
 ❯ AsyncFunction.run alchemy/src/alchemy.ts:476:12
 ❯ _apply alchemy/src/apply.ts:189:16
 ❯ alchemy/test/cloudflare/worker.test.ts:927:16
 ❯ Scope.<anonymous> alchemy/src/test/vitest.ts:208:13
 ❯ AsyncFunction.run alchemy/src/alchemy.ts:476:12
```

```sh
bun vitest run ./alchemy/test/cloudflare/worker.test.ts -t "create, update, and delete worker using entrypoint file"
```

Tests failing with `"has children and cannot be replaced."`:
* `worker.test.ts`
  * `create, update, and delete worker using entrypoint file` 
  * `create and delete worker with multiple bindings`
* `route.test.ts`
  * `update worker routes`
* `account-api-token.test.ts`
  * `create, update, and delete token`

alchemy/src/destroy.ts
```diff
@@ -108,10 +108,9 @@ export async function destroy<Type extends string>(
       return;
     }
 
-    const pendingDeletions =
-      await state.output[ResourceScope].get<PendingDeletions>(
-        "pendingDeletions",
-      );
+    const pendingDeletions = await state.output[ResourceScope]
+      .get<PendingDeletions>("pendingDeletions")
+      .catch(() => []);
```
sam-goodwin: my worry with this is what if `state.get` fails for a transient reason, will we assume it doesn't have any?

Mkassabov: fair point. will replace with a specific error and check against that.

alchemy/src/apply.ts
```diff
@@ -161,34 +164,67 @@ async function _apply<Out extends Resource>(
...
+          return provider.handler.bind(ctx)(resource[ResourceID], props);
+        },
+      );
+    } catch (error) {
+      if (error instanceof ReplacedSignal) {
+        ctx.phase = "create";
+
+        if (
+          Scope.current.children.get(resource[ResourceID])?.children.size! > 0
+        ) {
```
sam-goodwin: Might be safer to just reference `scope` instead of `Scope.current`

Mkassabov: fixed!

alchemy/src/apply.ts
```diff
@@ -161,34 +164,67 @@ async function _apply<Out extends Resource>(
...
+      replace: async () => {
         if (isReplaced) {
           logger.warn(
             `Resource ${resource[ResourceKind]} ${resource[ResourceFQN]} is already marked as REPLACE`,
           );
           return;
         }
+
         isReplaced = true;
+        throw new ReplacedSignal();
```
sam-goodwin: should this error if `ctx.phase === "create"`?

alchemy/src/apply.ts
```diff
@@ -161,34 +164,67 @@ async function _apply<Out extends Resource>(
...
+          throw new Error(
+            `Resource ${resource[ResourceFQN]} has children and cannot be replaced.`,
+          );
+        }
+
+        output = await alchemy.run(
+          resource[ResourceID],
+          { isResource: true, parent: scope },
+          async () => provider.handler.bind(ctx)(resource[ResourceID], props),
+        );
```
sam-goodwin: What is this calls replace again? It should be in create phase and error?

alchemy/src/apply.ts
sam-goodwin: Can probably also link to an issue tracking enhancing REPLACE to support children

alchemy/src/context.ts
```diff
@@ -53,7 +53,7 @@ export interface BaseContext<Out extends Resource> {
    * Indicate that this resource is being replaced.
    * This will cause the resource to be deleted at the end of the stack's CREATE phase.
    */
-  replace(): void;
+  replace(): Promise<void>;
```
sam-goodwin: `Promise<never>`?

alchemy/src/apply.ts
Mkassabov: we change ctx.phase = "create" on line 195 above it.
seemed simpler than creating an entire new context, no?

alchemy/src/context.ts
sam-goodwin: Or just `never` - the async write to state can be done outside of the replace handler but still durably before calling `create` for the new resource

Mkassabov: needed when I was doing more in replace. will remove

alchemy/src/apply.ts
Mkassabov: what about create and adopt:true at the same time. I feel like that would cause other issues

alchemy/src/destroy.ts
```diff
@@ -102,14 +113,35 @@ export async function destroy<Type extends string>(
...
+    );
+
+    if (options?.replace == null && pendingDeletion !== undefined) {
+      await destroy(pendingDeletion.resource, {
+        replace: {
+          props: pendingDeletion.oldProps,
+          output: pendingDeletion.resource,
+        },
+      });
+    }
```
sam-goodwin: What is this special handling for? Can destroy be generic and not specialized? Why do we need `options?.replace`?

alchemy/src/context.ts
Mkassabov: done

alchemy/src/destroy.ts
Mkassabov: 1. moving to the error based replacement means we can probably remove this actually.
2. not sure I understand, could you elaborate?
3. how else would we destroy the old resource if we don't have a reference to the old props and the old output?

alchemy/src/scope.ts
sam-goodwin: Could we just instantiate `Scope`? I don't think you can rely on the resource always existing to find the real scope. But you can manually create it?

alchemy/src/destroy.ts
Mkassabov: removed the check, but we still need to keep replace around

alchemy/src/scope.ts
Mkassabov: you should be able to rely on the resource always existing
- in replace, the resource should exist or this should fail so we don't delete an active resource
- in delete for a scope, we delete pendingDeletions before anything else
- if you call delete for the scope and pendingDeletion fails, it should prevent deletion of the actual resource as well since it will error before the deletion call of the resource

also if you made a new scope how would you be able to make calls to scope.state?

sam-goodwin: What if we register pending deletions, then crash, then comment out the resource and re-run? The resource won't exist and will be scheduled for deletion.

alchemy/src/apply.ts
sam-goodwin: `this.replace()` has no meaning when creating a resource for the first time?

alchemy/src/apply.ts
sam-goodwin: Right but the `replace` function inside `ctxt` is still operating as if it's an `update` and allowing replacements.

alchemy/src/destroy.ts
sam-goodwin: The main question I have is why does `destroy` need to differentiate between destroying a replaced resource vs destroying an orphaned resource?

alchemy/src/apply.ts
Mkassabov: sure will add an error here

Mkassabov: done

alchemy/src/destroy.ts
Mkassabov: because the deletion handler needs the original props to delete the resource with?

alchemy/src/apply.ts
Mkassabov: true, will give it its own context

Mkassabov: done

alchemy/src/alchemy.ts
```diff
@@ -181,24 +182,34 @@ async function _alchemy(
...
-      Scope.globals.push(root);
+      Scope.globals.push(stage);
     }
     if (mergedOptions?.phase === "destroy") {
-      await destroy(root);
+      await destroy(stage);
       return process.exit(0);
     }
-    return root;
+    return stage;
```
sam-goodwin: If we are now returning the stage, does that mean app.finalize will not be called on the actual root?

alchemy/src/apply.ts
```diff
@@ -162,33 +165,83 @@ async function _apply<Out extends Resource>(
       props: state.oldProps,
       state,
       replace: () => {
+        if (phase === "create") {
+          throw new Error(
+            `Resource ${resource[ResourceKind]} ${resource[ResourceFQN]} cannot be replaced in create phase.`,
+          );
```
sam-goodwin, line 171: "Replace cannot be triggered during the create phase

alchemy/src/apply.ts
```diff
@@ -162,33 +165,83 @@ async function _apply<Out extends Resource>(
...
+                id: resource[ResourceID],
+                fqn: resource[ResourceFQN],
+                seq: resource[ResourceSeq],
+                props: state.props,
+                state,
+                replace: () => {
+                  throw new Error(
+                    `Resource ${resource[ResourceKind]} ${resource[ResourceFQN]} cannot be replaced in create phase.`,
+                  );
+                },
```
sam-goodwin, line 223: nit: spread ?

...ctx,

alchemy/src/destroy.ts
```diff
@@ -18,6 +19,10 @@ export class DestroyedSignal extends Error {}
 export interface DestroyOptions {
   quiet?: boolean;
   strategy?: "sequential" | "parallel";
+  replace?: {
+    props?: ResourceProps | undefined;
```
sam-goodwin, line 23: I'm seeing this compared with `== null`, should that be `undefined`?

alchemy/src/destroy.ts
```diff
@@ -127,10 +133,10 @@ export async function destroy<Type extends string>(
           parent: scope,
         },
         async (scope) => {
-          nestedScope = scope;
+          nestedScope = options?.replace?.props == null ? scope : undefined;
```
sam-goodwin, line 136: Why are we checking null? Does it mean anything different than undefined?

alchemy/src/destroy.ts
```diff
@@ -146,7 +152,21 @@ export async function destroy<Type extends string>(
       await destroy(nestedScope, options);
     }
 
-    await scope.delete(instance[ResourceID]);
+    if (options?.replace == null) {
+      if (nestedScope) {
+        await destroy(nestedScope, options);
+      }
+      await scope.deleteResource(instance[ResourceID]);
```
sam-goodwin, line 159: Not sure what null means and it doesn't seem to be represented in the types

alchemy/src/scope.ts
```diff
@@ -179,6 +202,84 @@ export class Scope {
...
+      }
+      if (this?.parent?.scopeName === this.root.scopeName) {
+        const scopeState =
+          (await this?.parent?.state.get(this.scopeName!)) ??
+          this.createRootState();
+        return scopeState.data[key];
+      }
+      throw new RootScopeStateAttemptError();
+    });
+  }
```
sam-goodwin: Perhaps we can refactor this to remove some of the boilerplate 

```ts

// Inside the same class ──────────────────────────────────────────────────────
/**
 * Centralises the “lock → locate the right scope → hand the caller a live
 * ScopeState instance and a persist() helper”.
 *
 * @param fn   Your operation on the scope state.
 *             • `state` is already resolved and, if we’re at the root, created.
 *             • `persist` will write the (possibly-mutated) state back.
 */
private async withScopeState<R>(
  fn: (
    state: ScopeState,                          // current state for this.scopeName
    persist: (next: ScopeState) => Promise<void> // helper to save changes
  ) => Promise<R>
): Promise<R> {
  return this.dataMutex.lock(async () => {
    // 1. We must know where to look.
    if (!this.parent || !this.scopeName) {
      throw new RootScopeStateAttemptError();
    }

    // 2. Pull (or lazily create) the state bucket we care about.
    const isRoot = this.parent.scopeName === this.root.scopeName;
    const state =
      (await this.parent.state.get(this.scopeName)) ??
      (isRoot ? this.createRootState() : undefined);

    if (!state) throw new RootScopeStateAttemptError();

    // 3. Give the caller the live state + a helper to persist it.
    return fn(state, updated => this.parent!.state.set(this.scopeName!, updated));
  });
}

// ───────────────────────── public facade methods ────────────────────────────
public async set<T>(key: string, value: T): Promise<void> {
  return this.withScopeState<void>(async (state, persist) => {
    state.data[key] = value;
    await persist(state);               // only one line to save!
  });
}

public async get<T>(key: string): Promise<T> {
  return this.withScopeState<T>(async state => state.data[key]);
}

public async delete(key: string): Promise<void> {
  return this.withScopeState<void>(async (state, persist) => {
    delete state.data[key];
    await persist(state);
  });
}

```

alchemy/src/scope.ts
```diff
@@ -250,6 +366,31 @@ export class Scope {
...
+          quiet: this.quiet,
+          strategy: "sequential",
+          replace: {
+            props: oldProps,
+            output: resource,
+          },
+        });
+      }
+    }
+  }
```
sam-goodwin, line 386: nit: leave this in destroy.ts instead of indirection

alchemy/src/apply.ts
Mkassabov: this would omit what resource is causing the problem

alchemy/src/apply.ts
Mkassabov: tried that, gave an ugly ts error
image

alchemy/src/alchemy.ts
Mkassabov: it shouldn't matter because everything is also in stage scope, but we can return root scope, doesn't really matter

alchemy/src/destroy.ts
Mkassabov: in this case, there is no difference
force of habit to do this whenever I want to check null or undef over
`if(value)` which can check falsey values
`if(value === undefined)` or `if(value === null)` which only checks one and can slip pass ts if you're not careful since TS won't error on
```
const test: string | undefined = undefined;
if(test === null) {}
```
even though test will never be null and that will always be false

Mkassabov: LMK if you dislike it and I can stop

alchemy/src/destroy.ts
Mkassabov: see other comment:
https://github.com/sam-goodwin/alchemy/pull/417#discussion_r2160711769

alchemy/src/scope.ts
Mkassabov: agreed, done

alchemy/src/scope.ts
Mkassabov: we can't, since we need this in finalize not just destroy.

alchemy/src/destroy.ts
sam-goodwin: TIL
```
> undefined == null
true
```

PR #423, opened 2025-06-20
fix: improve package manager handling in create-alchemy

nbbaier:
- Adds proper `--` flag handling for `npm` vs other package managers when creating projects with the Astro and React Router templates
- Normalizes whitespace in logged commands
- Update `create` function to accept package manager parameter

### DIFF
--- a/alchemy/bin/alchemy.ts
+++ b/alchemy/bin/alchemy.ts
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 import { parseArgs } from "node:util";
-import { createAlchemy } from "./create-alchemy.ts";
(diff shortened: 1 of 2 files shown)


sam-goodwin:
We have an upcoming re-work of the CLI: https://github.com/sam-goodwin/alchemy/pull/405

@AmanVarshney01 is this something fixed in your PR?

AmanVarshney01:
> We have an upcoming re-work of the CLI: #405
> 
> @AmanVarshney01 is this something fixed in your PR?

Yes, It's fixed. Thanks for the contribution @nbbaier

PR #430, opened 2025-06-20
fix(fsss & fs): better support for windows file system

Mkassabov:
## What this does
- doesn't call create folder if its trying to create `.`
- fixes a colon in file paths in FS-state-stire

### DIFF
--- a/alchemy/src/fs/file.ts
+++ b/alchemy/src/fs/file.ts
@@ -184,9 +184,12 @@ export const File = Resource(
     }
 
     // Create directory and write file
-    await fs.promises.mkdir(path.dirname(filePath), {
-      recursive: true,
-    });
+    const dirName = path.dirname(filePath);
+    if (dirName !== ".") {
+      await fs.promises.mkdir(dirName, {
+        recursive: true,
+      });
+    }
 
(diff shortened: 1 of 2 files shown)


alchemy/src/fs/file.ts
```diff
@@ -184,9 +184,12 @@ export const File = Resource(
...
     // Create directory and write file
-    await fs.promises.mkdir(path.dirname(filePath), {
-      recursive: true,
-    });
+    const dirName = path.dirname(filePath);
+    if (dirName !== ".") {
+      await fs.promises.mkdir(dirName, {
+        recursive: true,
+      });
+    }
```
sam-goodwin, line 192: why is this necessary

Mkassabov: throws an `EEXIST` error on windows. fix for a different windows-specific issue

Mkassabov: image

Mkassabov: funnily enough its an `ENOENT` error if you have recursive true, which somehow makes me feel worse

Mkassabov: image

PR #437, opened 2025-06-20
fix(cloudflare): adopt DO that have migration tags

sam-goodwin:


alchemy/src/cloudflare/worker.ts
```diff
@@ -1322,12 +1322,47 @@ async function putWorkerInternal(
...
+            // because wrangler includes a migration tag and we do not
+            // currently, the only way to discover the old_tag is through the error message
+            // Get Worker Script Settings is meant to return it (according to the docs)
+            // but it doesn't work at runtime
+            //
+            // so, we catch the error and parse out the tag and then retry
+            if (error.message.includes("when expected tag is")) {
+              const newTag = error.message.match(
+                /when expected tag is (.*)/,
+              )?.[1];
```
sam-goodwin: @claude this needs to parse v{number} and not include any quotes.

sam-goodwin (commented):
@claude

PR #441, opened 2025-06-21
feat: SqliteStateStore

pavi2410:
[Todo...]

### DIFF
--- a/alchemy/src/sqlite/sqlite-state-store.ts
+++ b/alchemy/src/sqlite/sqlite-state-store.ts
@@ -1,0 +1,155 @@
+import { DatabaseSync } from "node:sqlite";
+import { mkdirSync } from "node:fs";
+import { dirname } from "node:path";
+import type { Scope } from "../scope.ts";
+import { serialize } from "../serde.ts";
+import { deserializeState, type State, type StateStore } from "../state.ts";
+
+export interface SqliteStateStoreOptions {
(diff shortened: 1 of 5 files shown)


pavi2410:
Todo:
- [ ] Test with Docker example

pavi2410:
I used `node:sqlite` to avoid external dependency. It is supported from Node 22. But, it seems that Bun has not implemented the `node:sqlite` module yet. I am now looking for a replacement SQLite library.

- https://github.com/WiseLibs/better-sqlite3

alchemy/src/sqlite/sqlite-state-store.ts
```diff
@@ -0,0 +1,157 @@
...
+        state_json TEXT NOT NULL,
+        created_at INTEGER NOT NULL DEFAULT (unixepoch()),
+        updated_at INTEGER NOT NULL DEFAULT (unixepoch()),
+        PRIMARY KEY (resource_id, scope_path)
+      );
+
+      CREATE INDEX IF NOT EXISTS idx_alchemy_state_scope_path 
+      ON alchemy_state(scope_path);
+    `);
+  }
```
sam-goodwin, line 50: This can be async. Try and use async io everywhere

sam-goodwin:
This has been implemented:
1. DO: https://github.com/sam-goodwin/alchemy/pull/630
2. Local: https://github.com/sam-goodwin/alchemy/pull/593

PR #445, opened 2025-06-22
feat(cloudflare): relax Durable Object RPC type constraint

### LINKED ISSUE #444: Relax Durable Object RPC type
@claude Currently we require the rpc type extend DurableObjectBrand, let's relax that and intersect it on internally instead of imposing the constraint on the user.

sam-goodwin:
Relaxed the Durable Object RPC type constraint to make the API more user-friendly.

Removes the requirement for RPC types to extend DurableObjectBrand and instead uses intersection types internally to apply branding where needed at runtime.

Fixes #444

### DIFF
--- a/alchemy/src/cloudflare/bound.ts
+++ b/alchemy/src/cloudflare/bound.ts
@@ -36,9 +36,7 @@ type BoundWorker<
 export type Bound<T extends Binding> = T extends _DurableObjectNamespace<
   infer O
 >
-  ? DurableObjectNamespace<
-      O extends Rpc.DurableObjectBranded | undefined ? O : any
-    >
+  ? DurableObjectNamespace<O & Rpc.DurableObjectBranded>
   : T extends { type: "kv_namespace" }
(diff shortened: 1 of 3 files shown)


alchemy/src/cloudflare/bound.ts
```diff
@@ -36,7 +36,9 @@ type BoundWorker<
 export type Bound<T extends Binding> = T extends _DurableObjectNamespace<
   infer O
 >
-  ? DurableObjectNamespace<O>
+  ? DurableObjectNamespace<
+      O extends Rpc.DurableObjectBranded | undefined ? O : any
+    >
```
sam-goodwin: This should just be O & Rpc.DurableObjectBranded

sam-goodwin (commented):
@claude

sam-goodwin:
@claude do the same for other instances of RPC like in Worker WorkerRef and WorkerStub. 

In a separate PR

PR #446, opened 2025-06-22
Use Docker API for managing docker resources

msanchezdev:


alchemy/src/docker/api/network.ts
```diff
@@ -0,0 +1,45 @@
...
+import { createDockerApi } from "./api.ts";
+import type { components } from "./types.ts";
+
+type Network = Required<Pick<components["schemas"]["Network"], "Name">> &
+  Omit<components["schemas"]["Network"], "Name"> &
+  Resource<"docker::with-api::Network">;
+
+export type NetworkProps = Required<Pick<Network, "Name">>;
+
+export const Network = Resource(
```
sam-goodwin, line 162: Good to provide "@example" snippets. Claude does a good job at this

alchemy/src/docker/api/network.ts
```diff
@@ -0,0 +1,45 @@
...
+    if (this.phase === "delete") {
+      try {
+        await docker.network({ id: props.Name }).remove();
+      } catch {
+      } finally {
+        try {
+          await docker.network({ id: props.Name }).inspect();
+        } catch {}
+      }
+      return this.destroy();
```
sam-goodwin: Are you sure you want to swallow errors

scripts/generate-docker-types.ts
```diff
@@ -0,0 +1,63 @@
...
+
+const contractUrl =
+  "https://docs.docker.com/reference/api/engine/version/v1.50.yaml";
+
+const contract = await fetch(
+  `https://converter.swagger.io/api/convert?url=${contractUrl}`,
+).then((res) => res.json());
+
+const nodes = await run(contract);
+await Bun.write("alchemy/src/docker/api/types.ts", astToString(nodes));
```
sam-goodwin, line 53: Can we generate docs too?

alchemy/src/docker/api/api.ts
```diff
@@ -0,0 +1,218 @@
...
+    // @ts-ignore: dyncamically inferred on usage
+    const { data, error, response } = await client.GET(path, ...init);
+
+    if (error) {
+      throw new Error(
+        `Failed to GET ${path}: ${response?.status} ${response?.statusText}: ${JSON.stringify(error, null, 2)}`,
+      );
+    }
+
+    return data as NonNullable<MethodResponse<typeof client, "get", T>>;
```
sam-goodwin: Cool!

alchemy/src/docker/api/api.ts
```diff
@@ -0,0 +1,218 @@
...
+        `Failed to DELETE ${path}: ${response?.status} ${response?.statusText}: ${JSON.stringify(error, null, 2)}`,
+      );
+    }
+
+    return data as NonNullable<MethodResponse<typeof client, "delete", T>>;
+  };
+
+  return {
+    containers: {},
+    network: (nameOrId: PickPathParam<"/networks/{id}", "get", "id">) => ({
```
sam-goodwin: Better to return methods outside of the resource lifecycle handler because functions cannot be serialized.

When a resource is skipped, we don't invoke your function and instead return the persisted state data.

A wrapper function can then implement methods like `network` that use the data

msanchezdev:
@sam-goodwin sorry, I am just getting started, code is not final, will look into generating docs later, first implementing the api client.

alchemy/src/docker/api/_docker-host.ts
```diff
@@ -0,0 +1,57 @@
...
+    this.scope.set(
+      "docker-host",
+      new Dockerode({
+        socketPath,
+        host: url.host,
+        port: url.port,
+        protocol: url.protocol as "http" | "https" | "ssh",
+      }),
+    );
+    return this({});
```
sam-goodwin: I don't think this needs to be in state. Return it as the output instead of empty. You should rarely need to use state like that.

alchemy/test/docker/api/example.ts
```diff
@@ -0,0 +1,60 @@
+import { createDockerApi } from "../../../src/docker/api/index.ts";
```
msanchezdev, line 1: Tests to be implemented

scripts/generate-docker-types.ts
msanchezdev: Not using script right now, will be adding it back later but we can generate docs from it.

msanchezdev:
Docker image Pulling, Pushing and building fully implemented now with convenient overrides when pulling/pushing apart from the base image. (in the example we have the base hello-world:1.0.4 image, which is built from source, if pull was enabled it would first try to pull from the selfhosted registry (already declared in the source code), and when pushing the image would be pushed to Github Container Registry under the same tag (unless overriden) and noticing the repository being specified as the base image hello-world doesn't have the github username).

All properties except for ref are optional. This is just a detailed example.
image

It can be as simple as this
image

Pulling an image and pushing to another registry:
image

Pulling (from a mirror?) and pushing from different registries:
image

msanchezdev:
Docker credential stores are also supported now on a host or per-registry basis. Leveraging credential management to the system keychain's. (By default the credential store set up for the Docker CLI will be used if no DockerHost is used, or read from docker's config.json if the DockerHost is set a config directory) 
image

alchemy/src/util/diff.ts
```diff
@@ -12,3 +13,332 @@ export const diff = <T>(a: T, b: NoInfer<T>) => {
...
+class UnexpectedTokenError extends Error {
+  constructor(token: string | undefined, expected: string) {
+    super(
+      `Unexpected token ${token ? JSON.stringify(token) : "EOF"}, expected ${expected}`,
+    );
+    this.name = "UnexpectedTokenError";
+  }
+}
+
+export function exists<O extends { returnValue?: boolean } = {}>(
```
Mkassabov, line 28: I worry that the changes here are a little outside the scope of what this PR needs.
There is an entire parser here and its just called `diff2` and `exists`. The maintenance burden of this change is a lot and I worry its complicated enough that contributors may not be able to easily understand what this is and when to use it.

alchemy/src/util/duration.ts
Mkassabov, line 1: I'm not sure what part of this is really alchemy specific, would it be possible to move this to one of the existing libraries out there

alchemy/test/docker/v2/image.test.ts
Mkassabov, line 1: This adds to my worries about scope. It feels like most of the image tests are just parser tests

alchemy/test/docker/v2/network.test.ts
```diff
@@ -0,0 +1,25 @@
...
+  // TODO: Network only with IPv6 Enabled
+  // TODO: Network with IPv4 and IPv6 Enabled
+  // TODO: Network with IPv4 and IPv6 Disabled
+  // TODO: Containers are reconnected after update
+  // TODO: Containers are preserved after network is destroyed
+  // TODO: Containers are reconnected if network update failed
+  // TODO: Containers IP addresses are preserved if network updates with no IP invalidating changes
+  // TODO: Cotnainers IP addresses are dropped if network updates with IP invalidating changes
+  // TODO: Containers IP addresses are kept if still valid after update
+});
```
Mkassabov, line 25: could we implement these as part of this PR? I worry that due to the complexity of the docker resource we would never actually write tests for this and then we'd have what seems like a pretty critical resource being at risk of issues

alchemy/tsconfig.json
```diff
@@ -15,6 +15,7 @@
     "target": "ESNext",
     "allowJs": true,
     "allowImportingTsExtensions": true,
-    "rewriteRelativeImportExtensions": true
+    "rewriteRelativeImportExtensions": true,
+    "incremental": true
   }
 }
```
Mkassabov, line 21: This PR isn't really the place to make a change like this. It looks minor but it warrants more discussion; IIRC we had an issue at one point when we had this enabled and thats why we have it turned off.

I also worry that the need to turn this on could hide performance issues with our types, thats something we're really trying to focus on as we move towards effect

scripts/generate-docker-types.ts
```diff
@@ -0,0 +1,204 @@
...
+      }
+    } else {
+      console.log(options.path, "missed");
+      console.log(inspect(schemaObject, { depth: null, colors: true }));
+    }
+
+    return undefined;
+  },
+});
+await Bun.write("alchemy/src/docker/api/types.ts", astToString(nodes));
```
Mkassabov, line 53: We generally try to avoid Bun specific behavior.
We're aiming to support older and older versions of node and potentially get alchemy working in a browser environment. Can we get this changed?

scripts/generate-docker-types.ts
Mkassabov, line 1: We generate sdks from openapi using heyapi now, would rather not introduce a new system

alchemy/src/util/diff.ts
msanchezdev: Yeah, I have been considering removing this function later in favor of a simpler way of checking for changes (this is part of the initial implementation that was started some months ago).

But my focus is next going to be writing tests to make sure we don't break functionality when making changes.

alchemy/src/util/duration.ts
msanchezdev: Yeah, considering replacing it with https://www.npmjs.com/package/parse-duration

alchemy/tsconfig.json
msanchezdev: Looks like it got mixed up, seems to have been already enabled when my PR was created and didn't get rebased, weird.

scripts/generate-docker-types.ts
msanchezdev: Yup, this got mixed up as well, was supposed to be removed as it was too old and not being used anymore

scripts/generate-docker-types.ts
msanchezdev: Looks like I removed this file in another branch, this is actually old and not used anymore.

alchemy/test/docker/v2/image.test.ts
msanchezdev: Resources tests are still WIP, this first PR is to have the base resources working as expected because the current implementation is missing details and not very versatile (as it uses docker's CLI instead of the API). Which was the initial agreement with @sam-goodwin. 

Hoping this could have been reviewed while taking into account tests are still not implemented as more time was taken than what discussed initially because of complexity.

Everything was tested properly. I'm now writing testcases to make sure nothing breaks.

fyi: @sam-goodwin

msanchezdev:
> I think this PR is a bit outside the scope of what we want Alchemy to be right now. Ultimately, the majority of the use cases for containers we expect users to have is deploying them to other systems. Looking at this PR, I'm not really seeing good interop with other resources. Ultimately, this is a lot of code (and complex code too) that creates a maintenance burden on the team. All to support primarily local Docker, which isn't a common use case for IaC solutions. In my eyes, this fits better as a standalone library that has Alchemy as a peer dependency. If you choose to go down that route, we would happily support it.

@Mkassabov This is just the base for full Container support. Indeed, the ideal use case is deploying them on other systems, right now this can be achieved either through TCP(+TLS), or through SSH which is a common method of managing remote docker hosts.

In the roadmap we would be adding support for managing Docker Swarms which are a way to connect multiple docker hosts together. Even though currently you can manage multiple docker hosts by just declaring them in your `alchemy.run.ts` file.

I am aware of the complexity of the code (a lot can be simplified, like duration, bytes and change checking). I haven't focused too much on polishing it more as too much time was already spent on it, and the focus was having the bare minimum working, I made sure to make it all work first then to do the proper refactoring.

Mkassabov:
The entire direction of this PR is not what we're going with. It's also likely that it would need heavy changes when we make the move to effect, which will be soon.


We need docker to be much simpler and much more heavily integrated with other cloud providers. Don't feel obligated to continue working on this pr. Thanks for your work so far though!

PR #460, opened 2025-06-23
feat(cloudflare): pin default worker compatibility date to build time

### LINKED ISSUE #458: Pin default worker compatibility date
@claude please update our build procedure to produce a YYYY-MM-DD string and include it in the package in npm and use that as the default compatibility date for workers. 

This pins it to a version at the time of publish

sam-goodwin:
Closes #458

This PR updates the build procedure to automatically pin the default worker compatibility date to the package build date.

## Changes
- Add build script to generate YYYY-MM-DD date string at build time
- Update worker compatibility date to use generated build date instead of hardcoded value
- Include date generation in package build process to pin compatibility to publish version
- Auto-generated build-date.ts file is excluded from version control

This ensures worker compatibility dates are automatically pinned to the package version at publish time.


.gitignore
```diff
@@ -18,4 +18,7 @@ wrangler.jsonc
 !.husky/pre-commit
 
 .smoke
-.svelte-kit
\ No newline at end of file
+.svelte-kit
+
+# Auto-generated build files
+alchemy/src/build-date.ts
```
sam-goodwin, line 24: @claude let's generate the file and commit it so it's always there

sam-goodwin: @claude let's generate the file and commit it so it's always there

PR #465, opened 2025-06-23
fix(cloudflare): do state store fails to upload

john-royal:


alchemy/src/cloudflare/do-state-store/internal.ts
```diff
@@ -237,5 +237,7 @@ async function bundleWorkerScript() {
   if (!result.outputFiles?.[0]) {
     throw new Error("Failed to bundle worker.ts");
   }
-  return result.outputFiles[0].text;
+  return new File([result.outputFiles[0].text], "worker.js", {
+    type: "application/javascript+module",
+  });
```
sam-goodwin, line 242: why does it return a file @john-royal

PR #471, opened 2025-06-24
chore: astro website

sam-goodwin:


just-be-dev:
You'll need to update `website.run.ts` too. It's still configured for vuepress.

sam-goodwin:
> You'll need to update `website.run.ts` too. It's still configured for vuepress.  

I moved to alchemy-web/alchemy.run.ts

sam-goodwin:
Ci/cd is already set up and working for this pr I would hope

just-be-dev:
The latest deployment failed due to an [Astro bug](https://github.com/withastro/astro/issues/12036). When configured in static mode, the build fails if there's a dynamic redirect. It's [suggested in the docs](https://docs.astro.build/en/guides/routing/#configured-redirects) that this _should_ work. 

For now, I'm just going to enable SSR and have a single route that performs the dynamic redirect. We'll need SSR for things like OG images later anyway. We could alternatively just handle this with a cloudflare level redirect.

just-be-dev:
The only thing this is really missing is a redirect from `/docs` to `/`. Merging #500 will make that easy to add. Astro has a built in redirect mechanism, but as stated above there's a bug so it's not working.

Issue #438, opened 2025-06-21
Provide better DX around expired auth tokens


zephraph:
If you're trying to deploy something to cloudflare and either you haven't ran `wrangler login` yet or if the token expires your deploy will fail with an error like this

```
node:internal/modules/run_main:122
    triggerUncaughtException(
    ^

Error: Failed to refresh auth token: 400 Bad Request
    at refreshAuthToken (/Users/just-be/Code/abstract.industries/node_modules/.pnpm/alchemy@0.32.1_@ai-sdk+openai-compatible@0.2.14_zod@3.25.64__@ai-sdk+openai@1.3.22_zod@_f6944000cd99cd1e8270e847995c99a3/node_modules/alchemy/lib/cloudflare/auth.js:47:15)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async getRefreshedAuthConfig (/Users/just-be/Code/abstract.industries/node_modules/.pnpm/alchemy@0.32.1_@ai-sdk+openai-compatible@0.2.14_zod@3.25.64__@ai-sdk+openai@1.3.22_zod@_f6944000cd99cd1e8270e847995c99a3/node_modules/alchemy/lib/cloudflare/auth.js:65:26)
    at async getCloudflareAuthHeaders (/Users/just-be/Code/abstract.industries/node_modules/.pnpm/alchemy@0.32.1_@ai-sdk+openai-compatible@0.2.14_zod@3.25.64__@ai-sdk+openai@1.3.22_zod@_f6944000cd99cd1e8270e847995c99a3/node_modules/alchemy/lib/cloudflare/auth.js:26:24)
    at async CloudflareApi.fetch (/Users/just-be/Code/abstract.industries/node_modules/.pnpm/alchemy@0.32.1_@ai-sdk+openai-compatible@0.2.14_zod@3.25.64__@ai-sdk+openai@1.3.22_zod@_f6944000cd99cd1e8270e847995c99a3/node_modules/alchemy/src/cloudflare/api.ts:160:11)
    at async getWorkerScriptMetadata (/Users/just-be/Code/abstract.industries/node_modules/.pnpm/alchemy@0.32.1_@ai-sdk+openai-compatible@0.2.14_zod@3.25.64__@ai-sdk+openai@1.3.22_zod@_f6944000cd99cd1e8270e847995c99a3/node_modules/alchemy/src/cloudflare/worker.ts:1542:20)
    at async Promise.all (index 1)
[...]

PR #475, opened 2025-06-24
feat: automatic cloudflare login

john-royal:
Automatically logs in with Cloudflare when Wrangler config is unavailable or if the token refresh fails. Should help with #438.

- This does introduce a new dependency, `open` — this is because opening the URL cross-platform gets messy fast. I tried a simpler in-house method but seemed to completely butcher the URL. Happy to look at this further if we want to avoid the dependency.
- Auto-login is disabled when CI is detected or if the terminal seems to be non-interactive.
- Right now this is built into the current Cloudflare client setup, but we might want to make this part of the `alchemy create` CLI.

Idea: It would be really cool if we marked API requests as requiring specific scopes so we can start the reauthorization flow automatically if we hit an auth error.


sam-goodwin:
@claude please add a new command in alchemy/bin for `alchemy login` that:
1. opens the same login flow to cloudflare like this PR does
2. allows overriding the scopes via a CLI argument

john-royal:
Replaced by #631

PR #483, opened 2025-06-26
chore: dynamic OG generation

zephraph:
Adds an `/og/` route that will render an og card for any given page so long as there's a `.png` at the end. `/og/index.png` will render the default, `/og/getting-started.png` will render for the getting started page, etc. This is all done at build time with playwright. The og route sets `dist/og` as a symlink to `.astro/og-cache` which itself has a json manifest file that keeps track of all the generated og images and their content's digest. If you've run generation for everything, then the build will use the cached og images. That should only be on the order of 3ms. For pages whose content have changed or haven't yet been generated, it'll spin up a playwright instance and generate the og image. I've got that down to ~350ms per OG. If a page is deleted, the build process will detect that and delete any associated cached ogs (if they exist). 

It's not perfect because it _does_ increase the build time, but it's operationally simpler than what I was trying to do before. 

I've also added an env var `GENERATE_OG_IMAGES` which, if not set, will cause the build to _skip_ og generation. I've updated the publish workflow to enable that when running astro build. That means it'll skip OG generation by default in PR previews, but it'll have it in prod builds. 

image


sam-goodwin:
Looks like the website deploy failed because SSR is enabled.

Gotta look into why the build doesn't fail when that happens

image

alchemy-web/alchemy.run.ts
```diff
@@ -30,6 +30,15 @@ const url = domain ? `https://${domain}` : website.url;
...
 console.log(url);
 
+export const ogWorker = await Worker("alchemy-og-worker", {
+  entrypoint: "./src/og-worker.ts",
+  routes: [
+    {
+      pattern: "og.alchemy.run/*",
+    },
+  ],
+});
```
just-be-dev: cc @sam-goodwin would be good for you to review this specifically

alchemy-web/src/routeData.ts
```diff
@@ -0,0 +1,45 @@
...
+  // Get the content collection entry for this page.
+  try {
+    route = context.locals.starlightRoute;
+  } catch (_) {
+    // This is a non-starlight route, so we want to skip the og generation
+    return;
+  }
+
+  // Base OG image URL
+  const baseImageUrl = `/og/${route.id ?? "index"}`;
```
just-be-dev: The logic here needs to be updated. This shouldn't be `/og/` it should be something more like `og.alchemy.run/...`. I don't exactly know how best to get the domain though. @sam-goodwin this is what I was talking about when mentioning it'd be nice to have a config output that didn't require deployment. It's a bit of a chicken and egg situation because you need the config to be able to deploy but you can only get it by doing the deploy?

sam-goodwin: Why do we need a subdomain for og images instead of generating at build time?

just-be-dev:
Okay, here's what's left:
1. The `baseImageUrl` in `routeData.ts` needs to be updated to point at the og worker. 
2. The og worker needs to get the development URL treatment. 
3. Probably should add some cache control headers or manually use the cache API to ensure the og worker isn't generating the og images every time.

alchemy-web/src/pages/og/[...route].png.ts
```diff
@@ -0,0 +1,499 @@
...
+    mkdirSync(CACHE_DIR, { recursive: true });
+  }
+}
+
+function ensureDistSymlink() {
+  const distOgDir = join(process.cwd(), "dist", "og");
+  const distDir = join(process.cwd(), "dist");
+
+  // Ensure dist directory exists
+  if (!existsSync(distDir)) {
```
Comment (sam-goodwin) [inline alchemy-web/src/pages/og/[...route].png.ts]: nit: can it be async IO? I rarely see a case where it can't.

alchemy-web/src/pages/og/[...route].png.ts
```diff
@@ -0,0 +1,499 @@
...
+      if (existsSync(cacheFile)) {
+        filesToDelete.push(cacheFile);
+      }
+    }
+  }
+
+  // Delete outdated cached files
+  for (const file of filesToDelete) {
+    try {
+      unlinkSync(file);
```
Comment (sam-goodwin) [inline alchemy-web/src/pages/og/[...route].png.ts]: This looks very slow.

alchemy-web/src/pages/og/[...route].png.ts
```diff
@@ -0,0 +1,499 @@
...
+  const oldManifest = loadManifest();
+  const filesToDelete: string[] = [];
+
+  // Find cached files that are no longer needed
+  for (const oldPath in oldManifest) {
+    if (!currentPaths.has(oldPath)) {
+      // Convert /og/path/to/file.png to cache/path/to/file.png
+      const relativePath = oldPath.replace(/^\/og\//, "");
+      const cacheFile = join(CACHE_DIR, relativePath);
+      if (existsSync(cacheFile)) {
```
Comment (sam-goodwin) [inline alchemy-web/src/pages/og/[...route].png.ts]: Slow synchronous loop

alchemy-web/src/pages/og/[...route].png.ts
```diff
@@ -0,0 +1,499 @@
...
+    }
+  }
+
+  // Save the new manifest
+  saveManifest(newManifest);
+
+  return paths;
+}
+
+export const GET: APIRoute = async ({ props, params }) => {
```
Comment (sam-goodwin) [inline alchemy-web/src/pages/og/[...route].png.ts:227]: This is great. Love the solution you came up with

alchemy-web/src/pages/og/[...route].png.ts
Comment (just-be-dev) [reply alchemy-web/src/pages/og/[...route].png.ts]: Can be, I just haven't optimized these parts yet.

alchemy-web/src/pages/og/[...route].png.ts
Comment (just-be-dev) [reply alchemy-web/src/pages/og/[...route].png.ts]: These are all fairly negligible. Batched stuff up and made it more async. This part only runs once at the beginning of the build and the N is small enough that it doesn't really matter.

PR #486, opened 2025-06-26
Read `apiToken` when creating DOStateStore

MathiasWP:
I noticed that when only passing `apiToken` into the DOStateStore it throws `Error: Environment variable ALCHEMY_STATE_TOKEN is not set`, but setting the same API token in `worker.token` works. IMO it aligns better with the API of the other configs to allow `apiToken`, instead of using a very specific `worker.token` config

### DIFF
--- a/alchemy/src/cloudflare/do-state-store/store.ts
+++ b/alchemy/src/cloudflare/do-state-store/store.ts
@@ -58,6 +58,7 @@ export class DOStateStore implements StateStore {
   private async createClient() {
     const token =
       this.options.apiToken?.unencrypted ??
+      (await alchemy.secret.env.CLOUDFLARE_API_TOKEN).unencrypted ??
       this.options.worker?.token ??
       (await alchemy.secret.env.ALCHEMY_STATE_TOKEN).unencrypted;
     if (this.options.worker && "url" in this.options.worker) {


alchemy/src/cloudflare/do-state-store/store.ts
```diff
@@ -57,6 +57,8 @@ export class DOStateStore implements StateStore {
 
   private async createClient() {
     const token =
+      this.options.apiToken?.unencrypted ??
+      (await alchemy.secret.env.CLOUDFLARE_API_TOKEN).unencrypted ??
       this.options.worker?.token ??
       (await alchemy.secret.env.ALCHEMY_STATE_TOKEN).unencrypted;
```
sam-goodwin, line 63: These are not the same tokens. ALCHEMY_STATE_TOKEN is a special token for authorizing with the `alchemy-state` Worker+DO we deploy into your account for maintaining state.

MathiasWP: Oh, then i have misunderstood. I struggled to find documentation about the ALCHEMY_STATE_TOKEN. What value should this token have?

sam-goodwin: You're the second person to run into this problem today. Clearly a gap in our docs! 

The DOStateStore is a store we provide to store state for your apps+stages in a Cloudflare Worker backed by a SQLite Durable Object.

The problem solved by `ALCHEMY_STATE_TOKEN`  is it's a token we upload to the Worker that you use to authorize access to it, otherwise it would be open to the world.

There's an alternative solution using Zero Trust but it requires manual steps. Maybe we should explore that though.

@claude please document this in alchemy-web/docs/concepts/state.md and in the alchemy-web/docs/getting-started.md. Let's treat DOStateStore as the recommended approach and walk users through setting it up.

sam-goodwin: Oh claude problably won't work on an external PR, tracking. here https://github.com/sam-goodwin/alchemy/issues/490

sam-goodwin: Here's a preview of new docs @MathiasWP https://9b8aa2f6-alchemy-website-test.alchemy-run.workers.dev/guides/do-state-store/

sam-goodwin: https://alchemy.run/guides/do-state-store/

sam-goodwin:
Closing this as it's now documented. Please open an issue if there's a persistent problem

https://alchemy.run/guides/do-state-store/

PR #493, opened 2025-06-26
feat: miniflare container bindings

john-royal:
## Changes
- Modified the Cloudflare `Container` resource so that in dev mode, the image is not pushed to Cloudflare and uses Miniflare's image prefix so Miniflare can read the image.
- Updated Miniflare options to use the correct image reference and Docker socket path.

## Caveats
- To use dev mode with a remote Worker, you must set `dev: { remote: true }` on the `Container` so it's still pushed to Cloudflare.
- If you set `dev: { remote: true }` on a `Container`, you cannot use it as a local binding. We might want to add some way to extract the DO and push that up to Cloudflare independently of the rest of your worker script so remote bindings still work.
- When you call `container.fetch()` while the container is still starting, you'll see this error in your console, which can safely be ignored:
  ```
  Error checking if container is ready: connect(): Connection refused: container port not found. Make sure you exposed the port in your container definition.
  ```
  The binding still works and Wrangler logs the same message, so this seems like a problem with Miniflare internals.

### DIFF
--- a/alchemy/src/cloudflare/container.ts
+++ b/alchemy/src/cloudflare/container.ts
@@ -1,6 +1,7 @@
 import type { Context } from "../context.ts";
-import { Image, type ImageProps, type ImageRegistry } from "../docker/image.ts";
+import { Image, type ImageProps } from "../docker/image.ts";
 import { Resource } from "../resource.ts";
+import { Scope } from "../scope.ts";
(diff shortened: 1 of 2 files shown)


john-royal:
@claude Update the docs for dev mode and the container resource to account for this change

PR #495, opened 2025-06-26
fix: bun --watch should not set dev: true

### LINKED ISSUE #494: `bun --watch` should not set dev:true
when a user runs `bun --watch alchemy.run.ts` or `node --watch` etc., we infer that to be "dev: true" by default, but dev: true should be opted in via --dev

```sh
bun --watch alchemy.run.ts --dev
```

`bun --watch` without `--dev` means continually push to live services. for that we will need to modify worker.ts to use esbuild context and push changes to the worker (with metadata, etc.). This will be tricky for the noBundle: true use-case which doesn't invoke esbuild, but we might be able to skip that for now since it's mostly used for websites which currently use `vite dev` instead

john-royal:
Closes #494

### DIFF
--- a/alchemy/src/alchemy.ts
+++ b/alchemy/src/alchemy.ts
@@ -34,11 +34,7 @@ function parseCliArgs(): Partial<AlchemyOptions> {
     options.phase = "read";
   }
 
-  if (
-    args.includes("--dev") ||
-    args.includes("--watch") ||
-    process.execArgv.includes("--watch")
-  ) {
+  if (args.includes("--dev")) {
     options.dev = true;
   }
 
--- a/alchemy-web/docs/guides/dev.md
+++ b/alchemy-web/docs/guides/dev.md
@@ -26,7 +26,7 @@ This starts Alchemy in development mode, which will:
 
 ### Watching Your Alchemy Configuration
 
-Alchemy does not watch your `alchemy.run.ts` file for changes. To automatically apply changes to your configuration, you can the watch mode associated with your runtime environment. For example:
+Alchemy does not watch your `alchemy.run.ts` file for changes. To automatically apply changes to your configuration, you can use the watch mode associated with your runtime environment. For example:
 
 ```bash
 # Using bun's watch mode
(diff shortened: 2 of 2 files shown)


sam-goodwin:
@claude update the dev mode docs

alchemy/src/alchemy.ts
```diff
@@ -34,11 +34,7 @@ function parseCliArgs(): Partial<AlchemyOptions> {
...
   }
 
-  if (
-    args.includes("--dev") ||
-    args.includes("--watch") ||
-    process.execArgv.includes("--watch")
-  ) {
+  if (args.includes("--dev")) {
     options.dev = true;
   }
```
sam-goodwin, line 39: @john-royal is this right?

We need `--watch` to enable watching (push esbuild changes to live Worker) and `--dev` to enable local mode.

john-royal: We don't currently push to the live Worker, so pushing ESBuild changes is only supported if it's running locally. Definitely want to add that but there are some complications there.

How do you want this to work? I thought the way we wanted this to work was that `--dev` enables watching, and if you want the Worker to be remote, then you set `dev: { remote: true }` on the resource itself. Correct me if I'm wrong.

sam-goodwin: I was thinking dev enables local. Maybe we should change it to --local?

We need:
- watch (push changes)
- local (emulate locally instead of remote)

john-royal: Would that override the dev: { remote: boolean } property on resources? Replace it entirely?

sam-goodwin: That's a good reason to keep it as --dev instead of --local

A global flag should not take precedence over an explicit `dev` config on the resource.

sam-goodwin:
Should we merge this or close in favor of https://github.com/sam-goodwin/alchemy/pull/543 ?

sam-goodwin:
> Should we merge this or close in favor of #543 ?

@john-royal

PR #496, opened 2025-06-26
fix(cloudflare): consistent ports for miniflare dev server

john-royal:
Finds an open port starting at 1337 to use for Miniflare dev.


sam-goodwin:
> 1337

Bruh. We just think the same.

PR #500, opened 2025-06-26
feat: add RedirectRule resource for Cloudflare single redirects

### LINKED ISSUE #498: Add support for Cloudflare's Rules to enable simpler redirects
Unlike [vercel](https://vercel.com/docs/redirects#configuration-redirects) cloudflare doesn't provide a trivial mechanism for redirects. Typically redirection or rewrites are handled in a worker directly. There are times though, especially when you're using a framework, where it's not easy or convenient to define a redirect that way. 

Given that, I think it would be helpful for Alchemy to add support for [Cloudflare's Rules](https://developers.cloudflare.com/rules/) feature. Rules are fairly complex though, there's a lot they can do. I'd like to just see a narrow set of rules to start with. 

The examples below don't explicitly show it, but the zone will either need to be provided or inferred if possible. 

## Redirect Rule

The first I'd like to see is a redirect rule. This would be based on CF's [Single Redirects](https://developers.cloudflare.com/rules/url-forwarding/single-redirects/) rule type. See the [available options documentation](https://developers.cloudflare.com/rules/url-forwarding/single-redirects/settings/) for how a single redirect can be configured. Here's the API I'd like to see

```ts
// Wildcard Redirect
await RedirectRule("my-wildcard-redirect", {
  requestUrl: "https://*.example.com/files/*",
  targetUrl: "https://example.com/${1}/files/${2}",
  statusCode: 301,
  preserveQueryString: true
});

// Static redirect
await RedirectRule("my-wildcard-redirect-2", {
  // Note static has no `requestUrl`
  targetUrl: "https://example.com/",
  statusCode: 301,
  preserveQueryString: true
});

// Dynamic redirect
await RedirectRule("my-wildcard-redirect-3", {
  // See https://developers.cloudflare.com/ruleset-engine/rules-language/expressions/ for details on expressions
  expression: "http.request.uri.path matches "/autodiscover\.(xml|src)$"
  statusCode: 301,
  preserveQueryString: true
});
```

zephraph:
Add support for Cloudflare's Rules to enable simpler redirects

Implements RedirectRule resource supporting:
- Wildcard redirects with requestUrl patterns
- Static redirects (no pattern, matches all requests)
- Dynamic redirects with custom expressions

Resolves #498

### DIFF
--- a/alchemy/test/cloudflare/redirect-rule.test.ts
+++ b/alchemy/test/cloudflare/redirect-rule.test.ts
@@ -201,11 +201,14 @@ describe("RedirectRule", () => {
       });
 
       // Should NOT throw error when neither requestUrl nor expression are provided (static redirect)
-      const redirectRule = await RedirectRule(`${BRANCH_PREFIX}-static-redirect2`, {
-        zone: zone.id,
-        targetUrl: `https://${testDomain}/`,
-        statusCode: 301,
-      });
+      const redirectRule = await RedirectRule(
+        `${BRANCH_PREFIX}-static-redirect2`,
+        {
+          zone: zone.id,
+          targetUrl: `https://${testDomain}/`,
+          statusCode: 301,
+        },
+      );
 
       expect(redirectRule.requestUrl).toBeUndefined();
       expect(redirectRule.expression).toBeUndefined();
--- a/alchemy/test/cloudflare/redirect-rule.test.ts
+++ b/alchemy/test/cloudflare/redirect-rule.test.ts
@@ -5,6 +5,7 @@ import { RedirectRule } from "../../src/cloudflare/redirect-rule.ts";
 import { Zone } from "../../src/cloudflare/zone.ts";
 import { destroy } from "../../src/destroy.ts";
 import { BRANCH_PREFIX } from "../util.ts";
+import { fetchAndExpectStatus } from "./fetch-utils.ts";
 
 import "../../src/test/vitest.ts";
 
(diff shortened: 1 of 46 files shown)


alchemy/test/cloudflare/redirect-rule.test.ts
```diff
@@ -0,0 +1,232 @@
...
+      });
+
+      // Create a wildcard redirect rule
+      redirectRule = await RedirectRule(`${BRANCH_PREFIX}-wildcard-redirect`, {
+        zone: zone.id,
+        requestUrl: `https://*.${testDomain}/files/*`,
+        targetUrl: `https://${testDomain}/\${1}/files/\${2}`,
+        statusCode: 301,
+        preserveQueryString: true,
+      });
```
sam-goodwin: Can we avoid using an actual zone here? Would prefer to just workers.dev if we can, but if it's not supported then ok. But we'll have to get an actual domain to test it...

alchemy/test/cloudflare/redirect-rule.test.ts
```diff
@@ -0,0 +1,232 @@
...
+
+      // Verify the rule was created by querying the API directly
+      const getRuleResponse = await api.get(
+        `/zones/${zone.id}/rulesets/${redirectRule.rulesetId}/rules/${redirectRule.ruleId}`,
+      );
+      expect(getRuleResponse.status).toEqual(200);
+
+      const ruleData: any = await getRuleResponse.json();
+      expect(ruleData.result.action).toEqual("redirect");
+      expect(ruleData.result.enabled).toEqual(true);
```
sam-goodwin: Ideally we would actually invoke the URL and test it redirects intead of just asserting it exists. Check behavior.

alchemy/test/cloudflare/redirect-rule.test.ts
just-be-dev: Unclear to me: https://developers.cloudflare.com/rules/transform/url-rewrite/create-api/. It _seems_ like the zone is required, but can you get a zone_id for a worker URL?

alchemy/test/cloudflare/redirect-rule.test.ts
just-be-dev: @claude update the tests to ensure the redirect actually happens.

PR #502, opened 2025-06-27
fix(cloudflare): resolve paths relative to wrangler.json file location

### LINKED ISSUE #501: WranglerJson should ensure that relative path for `main` is relative to the location of the json file
Customer issue:

I'm having some issues with the generated WrangerJSON within my monorepo. Everything works well except that the main property of the wrangler file is ./packages/backend/src/index.ts and I would've expected it to be ./src/index.ts

This is the relevant part of my alchemy file, which is at the root of the monorepo:

```ts
export const worker = await Worker("backend", {
  entrypoint: "./packages/backend/src/index.ts",
  compatibilityFlags: ["nodejs_compat"],
  bindings: {
    Browser: new BrowserRendering(),
    BrowserSession: browserSession,
  },
});

await WranglerJson("wrangler.jsonc", {
  worker,
  path: "packages/backend/wrangler.jsonc",
});
```

here, the file is being written to `packages/backend/wrangler.jsonc` and the `main` is `"./packages/backend/src/index.ts"`. When writing out the wrangler json, we should be resolving these paths to absolute directories and then writing them to wrangler.json relative to the dirname of that file.

sam-goodwin:
Fixes #501

This PR fixes the WranglerJson resource to ensure that paths in the generated wrangler.json file are relative to the location of the json file itself, not to the current working directory.

## Changes

- Resolve entrypoint and assets paths to absolute paths relative to cwd
- Convert paths to be relative to the wrangler.json file location using path.relative()
- Handle both props.assets and assets bindings consistently
- Add comprehensive test for path resolution behavior
- Update existing test expectations to match new behavior

## Problem Solved

Before: In monorepo setup, WranglerJson would generate `main: "./packages/backend/src/index.ts"` in wrangler.json located at `packages/backend/wrangler.jsonc`

After: Now correctly generates `main: "./src/index.ts"` relative to the wrangler.json file location

### DIFF
--- a/alchemy/src/cloudflare/wrangler.json.ts
+++ b/alchemy/src/cloudflare/wrangler.json.ts
@@ -47,6 +47,13 @@ export interface WranglerJsonProps {
     binding: string;
     directory: string;
   };
+
+  /**
+   * Base directory for resolving relative paths
+   *
+   * @default process.cwd()
+   */
+  cwd?: string;
 }
 
 /**
@@ -89,6 +96,9 @@ export const WranglerJson = Resource(
     _id: string,
     props: WranglerJsonProps,
   ): Promise<WranglerJson> {
+    // Base directory for resolving relative paths
+    const cwd = props.cwd || process.cwd();
+
     // Default path is wrangler.json in current directory
     const filePath = props.path || "wrangler.jsonc";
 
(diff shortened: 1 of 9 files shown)


sam-goodwin:
@claude this broke Website which is no longer respecting cwd. Do we need to pass that through? Or is something else broken?

alchemy/src/cloudflare/wrangler.json.ts
```diff
@@ -47,6 +47,13 @@ export interface WranglerJsonProps {
     binding: string;
     directory: string;
   };
+
+  /**
+   * Base directory for resolving relative paths
+   *
+   * @default process.cwd()
+   */
+  cwd?: string;
```
sam-goodwin, line 56: @claude Ideally we can infer this from the `worker` property instead since if a user specifies `cwd` in WorkerProps or Worker, then it should default to that

PR #503, opened 2025-06-27
feat(dev): allow logging on inspect.localhost:<port>

Mkassabov:
## What this does
- for a worker on `localhost:1024` the inspector is available on `inpsect.localhost:1024`

## Currently soft-blocked by a miniflare bug.
see Miniflare [PR: 9784](https://github.com/cloudflare/workers-sdk/pull/9784)
When we add new workers to miniflare via `setOptions` an inspector proxy ins't started for that worker.
There does seem to be a way to get all of the proxies, but its not reliable.

## TODO:
- [ ] merge in port access from main
- [ ] add dev option to enable disable inspector
- [ ] add option to disable a worker from logging to the terminal

### DIFF
--- a/alchemy/package.json
+++ b/alchemy/package.json
@@ -137,6 +137,7 @@
     "libsodium-wrappers": "^0.7.15",
     "turndown": "^7.0.0",
     "unenv": "2.0.0-rc.15",
+    "ws": "^8.18.2",
     "yaml": "^2.0.0"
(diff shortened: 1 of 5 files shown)


Mkassabov:
moved to [PR: 532](https://github.com/sam-goodwin/alchemy/pull/532) so its no longer from a fork

PR #508, opened 2025-06-27
Add instrumentation to site

zephraph:
As we look to improve onboarding, its helpful to know where people are following off as they're visiting the site and exploring the docs. This PR adds [posthog](https://posthog.com) to help get better visibility into user behavior. 

I think there's also a good content opportunity here. Posthog [has a whole article](https://posthog.com/docs/advanced/proxy/cloudflare) on how to setup a reverse proxy for posthog (to get around common adblockers). With alchemy, you don't _need_ an article, it's just one piece of worker config. It makes me wonder how it could be packaged up further?


sam-goodwin:
Looks like this got implemented twice

PR #513, opened 2025-06-29
Just more windows support (primarily for the test suite)

Mkassabov:
## fixes
- we now have cli build scripts that are windows compatible
- esbuild test always uses posix paths
- the default name now checks for `process.env.USERNAME` which is the windows equivalent to `process.env.USER`


alchemy/package.json
```diff
@@ -10,10 +10,14 @@
     "directory": "alchemy"
   },
   "scripts": {
-    "build:cli": "./scripts/bundle-cli.sh",
+    "build:cli": "run-script-os",
+    "build:cli:windows": "powershell -File ./scripts/bundle-cli.ps1",
+    "build:cli:default": "./scripts/bundle-cli.sh",
```
sam-goodwin, line 15: Not sure why this is needed? Is it so we can bundle the cli while developing alchemy on a windows machine?

Mkassabov: Yeah.

The bigger problem is that build:cli is part of build, so without this you can't run the build command on windows

PR #514, opened 2025-06-29
Add `alchemy deploy` to the CLI

zephraph:
Introduces a dedicated deploy command to the Alchemy CLI, making deployment operations more explicit and discoverable. I'm leaving this in draft because it's as much a proposal as anything. The command could also be updated to `alchemy up`. 

## Changes

  - Added new deploy command with comprehensive options for managing
  Alchemy deployments
  - Supports all common deployment scenarios including read-only
  mode, destroy operations, and stage targeting
  - Includes watch mode for automatic redeployment on file changes

## Command Usage

```
  Usage: alchemy deploy [options]

  Deploy an Alchemy project

  Options:
    --path [string]      Path to the project directory (defaults to
  current
                         directory)
    --quiet [boolean]    Suppress Create/Update/Delete messages
  (default: false)
    --no-quiet           Negate `--quiet` option.
    --read [boolean]     Read-only mode - doesn't modify resources
  (default:
                         false)
    --no-read            Negate `--read` option.
    --destroy [boolean]  Destroy all resources (default: false)
    --no-destroy         Negate `--destroy` option.
    --stage [string]     Specify which stage/environment to target
    --watch [boolean]    Watch for file changes and redeploy
  automatically
                         (default: false)
    --no-watch           Negate `--watch` option.
    -h, --help           display help for command
```    

## Rationale

  Previously, users needed to run their `alchemy.run.ts` file directly
   with package manager commands like `bun ./alchemy.run.ts` or `npx tsx
   ./alchemy.run.ts`. This approach had several limitations:

  1. Discoverability: New users had to consult documentation to
  understand how to deploy their infrastructure and what options were available. If they passed `--help` while invoking alchemy, it'd ignore it and just run the deploy. 
  2. Inconsistency: Different package managers required different
  commands and flags
  3. User Expectations: Most infrastructure tools provide explicit
  deployment commands (e.g., terraform apply, pulumi up)

  The new alchemy deploy command:
  - Provides a consistent interface across all package managers
  - Makes available options immediately discoverable through --help
  - Matches typical user expectations from other IaC tools
  - Automatically detects and uses the appropriate package manager
  and runtime

## Backwards compatibility

The previous approach still works and this `deploy` command just builds on it. It's still possible to run `alchemy.run.ts` directly. 

##  Watch Mode

The command also includes a `--watch` flag that monitors your project
   files and automatically redeploys when changes are detected.


alchemy/bin/commands/deploy.ts
```diff
@@ -0,0 +1,216 @@
...
+    quiet,
+    read,
+    destroy,
+    stage,
+    watch: watchMode,
+  } = input;
+
+  // Check for alchemy.run.ts or alchemy.run.js
+  const runTsPath = resolve(path, "alchemy.run.ts");
+  const runJsPath = resolve(path, "alchemy.run.js");
```
sam-goodwin, line 32: Should we allow the user to pass a path and default to alchemy.run.ts ?

alchemy/bin/commands/deploy.ts
```diff
@@ -0,0 +1,216 @@
...
+
+  let runFile: string | null = null;
+  if (existsSync(runTsPath)) {
+    runFile = runTsPath;
+  } else if (existsSync(runJsPath)) {
+    runFile = runJsPath;
+  }
```
sam-goodwin, line 39: nit: async io

alchemy/bin/commands/deploy.ts
```diff
@@ -0,0 +1,216 @@
...
+      break;
+  }
+
+  // Execute the deploy
+  async function executeDeploy() {
+    const s = spinner();
+    s.start(`Executing: ${command}`);
+
+    try {
+      const { stdout, stderr } = await execAsync(command, {
```
sam-goodwin, line 97: We could also dynamically import the file, not sure if that's better. It would allow us to introspect the state for later use cases.

sam-goodwin: Hmm but might not work as well for different runtimes. Ok new process makes more sense.

alchemy/bin/commands/deploy.ts
```diff
@@ -0,0 +1,216 @@
...
+      }
+    } catch (error: any) {
+      s.stop("Deploy failed");
+      log.error(pc.red(`Deploy failed: ${error.message}`));
+      if (error.stdout) {
+        console.log(error.stdout);
+      }
+      if (error.stderr) {
+        console.error(error.stderr);
+      }
```
sam-goodwin, line 122: Can we stream instead? Does this mean output is buffered until the end?

alchemy/bin/commands/deploy.ts
```diff
@@ -0,0 +1,216 @@
...
+          log.info(pc.yellow(`\nFile changed: ${filename}`));
+        }
+        log.info(pc.yellow("Redeploying..."));
+        await executeDeploy();
+        log.info(pc.cyan("\nWatching for changes..."));
+      }, 300);
+    };
+
+    // Watch the main alchemy.run file
+    const mainWatcher = watch(runFile, (eventType) => {
```
sam-goodwin, line 156: For watch we should be passing --watch instead of watching the file. 

Their script can be spread across many files so we want to rely on the runtime's watch which knows which modules have been loaded and need to be watched.

alchemy/bin/services/package-manager.ts
```diff
@@ -3,17 +3,30 @@ import { execa } from "execa";
...
+): PackageManager {
+  if (fs.pathExistsSync(`${path}/deno.lock`)) return "deno";
+  if (
+    fs.pathExistsSync(`${path}/deno.json`) ||
+    fs.pathExistsSync(`${path}/deno.jsonc`)
+  )
+    return "deno";
+  if (fs.pathExistsSync(`${path}/bun.lockb`)) return "bun";
+  if (fs.pathExistsSync(`${path}/pnpm-lock.yaml`)) return "pnpm";
+  if (fs.pathExistsSync(`${path}/yarn.lock`)) return "yarn";
```
sam-goodwin, line 17: nit: async

sam-goodwin (changes requested):

alchemy/bin/commands/deploy.ts
just-be-dev: Yeah, seems reasonable. Maybe just default it to `alchemy.run.ts` and let them supply it otherwise?

alchemy/bin/commands/deploy.ts
just-be-dev: I think we can always drop the `--watch` portion of this for now.

sam-goodwin:
Closing in favor of https://github.com/sam-goodwin/alchemy/pull/636

PR #515, opened 2025-06-30
Add cloudflare llm.txt

zephraph:
I've been working on a few projects with claude code that are trying to use alchemy and it's been a bit of a struggle how little it can grasp how to properly implement an alchemy file. I've implemented a simple cloudflare based llm.txt that has a subset of docs but at least helps get the bots on the right path. I think longer term we'll want a proper mcp server, but for now this seems like it'll work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,4 +19,4 @@ wrangler.jsonc
 
 .smoke
 .svelte-kit
-
+.claude/*.local.json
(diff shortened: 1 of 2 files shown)


sam-goodwin:
Thanks for this, i might start with this in our vibe-rules distribution!

PR #519, opened 2025-07-01
fix: resolve wrangler.jsonc relative to worker cwd

john-royal:
I figured I'd call it `directory` in props and state to signal that it's a relative path in state for portability. Lots of wrangling required with absolute/relative conversions internally. Testing right now before exiting draft state.

UPDATE: Tests are passing, now testing manually


alchemy/src/cloudflare/wrangler.json.ts
```diff
@@ -134,7 +137,7 @@ export const WranglerJson = Resource(
     // Return the resource
     return this({
       ...props,
-      path: filePath,
+      path: path.relative(directory, filePath),
```
sam-goodwin: You need to set this on `main` in the data written to the file, not just here

alchemy/test/cloudflare/wrangler-json.test.ts
```diff
@@ -522,5 +522,37 @@ describe("WranglerJson Resource", () => {
...
+
+        const { spec } = await WranglerJson(
+          `${BRANCH_PREFIX}-test-wrangler-json-cwd`,
+          { worker },
+        );
+
+        expect(spec.main).toBe("worker.ts");
+        await expect(
+          fs.access(path.join(tempDir, "wrangler.jsonc")),
+        ).resolves.toBeUndefined();
```
sam-goodwin: Should read and assert the value of `main` and `assets`

alchemy/src/cloudflare/website.ts
```diff
@@ -165,10 +159,11 @@ export async function Website<B extends Bindings>(
...
           props.compatibilityDate ?? DEFAULT_COMPATIBILITY_DATE,
         name: workerName,
-        entrypoint: mainPath
-          ? // this path should be relative to the process.cwd() because it is used by esbuild
-            path.relative(process.cwd(), mainPath)
-          : undefined,
+        cwd:
+          props.cwd && path.isAbsolute(props.cwd)
+            ? path.relative(process.cwd(), props.cwd)
+            : props.cwd,
```
sam-goodwin: Can we use `resolveAbsPath` to standardize this resolution?

alchemy/src/cloudflare/website.ts
```diff
@@ -195,18 +190,41 @@ export default {
...
+          if (inputMainPath && path.isAbsolute(inputMainPath)) {
+            inputMainPath = path.relative(wranglerDir, inputMainPath);
+          }
+          inputAssetsDirPath = path.relative(wranglerDir, inputAssetsDirPath);
+        } else {
+          if (inputMainPath && path.isAbsolute(inputMainPath)) {
+            inputMainPath = path.relative(cwd, inputMainPath);
+          }
+          inputAssetsDirPath = path.relative(cwd, inputAssetsDirPath);
+        }
```
sam-goodwin: I didn't expect this to need to change? Couldn't it just pass through `cwd` and have `WranglerJson` do the resolution?

alchemy/src/cloudflare/website.ts
```diff
@@ -189,24 +174,23 @@ export default {
...
-          (f ? path.relative(path.dirname(wranglerJsonPath), f) : f) as S;
+        const wranglerPath = path.relative(cwd, wranglerJsonPath);
+        const wranglerDir = path.dirname(wranglerPath);
+
         await WranglerJson("wrangler.jsonc", {
-          path: wranglerJsonPath,
+          path: wranglerPath,
           worker: workerProps,
-          main: relativeToWrangler(wranglerMainPath),
+          main: mainPath ? path.relative(wranglerDir, mainPath) : undefined,
```
sam-goodwin, line 188: Why would it be undefined?

john-royal: The type is optional. I think the intended behavior is:

```ts
  /**
   * The entrypoint to your server
   *
   * @default - a simple server that serves static assets is generated
   */
  main?: string;
  ```
  
  Should we just make it "worker.js" or something? I think the Vite plugin actually needs a real script to reference.

sam-goodwin: Ah I see. It's allowed for there to be no worker? Like a Pages only solution

john-royal: This is already implemented (I copied that typedef from the actual file). If there's no `main` or `entrypoint`, the website component already passes a generic `script` property to the 

I can't tell if it actually works though, let me double check.

Issue #529, opened 2025-07-02
Rename phase "up" to "apply" and deprecate (but still support "up")


sam-goodwin:
@claude, we are renaming the "up" phase to "apply" to align with our internal concepts and language

jonbeckman:
@sam-goodwin can we also change this while you're at it?
```
- export type Phase = "up" | "destroy" | "read";
+ export const PhaseOptions = ["up", "destroy", "read"] as const
+ export type Phase = (typeof PhaseOptions)[number]
```

Discord context: https://discord.com/channels/1359694195782320389/1359694196830765059/1389946528336056471

PR #531, opened 2025-07-02
feat: rename "up" phase to "apply" with backward compatibility

sam-goodwin:
Implements #529 - Rename phase "up" to "apply" and deprecate (but still support "up")

## Summary
This PR renames the "up" phase to "apply" throughout the Alchemy codebase while maintaining full backward compatibility.

## Changes
- Updated Phase type to include both "apply" and "up" (with "apply" as primary)
- Changed default phase from "up" to "apply" across the codebase
- Added deprecation warnings when "up" is used explicitly
- Support both --apply and --up CLI flags (--up shows deprecation warning)
- Updated all core infrastructure, test frameworks, and examples
- Updated comprehensive documentation to use "apply" as preferred phase

## Backward Compatibility
- All existing code using "up" continues to work
- Deprecation warnings guide users to "apply"
- State files with "up" phase remain readable

### DIFF
--- a/alchemy-web/docs/concepts/destroy.md
+++ b/alchemy-web/docs/concepts/destroy.md
@@ -13,7 +13,7 @@ Resource destruction in Alchemy removes resources from both your state file and 
 ```typescript
 // Destroys all resources in the application
 const app = await alchemy("my-app", {
-  phase: "destroy"  // or process.argv.includes("--destroy") ? "destroy" : "up"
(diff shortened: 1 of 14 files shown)


sam-goodwin:
@claude

Please export an array of the allowed phase values

```
- export type Phase = "up" | "destroy" | "read";
+ export const PhaseOptions = ["up", "destroy", "read"] as const
+ export type Phase = (typeof PhaseOptions)[number]
```

PR #509, opened 2025-06-27
feat(turso): Turso Provider

sam-goodwin:
Closes #506

Issue #506, opened 2025-06-27
Turso Provider


osdiab:
Would be cool to have a way to use this with Turso, a SQLite provider.

https://turso.tech/

Thanks!

sam-goodwin:
Got a start on this generated https://github.com/sam-goodwin/alchemy/pull/509, would love your feedback

john-royal:
Putting this on the roadmap. Can't guarantee a release date but I'll try to work on it this week or next.

PR #533, opened 2025-07-02
feat: turso provider with effect

john-royal:


john-royal:
@claude Update the tests and documentation

sam-goodwin:
nice, should we close https://github.com/sam-goodwin/alchemy/pull/509

sam-goodwin:
Closing this. Re-open when we want to prioritize this

mountain1009:
Will this be resumed?

kandros:
Is there any blocker about this or just lack of time for the final touches and testing?

mchlggr:
Alchemy has been very helpful! It would be cool to use this with Turso as a SQLite provider, one of the only parts of my stack without an official provider. Is this PR in a position to be tested?
Thanks

alchemy/src/turso/api-token.ts
aleclarson, line 97: This appears to be an incomplete example, since there's no difference from normal usage.

john-royal:
Just noticed all the comments on here. We can definitely put Turso on our roadmap. If you're interested, you can follow issue #506.

This branch is a mess and 6 months out of date, so it would be a separate PR. I'll update the issue when I start working on it - hopefully later this week or next. Can't promise a release date though.

PR #538, opened 2025-07-03
feat(os): allow secrets in Exec environment variables

### LINKED ISSUE #537: Exec env can't take secrets
The env property of Exec can't be passed an `alchemy.secret` value

sam-goodwin:
This PR implements the ability to pass `alchemy.secret` values as environment variables to the Exec resource.

## Changes
- Updated `ExecProps.env` type to accept `Record<string, string | Secret>`
- Added runtime handling to unpack secrets using `value.unencrypted`
- Added comprehensive tests for secret-only and mixed environment variables
- Maintained backward compatibility with existing string-only env vars

## Usage Example
```typescript
const deploy = await Exec("deploy-app", {
  command: "npm run deploy",
  env: {
    NODE_ENV: "production",                    // Regular string
    API_KEY: secret(process.env.API_KEY),     // Secret value
    DATABASE_URL: secret(process.env.DATABASE_URL)  // Secret value
  }
});
```

Fixes #537

### DIFF
--- a/alchemy/src/os/exec.ts
+++ b/alchemy/src/os/exec.ts
@@ -48,7 +48,7 @@ export interface ExecProps {
   /**
    * Environment variables to set
    */
-  env?: Record<string, string | Secret>;
+  env?: Record<string, string | Secret<string>>;
 
   /**
    * Whether to inherit stdio from parent process
@@ -125,14 +125,14 @@ export interface Exec extends Resource<"os::Exec">, ExecProps {
  *
  * @example
  * // Run a command with secrets in environment variables
- * import { secret } from "../secret.ts";
+ * import alchemy from "alchemy";
  *
  * const deploy = await Exec("deploy-app", {
  *   command: "npm run deploy",
  *   env: {
  *     NODE_ENV: "production",
- *     API_KEY: secret(process.env.API_KEY),
- *     DATABASE_URL: secret(process.env.DATABASE_URL)
+ *     API_KEY: alchemy.secret(process.env.API_KEY),
+ *     DATABASE_URL: alchemy.secret(process.env.DATABASE_URL)
  *   }
  * });
  *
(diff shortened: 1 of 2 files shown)


alchemy/src/os/exec.ts
```diff
@@ -342,13 +369,14 @@ export async function exec(
...
-  const { glob, readFile } = await import("node:fs/promises");
+  const { glob } = await import("glob");
+  const { readFile } = await import("node:fs/promises");
 
   const hashes = new Map<string, string>();
 
   await Promise.all(
     patterns.flatMap(async (pattern) => {
-      const files = await Array.fromAsync(glob(pattern, { cwd }));
+      const files = await glob(pattern, { cwd });
```
sam-goodwin, line 379: @claude Why did this change? It seems unnecessary and irrelevant to the requested change

Mkassabov: asked claude on my own and got this

> When Each Approach Makes Sense
> Async Iterable (Node.js built-in): Better for very large result sets where you want to process files one by one without loading everything into memory
> Promise<Array> (Third-party): Better for most common use cases where you need the complete list and performance is important

not against keeping it, since I think `glob` is slightly more modern than the glob from fs

Mkassabov:
@sam-goodwin checked and confirmed this works. pushed a minor type fix to it.

alchemy/src/os/exec.ts
sam-goodwin: We need to support Node LTS (node 20+), would `glob` help us do that?

Mkassabov: Yeah fs glob is node 22, so we can't use it

sam-goodwin:
Thanks for checking it out and testing it @Mkassabov! Saved me the effort

Issue #542, opened 2025-07-03
Planetscale password


Mkassabov:
we need a planetscale password resource.

Primarily because branches are useless without the subsequent password since all each branch of a pscale database requires its own password.

https://api-docs.planetscale.com/reference/create_password

PR #551, opened 2025-07-04
Planetscale password resource

Mkassabov:
fixes: https://github.com/sam-goodwin/alchemy/issues/542


Mkassabov:
was playing around with sst's opencode.
it did like 90% of this.

alchemy/src/planetscale/password.ts
```diff
@@ -0,0 +1,198 @@
...
+    username: string;
+
+    /**
+     * The encrypted password for database authentication
+     */
+    password: Secret<string>;
+  };
+}
+
+export const Password = Resource(
```
sam-goodwin, line 254: @claude please generate the docs with "@example" usage. also create providers/planetscale/password.md following same standards

alchemy/src/planetscale/password.ts
```diff
@@ -0,0 +1,299 @@
...
+
+  /**
+   * The database name where the password will be created
+   */
+  databaseName: string;
+
+  /**
+   * The branch name where the password will be created
+   */
+  branchName?: string;
```
sam-goodwin: @claude this should be (following our tenet of always supporting both an identifier and the alchemy resource instance

database: string | Database
branch?: string | Branch

please update code, examples and docs

sam-goodwin: @claude

Mkassabov:
looking at this im realizing there is a problem with replace.
if replacing a resource that requires a unique name you just kind of can't. because the creation fails since we need to delete first.

we either want to delete first, which I don't think is a good idea since that could mean a partial outage, or do some weird thing with appending a random string to the end of names

sam-goodwin:
Cut an issue for replace conundrum and merge this Pr? Is it ready?

Mkassabov:
> Cut an issue for replace conundrum and merge this Pr? Is it ready?

yeah been using it all weekend, no real issues

PR #574, opened 2025-07-05
Improve site build performance by caching build assets and removing old OG files

zephraph:
Does what it says on the tin. We hadn't fixed CI to cache based on astro assets yet so astro was effectively building from scratch each time. Also, the OG image uploads likely take _forever_ and we're not using them anymore so this drops them entirely.


.github/workflows/pr-preview.yml
```diff
@@ -26,6 +26,14 @@ jobs:
           aws-role-arn: ${{ secrets.AWS_ROLE_ARN }}
           aws-region: us-west-2
 
+      - name: Cache Astro cache
+        uses: actions/cache@v4
+        with:
+          path: |
+            alchemy-web/node_modules/.astro
+            alchemy-web/.astro
+          key: astro-pr-cache-${{ github.event.pull_request.number }}
```
sam-goodwin, line 35: Will this work on main?

just-be-dev: I don't think I understand the question. This workflow only runs on PRs so it won't run on main. I have a different cache key in publish.yml which will work on main fine.

PR #590, opened 2025-07-07
Don't include posthog in dev

zephraph:
This just drops the posthog script contents in dev.


alchemy-web/astro.config.mjs
```diff
@@ -37,12 +37,18 @@ export default defineConfig({
...
         {
           tag: "script",
-          content: postHogScript
-            .replace(
-              "<POSTHOG_CLIENT_API_HOST>",
-              process.env.POSTHOG_CLIENT_API_HOST,
-            )
-            .replace("<POSTHOG_PROJECT_ID>", process.env.POSTHOG_PROJECT_ID),
+          content:
+            process.env.NODE_ENV === "production"
```
Mkassabov: can we set this as an env variable like `ENABLE_POSTHOG` and then have that set by the branch name in alchemy.run.ts

Mkassabov (changes requested):
Is there a reason we need to remove this in the code? can't we filter this in posthogs side?

sam-goodwin:
What's the status here? Is this ready to be merged?

alchemy-web/astro.config.mjs
zephraph: @claude will you implement this?

Mkassabov (approved):

PR #593, opened 2025-07-07
feat(sqlite): sqlite state store with drizzle

john-royal:
Closes https://github.com/sam-goodwin/alchemy/issues/599

### DIFF
--- a/alchemy/src/cloudflare/bundle/esbuild.ts
+++ b/alchemy/src/cloudflare/bundle/esbuild.ts
@@ -33,10 +33,10 @@ export class ESBuildBundleProvider implements WorkerBundleProvider {
     this.props.cwd = path.resolve(this.props.cwd);
     this.props.entrypoint = path.isAbsolute(props.entrypoint)
       ? path.relative(this.props.cwd, props.entrypoint)
-      : props.entrypoint;
+      : path.normalize(props.entrypoint);
     this.props.outdir = path.isAbsolute(props.outdir)
       ? path.relative(this.props.cwd, props.outdir)
-      : props.outdir;
+      : path.normalize(props.outdir);
   }
 
   async create(): Promise<WorkerBundle> {


alchemy/package.json
```diff
@@ -104,6 +104,22 @@
...
+      "import": "./lib/sqlite-state-store/bun-sqlite.js"
+    },
+    "./sqlite-state-store/d1": {
+      "bun": "./src/sqlite-state-store/d1.ts",
+      "import": "./lib/sqlite-state-store/d1.js"
+    },
+    "./sqlite-state-store/libsql": {
+      "bun": "./src/sqlite-state-store/libsql.ts",
+      "import": "./lib/sqlite-state-store/libsql.js"
+    },
```
sam-goodwin:
```ts
import { D1StateStore } from "alchemy/cloudflare";
import { DOSqlite3StateStore } from "alchemy/cloudflare";
import { DOFSStateStore } from "alchemy/cloudflare";

import { SQLite3StateStore } from "alchemy/sqlite";
import { FileSystemStateStore } from "alchemy/fs";
```

```ts
new SQLite3StateStore(scope, {
  engine: "libsql" | new LibSQLEngine()
})
```

sam-goodwin: Do `DOSqlite3StateStore` in a follow up PR

john-royal: LMK if this looks good to you now!

alchemy/src/sqlite/sqlite-state-store.ts
```diff
@@ -0,0 +1,158 @@
...
+interface BunSQLiteStateStoreOptions {
+  engine: "bun";
+  filename?: string;
+  options?: BunSQLiteOptions;
+}
+
+interface BetterSQLite3StateStoreOptions {
+  engine: "better-sqlite3";
+  filename?: string;
+  options?: BetterSQLite3Options;
```
sam-goodwin: We can flatten the `BetterSQLite3Options` into the interface

john-royal: Done

john-royal:
@sam-goodwin Was able to fix the concurrency errors you were having by enabling WAL and adding retries.

Excessive concurrency from different processes is still slow especially with the libsql driver (bun:sqlite actually handles it pretty well), so I kept the isolated test database files.

Also removed the `better-sqlite3` driver since it didn't test well and the file auto-remove function since concurrent processes require the `retain` option to be enabled. Not sure what the best complexity/DX tradeoff is there — the empty files hanging around is a little annoying, especially with the additional files added for WAL.

sam-goodwin:
> the empty files hanging around is a little annoying, especially with the additional files added for WAL.

What are the additional files for WAL? Is it not just .alchemy/state.sqlite ?

john-royal:
There are two additional files: `filename-wal` for the WAL itself and `filename-shm` which is a shared memory file. I think that's just how the mechanism works in SQLite.

sam-goodwin:
Not a big deal. 3 files.

john-royal:
Agreed! Think this is good to merge at your discretion, unless you run into anything I missed.

PR #602, opened 2025-07-07
fix: prevent telemetry failures from causing system failures

### LINKED ISSUE #594: Remove Error if telemetry fails
Telemetry failing shouldn't cause all of alchemy to fail

Mkassabov:
Fixes #594

Added proper error handling to prevent telemetry failures from causing the entire alchemy system to fail.

## Changes
- Add error handling in scope.init() to catch telemetry initialization failures
- Add error handling in scope.finalize() to catch telemetry finalization failures
- Both methods now log warnings instead of throwing errors, allowing the system to continue functioning

### DIFF
--- a/alchemy/src/scope.ts
+++ b/alchemy/src/scope.ts
@@ -219,7 +219,14 @@ export class Scope {
   }
 
   public async init() {
-    await Promise.all([this.state.init?.(), this.telemetryClient.ready]);
+    await Promise.all([
+      this.state.init?.(),
+      this.telemetryClient.ready.catch((error) => {
+        if (!this.quiet) {
+          console.warn("Telemetry initialization failed:", error);
+        }
+      }),
+    ]);
   }
 
   public async deinit() {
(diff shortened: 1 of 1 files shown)


sam-goodwin (approved):

alchemy/src/scope.ts
```diff
@@ -219,7 +219,14 @@ export class Scope {
   }
 
   public async init() {
-    await Promise.all([this.state.init?.(), this.telemetryClient.ready]);
+    await Promise.all([
+      this.state.init?.(),
+      this.telemetryClient.ready.catch((error) => {
+        if (!this.quiet) {
+          console.warn("Telemetry initialization failed:", error);
+        }
```
Mkassabov: @claude can we use the logger everywhere please, since the logger already checks for `quiet`

alchemy/src/scope.ts
```diff
@@ -399,7 +406,13 @@ export class Scope {
...
     }
 
-    await this.rootTelemetryClient?.finalize();
+    try {
+      await this.rootTelemetryClient?.finalize();
+    } catch (error) {
+      if (!this.quiet) {
+        console.warn("Telemetry finalization failed:", error);
+      }
+    }
```
Mkassabov: @claude just do .catch here, also use the logger here

PR #603, opened 2025-07-08
chore: only deploy proxy on main

Mkassabov:


alchemy-web/alchemy.run.ts
```diff
@@ -23,19 +23,24 @@ const app = await alchemy("alchemy:website", {
...
+    name: `${app.name}-${app.stage}-posthog-proxy`,
+    entrypoint: "src/proxy.ts",
+    domains: [POSTHOG_PROXY_HOST],
+    bindings: proxyBindings,
+  });
+}
 
 const website = await Website("website", {
-  name: "alchemy-website",
+  name: `${app.name}-${app.stage}-website`,
```
sam-goodwin: I don't think we should change this right now because we are already deployed. Although this is a great use-case that should be allowed - "rename worker".

sam-goodwin: claude revert thos

sam-goodwin: @claude revert thos

PR #610, opened 2025-07-08
feat(cloudflare): default source_map to true for Workers

### LINKED ISSUE #609: Default source_map: true
In Worker we should default to true when bundling.

sam-goodwin:
Default source maps to true for Workers so users have to opt out instead of opt in.

Closes #609

BREAKING CHANGE: rename `uploadSourceMaps` to `sourceMap`


Mkassabov:
@claude this also needs to be supported in `bundle-worker-dev`

PR #612, opened 2025-07-09
chore: fix outdated website links

john-royal:
Found some broken links scattered about our codebase. Here's a fix.

### DIFF
--- a/alchemy/src/cloudflare/user.ts
+++ b/alchemy/src/cloudflare/user.ts
@@ -1,5 +1,6 @@
 import type { Secret } from "../secret.ts";
 import { logger } from "../util/logger.ts";
+import { memoize } from "../util/memoize.ts";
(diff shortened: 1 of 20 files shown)


alchemy/src/cloudflare/user.ts
```diff
@@ -55,82 +56,73 @@ export interface CloudflareOrganization {
   roles: string[];
 }
 
-const accountCache: Record<string, CloudflareAccount[]> = {};
-
-export async function getCloudflareAccounts(
```
sam-goodwin: Why did this change?

john-royal: Great question! Stumbled upon this while working on the Cloudflare login PR. Just reverted the change (aside from the link fix) to keep this focused. My bad.

PR #619, opened 2025-07-09
feat(cloudflare): rename workers

Mkassabov:


alchemy/src/cloudflare/worker.ts
```diff
@@ -1085,6 +1085,22 @@ export const _Worker = Resource(
...
 
+    if (this.phase === "update") {
+      const oldName = this.output.name ?? this.output.id;
+      const newName = workerName;
+
+      if (oldName !== newName) {
+        const renameResponse = await api.patch(
+          `/accounts/${api.accountId}/workers/services/${oldName}`,
+          { id: newName },
+        );
```
sam-goodwin: different endpoint for wfp?

Mkassabov: this calls replace(force) now

alchemy/src/apply.ts
```diff
@@ -177,8 +185,19 @@ async function _apply<Out extends Resource>(
...
 
+        if (force) {
+          await destroy(resource, {
+            quiet: scope.quiet,
+            strategy: "sequential",
+            replace: {
+              props: state.oldProps,
+              output: resource,
+            },
+          });
```
sam-goodwin, line 196: Will this affect the state of the in-progress resource? Or no, because you set `replace`?

alchemy/test/cloudflare/worker.test.ts
```diff
@@ -1879,4 +1879,53 @@ describe("Worker Resource", () => {
       await assertWorkerDoesNotExist(api, workerName);
     }
   });
+
+  test("rename worker", async (scope) => {
```
sam-goodwin, line 1944: let's also add a test for renaming a WfP worker

alchemy/src/apply.ts
Mkassabov:
```
-- CREATE

creating - alchemy-test/dev/rename-test
creating - alchemy-test/dev/rename-test
created - alchemy-test/dev/rename-test
creating - alchemy-test/dev/taptap
creating - alchemy-test/dev/taptap
created - alchemy-test/dev/taptap

-- REPLACE

updating - alchemy-test/dev/taptap
updated - alchemy-test/dev/taptap
```

this is the result in order for .set (status - fqn).

this is exactly what I expect. since a force replace is practically just an update.
If you really want we can add a `destroy/destroyed` log around it, but that might just confuse the user?

Mkassabov: that duplicate .set call is bothering me, opened an issue: https://github.com/sam-goodwin/alchemy/issues/634

not really relevant in this PR.

sam-goodwin: Good thing to track. we should minimize the number of state calls

sam-goodwin:
I fixed the WfP rename test but the replace tests are failing

Mkassabov:
> I fixed the WfP rename test but the replace tests are failing

fixed. was missing an `await` on this.replace

PR #621, opened 2025-07-09
fix(cloudflare): prioritize explicit credentials over environment variables in createCloudflareApi

### LINKED ISSUE #620: createCloudflareApi should check all explicit overrides before defaulting to env variables
@claude please update the createCloudflareApi functionn to first check if explicit credentials (e.g. apiKey, apiToken, etc.) are set before short circuiting to a set environment variable. The current code has a bug where if CLOUDFLARE_API_TOKEN is set, then the user can't explciitly override using apiKey.

https://github.com/sam-goodwin/alchemy/blob/5816e0f5f175f7719c88105e2adf547487bec780/alchemy/src/cloudflare/api.ts#L51-L91

sam-goodwin:
Previously, if CLOUDFLARE_API_TOKEN was set in environment and user provided explicit apiKey, both would be set causing constructor to throw error.

Now explicit credentials take full precedence - if any explicit auth parameter is provided, environment variables are ignored for all auth methods.

Fixes #620

### DIFF
--- a/alchemy/src/cloudflare/api.ts
+++ b/alchemy/src/cloudflare/api.ts
@@ -61,13 +61,11 @@ export async function createCloudflareApi(
   } else if (options.apiKey !== undefined) {
     // Explicit apiKey provided
     apiKey = options.apiKey;
-  } else {
+  } else if (process.env.CLOUDFLARE_API_TOKEN) {
     // No explicit credentials, check environment variables: CLOUDFLARE_API_TOKEN > CLOUDFLARE_API_KEY
-    if (process.env.CLOUDFLARE_API_TOKEN) {
-      apiToken = alchemy.secret(process.env.CLOUDFLARE_API_TOKEN);
-    } else if (process.env.CLOUDFLARE_API_KEY) {
-      apiKey = alchemy.secret(process.env.CLOUDFLARE_API_KEY);
-    }
+    apiToken = alchemy.secret(process.env.CLOUDFLARE_API_TOKEN);
+  } else if (process.env.CLOUDFLARE_API_KEY) {
+    apiKey = alchemy.secret(process.env.CLOUDFLARE_API_KEY);
   }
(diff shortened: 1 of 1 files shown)


alchemy/src/cloudflare/api.ts
```diff
@@ -51,16 +51,22 @@ export interface CloudflareApiOptions {
...
+    : (options.apiKey ??
+      (process.env.CLOUDFLARE_API_KEY
+        ? alchemy.secret(process.env.CLOUDFLARE_API_KEY)
+        : undefined));
+  const apiToken = hasExplicitAuth
+    ? options.apiToken
+    : (options.apiToken ??
+      (process.env.CLOUDFLARE_API_TOKEN
+        ? alchemy.secret(process.env.CLOUDFLARE_API_TOKEN)
+        : undefined));
```
sam-goodwin: @claude Super weird way to do it. First check all of the explciitly provided options and choose the first one that it finds. If it finds none, then check all of the env variables in order of CLOUDFLARE_API_TOKEN > CLOUDFLARE_API_KEY

alchemy/src/cloudflare/api.ts
```diff
@@ -51,16 +51,24 @@ export interface CloudflareApiOptions {
...
+    // Explicit apiKey provided
+    apiKey = options.apiKey;
+  } else {
+    // No explicit credentials, check environment variables: CLOUDFLARE_API_TOKEN > CLOUDFLARE_API_KEY
+    if (process.env.CLOUDFLARE_API_TOKEN) {
+      apiToken = alchemy.secret(process.env.CLOUDFLARE_API_TOKEN);
+    } else if (process.env.CLOUDFLARE_API_KEY) {
+      apiKey = alchemy.secret(process.env.CLOUDFLARE_API_KEY);
+    }
+  }
```
sam-goodwin: @claude these can be chained as `else if`

john-royal:
Replacing with #667

PR #630, opened 2025-07-10
feat(cloudflare): SQLite-backed durable object state store

john-royal:


alchemy-web/src/content/docs/guides/do-state-store.mdx
```diff
@@ -78,7 +82,7 @@ DOStateStore provides a [State Store](/concepts/state-store) using a Cloudflare
      </TabItem>
    </Tabs>
 
-   Alchemy automatically creates an `alchemy-state` worker in your Cloudflare account.
+   Alchemy automatically creates an `alchemy-state-sqlite` worker in your Cloudflare account.
```
sam-goodwin: Not sure we should rename this ... sqlite is an implementation detail.

alchemy-web/src/content/docs/guides/do-state-store.mdx
```diff
@@ -145,10 +129,7 @@ Force the worker to be recreated even if it exists:
 ```typescript
 const app = await alchemy("my-app", {
   stateStore: (scope) => new DOStateStore(scope, {
-    worker: {
-      name: "my-app-state",
-      force: true
-    }
+    forceUpdate: true
```
sam-goodwin, line 132: can we start hashing the bundled worker and storing that as a version tag on the worker and use that to determine when to update rather than this `forceUpdate: true` flag?

alchemy-web/src/content/docs/guides/do-state-store.mdx
sam-goodwin: I think even if someone uses the existing worker + DO, it would probably still work since it's a different sqlite table. If you're extra worried, we could name the DO different.

alchemy/src/cloudflare/do-state-store.ts
```diff
@@ -0,0 +1,145 @@
...
+  forceUpdate?: boolean;
+  /**
+   * The token to use for the state store.
+   * @default process.env.ALCHEMY_STATE_TOKEN
+   * @note You must use the same token for all deployments on your Cloudflare account.
+   */
+  stateToken?: Secret<string>;
+}
+
+const TEMPLATE_WORKER_NAME = "do-sqlite-state-store" as const;
```
sam-goodwin: what is this?

alchemy/src/cloudflare/types.ts
```diff
@@ -68,3 +73,29 @@ export async function extractCloudflareError(
     return `HTTP ${response.status}: ${response.statusText}`;
   }
 }
+
+export async function extractCloudflareResult<T>(
```
sam-goodwin, line 77: Is this creating better stack traces for all cloudflare or just this?

Seems a bit odd in `types.ts`. I see we have `extractCloudflareError` too, not sure if I wrote that.

alchemy/src/cloudflare/worker-subdomain.ts
```diff
@@ -91,33 +82,48 @@ export async function disableWorkerSubdomain(
...
+  scriptName: string,
+) {
+  await extractCloudflareResult<SubdomainResponse>(
+    `enable subdomain for "${scriptName}"`,
+    api.post(
+      `/accounts/${api.accountId}/workers/scripts/${scriptName}/subdomain`,
+      { enabled: true, previews_enabled: true },
+    ),
+  );
+}
```
sam-goodwin, line 96: Did we already have a function for this? Let's make sure we aren't duplicating logic.

alchemy/src/cloudflare/types.ts
sam-goodwin: I like where this is going. The current `handleApiError` is just so bad.

alchemy/src/cloudflare/worker-subdomain.ts
```diff
@@ -91,33 +82,48 @@ export async function disableWorkerSubdomain(
...
+
+export async function getWorkerSubdomain(
```
sam-goodwin, line 98: Same with this function.

alchemy/src/cloudflare/worker/shared.ts
```diff
@@ -1,30 +1,116 @@
...
+    format: "esm",
+    nodeCompat: null,
+  });
+  return provider.create(false).then((bundle) => ({
+    ...bundle,
+    tag: `${name}:${bundle.hash}`,
+  }));
+});
+
+export async function getTemplateWorkerStatus(
```
sam-goodwin: I am having a hard time understanding the layout of all this. Am jumping around multiple files and not finding any logical structure. What is `worker/shared.ts`? Shared amongst all workers? Or just our internal workers?

I don't know what a template worker is. Is DOStateStore a "template worker"?

sam-goodwin: Perhaps we should co-locate all state stores together in `alchemy/state` that way we can share everything amongst them there.

alchemy/src/sqlite/proxy.ts
```diff
@@ -0,0 +1,132 @@
+import type { Scope } from "../scope.ts";
+import type { State, StateStore } from "../state.ts";
+import { withExponentialBackoff } from "../util/retry.ts";
+
+/**
+ * Wraps a local or remote state store implementation
+ * and handles serialization/deserialization.
+ */
+export abstract class StateStoreProxy implements StateStore {
```
sam-goodwin: Nice!

alchemy/src/sqlite/proxy.ts
```diff
@@ -0,0 +1,132 @@
...
+        () => dispatch(method, params),
+        () => true,
+      );
+      return result;
+    });
+  }
+
+  init(): Promise<void> {
+    return this.run("init", []);
+  }
```
sam-goodwin: Couldn't we do this with an actual proxy instead of enumerating each of the methods?

alchemy/src/sqlite/proxy.ts
```diff
@@ -0,0 +1,132 @@
...
+        Object.entries(states).map(async ([key, value]) => [
+          key,
+          await this.deserialize(value),
+        ]),
+      ),
+    );
+  }
+}
+
+export namespace StateStoreProxy {
```
sam-goodwin: it's type only, so `export declare namespae`

alchemy-web/src/content/docs/guides/do-state-store.mdx
sam-goodwin: We could also write a single worker that handles the different state stores (SQLite, DOFS, D1 via Binding). One worker for all?

alchemy-web/src/content/docs/guides/do-state-store.mdx
john-royal: We do! The force update option is just a break glass mechanism, mainly useful if you lose your token and need to force recreate it with the new one.

alchemy/src/cloudflare/types.ts
john-royal: I didn't write this file. I just added the `extractCloudflareResult` function here because:
- I noticed similar functionality being duplicated across 10,000 different locations
- I didn't want to add duplication 10,001
- the `CloudflareApiResponse<T>` type was already defined here, as well as an `extractCloudflareError` fn

I renamed the file to `api-response.ts` since that seems more descriptive.

Was hoping to use this mechanism to help standardize errors and stack traces across the board, but figured I'd start with places that are doing this manually instead of using `handleApiError`.

alchemy/src/cloudflare/worker-subdomain.ts
john-royal: I searched the codebase. This is the only one that calls this endpoint.

alchemy/src/cloudflare/worker-subdomain.ts
john-royal: This is the only one. I noticed there were a few of these floating around so I thought it made sense to consolidate in `worker-subdomain.ts`.

alchemy/src/sqlite/proxy.ts
john-royal: Didn't know that was a thing! Just changed that.

alchemy/src/cloudflare/worker/shared.ts
john-royal: Renamed them to "internal workers" - agree that makes more sense. There are two now that I removed the legacy DOFS state store:
- `do-state-store`
- `remote-binding-proxy`

The only shared file for those two is `internal-worker-bundle.ts` (extracted that from the old `shared` file, which you're right was a code smell), which just reads the file and returns the file formatted for CF (i.e. with `application/javascript+module`), along with a tag that we can use to know when to update the worker. Piggybacks on the FSBundleProvider we already have.

alchemy/src/sqlite/proxy.ts
john-royal: Great question! The reason why I did it this way is because the serialization logic depends on the method. For example:
- `set()` serializes the input value
- `get()` deserializes a single output value
- `all()` and `getBatch()` contain multiple output values that each need to be deserialized separately (or can they be deserialized together?)

Perhaps we can just run everything through serde instead so it can just be a proxy? The other thing is that I figured we'd want all the state stores to be classes (since the `new StateStore()` syntax is currently standard across the board), which (I think) wouldn't be possible if they were wrapping a proxy instead of inheriting another class.

alchemy/src/cloudflare/do-state-store.ts
john-royal: Removed

alchemy-web/src/content/docs/guides/do-state-store.mdx
john-royal: I wanted to talk about this since:
- On one hand, SQLite is 100% an implementation detail and ideally we wouldn't pollute your account with yet another worker
- On the other hand, this worker is different from the previous one and I don't want to accidentally nuke their old state without a warning
- Sometimes if you try to access a DO when the worker code or metadata has changed, Cloudflare will throw an error saying it's been reset since the code changed — figured probably best to avoid cross-pollination between the new implementation and the old, broken one

Not sure what's the right thing to do here. Do you have any ideas?

Maybe a CLI script that actually migrates everything? Though I'd be wary of footguns there.

alchemy/src/sqlite/proxy.ts
sam-goodwin: I'm fine with how it is. Nothing wrong with explicit. 

FYI - it is possible to use `new` syntax and a Proxy by returning it from the constructor. Black magic stuff but I've used this to great success lol

```ts
export class StateStoreProxy {
  constructor() {
    return new Proxy(this, {
      ..
    });
  }
}
```

alchemy-web/src/content/docs/guides/do-state-store.mdx
sam-goodwin: At this point I'm thinking the old store is just totally corrupted and we should just rip the band aid off.

sam-goodwin: Delete it and never look back. Users can recover by adoption and I doubt anyone is using it without bugs anyway.

alchemy/src/sqlite/proxy.ts
john-royal: Oh wow, good to know. Feels hacky to return from a constructor but definitely can see the use case for that.

alchemy-web/src/content/docs/guides/do-state-store.mdx
john-royal: Fair enough! I'd much rather be macOS and delete the old broken thing than be Windows and keep building workarounds to prop up the legacy stuff. Just updated the name.

PR #631, opened 2025-07-10
feat(cloudflare): login automatically or via cli

john-royal:
More robust implementation than #475. Uses `neverthrow`, which I'm interested in using more broadly to improve our error handling — seems like a simpler way to improve our error handling if we're not ready for an Effect overhaul.

The biggest cause of #438 seems to be a race condition where multiple API calls will try to refresh at the same time using the same token, which Cloudflare balks at because you can't reuse a refresh token. This fixes that with a single-flight mechanism.

~~Was having trouble getting the CLI to build.~~ Edit: it's because importing the default logger in a file imported by the CLI was blowing a hole in the space-time continuum.

Also have to update the redirect URLs — I added them to our website just to show something when you log in.


alchemy/src/cloudflare/auth-wrangler.ts
```diff
@@ -0,0 +1,314 @@
...
+import open from "open";
+import { HTTPServer } from "../util/http-server.ts";
+import { memoize, memoizeSync } from "../util/memoize.ts";
+import { detached, ensure, singleFlight } from "../util/neverthrow.ts";
+import { createXdgAppPaths } from "../util/xdg-paths.ts";
+
+const CLIENT_ID = "54d11594-84e4-41aa-b438-e81b8fa78ee7";
+const REDIRECT_URI = "http://localhost:8976/oauth/callback";
+const ALCHEMY_BASE_URL = "https://alchemy.run";
+export const DEFAULT_SCOPES = [
```
sam-goodwin, line 16: @john-royal i wonder if we can make this configurable/expand to other scopes like creating auth tokens

sam-goodwin: https://github.com/sam-goodwin/alchemy/issues/662

PR #633, opened 2025-07-10
chore(core): telemetry for state store + resource + context data

Mkassabov:


alchemy/src/state/state-store-wrapper.ts
```diff
@@ -0,0 +1,127 @@
...
+  fn: () => Promise<T>,
+): Promise<{ result: T; elapsed: number }> {
+  const start = performance.now();
+  const result = await fn();
+  const elapsed = performance.now() - start;
+  return { result, elapsed };
+}
+
+//todo(michael): we should also handle serde here
+export class StateStoreWrapper implements StateStore {
```
john-royal: Wonder if we should fold this into the [`StateStoreProxy`](https://github.com/sam-goodwin/alchemy/pull/630/files#diff-ff7020a6542a9b29bbd391fff6e32da9e5e8052508b51e8f1063774778d29ff3) - though the proxy isn't currently being used for every state store type, I suppose it can

john-royal: That or make this a proper proxy that just instruments the methods. I'd lean towards that.

alchemy/src/state/state-store-wrapper.ts
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  async init() {
+    if (this.stateStore.init == null) {
+      return;
+    }
+    const { elapsed } = await callWithPerformance(
+      this.stateStore.init.bind(this.stateStore),
+    );
+    this.telemetryClient.record({
```
john-royal: Should definitely also record if there's an error

alchemy/src/state/state-store-wrapper.ts
Mkassabov: my thinking was that this can also wrap serde that way state stores can never run into the issue of not serializing properly. IIRC @sam-goodwin mentioned wanting to be extracted into its own layer.

I worry if its just for telemetry then we end up with like 3 wrappers for state store when it can really just be one

alchemy/src/state/state-store-wrapper.ts
Mkassabov: done, also moved that wrapper to be a private method

alchemy/src/state/state-store-wrapper.ts
sam-goodwin: > I worry if its just for telemetry then we end up with like 3 wrappers for state store when it can really just be one

Is that an argument for using StateStoreProxy? Ideally we'd centralize this functionality.

Mkassabov: Yeah argument for using state store proxy

sam-goodwin: Ok, so will hold off on this PR until we align our state stores to be all under StateStoreProxy? Who is owning that work?

alchemy/src/util/telemetry/client.ts
```diff
@@ -54,13 +54,9 @@ export class TelemetryClient implements ITelemetryClient {
   }
 
   record(event: Telemetry.EventInput, timestamp = Date.now()) {
-    if (!this.context) {
-      return;
-    }
```
sam-goodwin, line 59: Why did this change? Looks like we have a `this.context!` below which is dangerous and could crash?

alchemy/src/util/telemetry/client.ts
```diff
@@ -96,6 +92,7 @@ export class TelemetryClient implements ITelemetryClient {
     if (events.length === 0) {
       return;
     }
+    const { userId, ...data } = this.context!;
```
sam-goodwin, line 95: what happens if `this.context` is undefined? Why remove the early return?

alchemy/src/util/telemetry/client.ts
Mkassabov: no clue, brought it back

alchemy/src/util/telemetry/client.ts
Mkassabov: no clue, brought the early return back

alchemy/src/util/telemetry/client.ts
Mkassabov: ohhh wait no I remember. because context is not on every event anymore. this check should've been moved to `send` not removed. add the check back on `send`

PR #641, opened 2025-07-11
fix(cloudflare): replace `dev: "prefer-local|remote"` with `local: boolean`, `watch: boolean`

sam-goodwin:
- [x] Replace `dev: "prefer-local|remote"` with `local: boolean`, `watch: boolean`
- [x] Parse the `http://localhost:*` output of dev commands instead of hard-code
- [x] Fix the ReactRouter template
- [x] Update smoke tests to also check that `dev` works before deployment for every smoke test

### DIFF
--- a/alchemy/src/util/safe-fetch.ts
+++ b/alchemy/src/util/safe-fetch.ts
@@ -20,6 +20,7 @@ export async function safeFetch(
         err?.code === "UND_ERR_SOCKET" ||
         err?.code === "ECONNRESET" ||
         err?.code === "UND_ERR_CONNECT_TIMEOUT" ||
+        err?.code === "EPIPE" ||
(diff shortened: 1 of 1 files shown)


alchemy/src/cloudflare/worker.ts
```diff
@@ -1191,7 +1199,7 @@ export const _Worker = Resource(
     }
 
     let putWorkerResult: PutWorkerResult;
-    if (dev.type === "remote") {
+    if (watch) {
```
Mkassabov, line 1229: why is this here? isn't watch only for watching resource changes?

sam-goodwin: That's what this does. It watches for changes to the worker bundle

PR #642, opened 2025-07-11
fix(cloudflare): remote worker tail crash

john-royal:


alchemy/src/cloudflare/worker.ts
```diff
@@ -1219,8 +1219,13 @@ export const _Worker = Resource(
         }),
       );
       putWorkerResult = await promise.value;
-      const tail = await createTail(api, id, workerName);
-      cleanups.push(() => tail.close());
+      createTail(api, id, workerName)
```
sam-goodwin: Should we await this

alchemy/src/cloudflare/worker/tail.ts
```diff
@@ -92,7 +93,9 @@ class TailClient {
         return;
       }
       logger.log(`reconnecting to tail for "${this.scriptName}"`);
-      this.connect(attempt + 1);
+      setTimeout(() => {
+        this.connect(attempt + 1);
+      }, 1000);
```
sam-goodwin: This looks concerning. What problem does it solve? Does it guarantee to solve it or just mitigate?

PR #643, opened 2025-07-11
fix(cloudflare): add duplex: "half" in miniflare http server

john-royal:


alchemy/src/cloudflare/miniflare/http-server.ts
```diff
@@ -59,6 +59,9 @@ function toWebRequest(request: http.IncomingMessage): Request {
       ["GET", "HEAD", "OPTIONS"].includes(method) || !request.readable
         ? undefined
         : (Readable.toWeb(request) as unknown as BodyInit),
+    // TODO: why is Cloudflare's request type showing up here?
```
sam-goodwin, line 62: Because of @cloudflare/workers-types

sam-goodwin (approved):

alchemy/src/cloudflare/miniflare/http-server.ts
john-royal: Is there a way to keep those types isolated to actual worker code? Interesting that it bleeds into our Node code

PR #650, opened 2025-07-12
fix(cloudflare): update keyword for adopting an existing vectorize index

sbekrin:
currently, alchemy fails to adopt existing vectorize index because cloudflare returns `vectorize.index.duplicate_name - Index name "<index-name>"` message which is missing the target _already exists_ keyword

### DIFF
--- a/alchemy/src/cloudflare/vectorize-index.ts
+++ b/alchemy/src/cloudflare/vectorize-index.ts
@@ -169,7 +169,7 @@ const _VectorizeIndex = Resource(
         if (
           props.adopt &&
           error instanceof CloudflareApiError &&
-          error.message.includes("already exists")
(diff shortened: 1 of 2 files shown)


alchemy/src/cloudflare/vectorize-index.ts
```diff
@@ -169,7 +169,7 @@ const _VectorizeIndex = Resource(
         if (
           props.adopt &&
           error instanceof CloudflareApiError &&
-          error.message.includes("already exists")
+          error.message.includes("vectorize.index.duplicate_name")
```
sam-goodwin, line 172: We should add a test for this case.

sam-goodwin: Pushed a commit with the test, looking good. Also found a bug with delete where we were not idempotent

PR #664, opened 2025-07-14
chore(docs): return replace docs

Mkassabov:
At some point we lost these docs.

Tying this to [PR#619](https://github.com/sam-goodwin/alchemy/pull/619) since we make replace require an await call


sam-goodwin:
Oh wait, I removed the replace docs and moved them to resource.md (along with adopt) to keep the doc tree a bit more focused. https://alchemy.run/concepts/resource/#replacement

sam-goodwin:
Oops, maybe not close it. But we should instead update resource.md with the `this.replace(true)`

@claude

Mkassabov:
> Oh wait, I removed the replace docs and moved them to resource.md (along with adopt) to keep the doc tree a bit more focused. https://alchemy.run/concepts/resource/#replacement

image


Is there a reason its not on the sidebar on the right? I swear I checked this but I guess I missed it cause I didn't scroll all the way down?

I guess claude fixed that too. huh

sam-goodwin:
> Is there a reason its not on the sidebar on the right? I swear I checked this but I guess I missed it cause I didn't scroll all the way down?

Probably using a ### instead of a ## ? We should fix it

sam-goodwin:
@claude pleae make sure replace and adopt show up on side bar by using ## instead of ###

Mkassabov:
thanks claude <3

PR #665, opened 2025-07-14
feat(cloudflare): CloudflareStateStore and deprecate DOStateStore

### LINKED ISSUE #659: DO State store deprecation
Bring back DOStateStore and leave it as `alchemy-state`

Rename the SQLite implementation to CloudflareStateStore and change its name to `alchemy-state-service`. 

First pass is just SQLite. 

In a follow up, we can also expose D1 and R2 as "backends" implemented through bindings. This should hopefully be faster.

john-royal:
- [x] Rename the SQLite-backed durable object store to `CloudflareStateStore`
- [x] Restore DOFS-backed store in `alchemy/state` directory with deprecation flag
- [x] Re-export as `DOStateStore` from `alchemy/cloudflare` for backwards compat
- [x] Update docs
- [x] **Question:** Should we include this in tests even though it's deprecated?

Closes #659


sam-goodwin:
> Should we include this in tests even though it's deprecated?

Yeah I think we should. Just to have some confidence we brought it back properly.

sam-goodwin:
By tests i mean just run `scope.test.ts` and `replace.test.ts` with it.

PR #667, opened 2025-07-14
chore(cloudflare): refactor oauth, improve error handling

john-royal:
- Rename `auth-wrangler.ts` to `oauth.ts`
- Simplify Cloudflare auth config
- Prioritize explicit configuration over environment variables (can replace #621)
- Improve error messages
- Log warning if more than 1 account ID is found

Question: In the original implementation, it looks like we prioritize API keys over API tokens. Assuming this is intentional so I kept this behavior.


sam-goodwin:
> Question: In the original implementation, it looks like we prioritize API keys over API tokens. Assuming this is intentional so I kept this behavior.

Probably an arbitrary decision, not sure if right.

Issue #686, opened 2025-07-16
Use Cloudflare's latest "compatibility date" version instead of current day


sam-goodwin:
When we build and publish a new version of Alchemy, we generate a `.ts` file with the current date that we use as the default compatibility date for cloudflare workers.

See:

https://github.com/sam-goodwin/alchemy/blob/main/alchemy/scripts/generate-build-date.ts

This results in a warning when running `alchemy dev` that every user is going to see:

> The latest compatibility date supported by the installed Cloudflare Workers Runtime is "2025-07-09",
but you've requested "2025-07-15". Falling back to "2025-07-09"...

Instead, we should get this information from cloudflare. Not sure if that's hosted on a URL somewhere or packaged in the `miniflare` or `workerd` packagers.

PR #693, opened 2025-07-16
fix(cloudflare): correct workers compatibility date

john-royal:
Uses the [same method as the `create-cloudflare` package](https://github.com/cloudflare/workers-sdk/blob/9f0c175ab668217f78debab4dfdb3e677040b9b0/packages/create-cloudflare/src/helpers/compatDate.ts#L19) to generate the default Workers compatibility date.

Placed the file in `cloudflare/compatibility-date.gen.ts` to signal that it's a generated file.

(Also I noticed that we had two `build-cli.ts` scripts doing the same thing so when I was alerted to it by a Biome error I couldn't resist fixing it, sorry.)


sam-goodwin:
Does this close https://github.com/sam-goodwin/alchemy/issues/686

john-royal:
Yup just linked the issue. my bad

PR #696, opened 2025-07-17
chore(biome): make biome more stable

Mkassabov:


bun.lock
```diff
@@ -15,6 +14,7 @@
         "braintrust": "*",
         "changelogithub": "^13.15.0",
         "glob": "^11.0.3",
+        "@biomejs/biome": "https://pkg.pr.new/biomejs/biome/@biomejs/biome@a6874f9",
```
sam-goodwin: How long until official version?

Mkassabov: no clue. will ask in the thread.
other option is to downgrade to 2.0.4 or 2.0.6.
the thread says 2.0.6 is stable. our monorepo is pinned on 2.0.4 and is def stable

Mkassabov:
relevant thread in biome: https://github.com/biomejs/biome/issues/6784

bun.lock
Mkassabov: https://github.com/biomejs/biome/issues/6784#issuecomment-3082022887

Mkassabov: @sam-goodwin 
> We generally don't provide ETAs, but I think its safe to say that at least a patch release will be soon.

will keep an eye on the thread and open a new PR.
I'm also happy to just manually override this for myself until they get an actual patch out if you don't want a dependency like this

PR #697, opened 2025-07-17
feat(cloudflare): support durable object websockets in miniflare

### LINKED ISSUE #695: Durable Object websockets don't work in Miniflare
Found this while working on the LiveStore implementation. I think this is an easy fix — we already proxy HTTP, just need to proxy websockets too. Working on it now.

john-royal:
Closes #695.

The way the proxy mechanism works for websockets is a little hacky — interacting with `response.webSocket` directly works in Node but not Bun, so what this does instead is intercept websocket connections on the HTTP proxy, grab Miniflare's internal URL for the worker, and proxy upgrade requests directly to there. The reconnect logic is extremely limited and only attempts to reconnect when the worker hot reloads, so the client connection is not interrupted.

I also included an extremely minimal example just to show that websockets do work.

I do want to clean up that Miniflare directory, but that can be done in a separate PR. This PR is mainly to unblock LiveStore support — and a plug a gaping hole in our local dev implementation.


alchemy/src/cloudflare/miniflare/miniflare-worker-options.ts
```diff
@@ -232,7 +234,9 @@ export async function buildMiniflareWorkerOptions({
     ),
     compatibilityDate,
     compatibilityFlags,
-    unsafeDirectSockets: [{ entrypoint: undefined, proxy: true }],
+    // This was true for compatibility with the dev registry, but that doesn't work anyway,
+    // and if we don't set it to false, websocket connections fail. I'm confused.
+    unsafeDirectSockets: [{ proxy: false }],
```
sam-goodwin, line 239: Do you know what else this decision will affect? Will multiple miniflare instances pointing at the same state dir still work?

alchemy/src/util/http.ts
```diff
@@ -0,0 +1,94 @@
+import http from "node:http";
+import type { AddressInfo } from "node:net";
+import { Readable } from "node:stream";
+
+export class HTTPServer {
```
sam-goodwin, line 5: Can you add some docs describing why we need this and what problem it solves?

alchemy/src/cloudflare/miniflare/miniflare-worker-proxy.ts
```diff
@@ -0,0 +1,98 @@
...
+  /**
+   * Used to proxy websocket connections to the worker.
+   * (This is a hack; used because trying to interact with `response.webSocket` doesn't work on Bun.)
+   */
+  getDirectURL: () => Promise<URL>;
+  /** Used to proxy HTTP requests to the worker. */
+  fetch: (request: Request) => Promise<Response>;
+}
+
+export class MiniflareWorkerProxy extends HTTPServer {
```
sam-goodwin, line 15: is this different than the `http`.ts` file?

alchemy/src/util/http.ts
sam-goodwin: I don't love it belonging in the `util` folder (or even having a util folder`). 

Is it related to the miniflare worker proxy?

sam-goodwin: Looks like miniflare worker proxy extends it, is there a good reason to prematurely abstrat with the base class instead of just direct implementation in the miniflare worker proxy ?

john-royal: It's used elsewhere too — Cloudflare OAuth is one (since we have to spin up a server to receive the callback) and the remote binding proxy (which currently doesn't proxy websockets) is another. It's not a new file — I just renamed it from `http-server.ts` to `http.ts` for... reasons I don't quite remember.

alchemy/src/cloudflare/miniflare/miniflare-worker-proxy.ts
john-royal: Yes, the `HTTPServer` spins up a `node:http` server with a given `fetch` handler (basically `Bun.serve` but cross platform) while this has some Miniflare-specific logic for proxying websockets as well.

alchemy/src/cloudflare/miniflare/miniflare-worker-options.ts
john-royal: The dev registry is for service bindings, while the state directory handles everything else (D1, KV, R2, etc).

The `unsafeDirectSockets` property only affects the dev registry (i.e. service bindings) and has no effect on the state directory.

alchemy/src/util/http.ts
john-royal: This isn't a new file; I just renamed it for... reasons I don't quite remember. We can change it back.

All this does is wrap a `node:http` server to accept a `fetch`-based handler — think `Bun.serve` but for Node.

Wouldn't quite call this premature abstraction because it's used in two other places without being extended:
- Cloudflare OAuth (to handle the callback)
- Remote binding proxy (those are just HTTP requests, no websocket handler needed)

alchemy/src/cloudflare/miniflare/miniflare-worker-options.ts
sam-goodwin: So does that mean service bindings don't work locally?

john-royal: Sorry, let me clarify: the dev registry is for service bindings to external Miniflare instances (e.g. the CF Vite plugin). Doesn't work right now, this doesn't change that. We're in touch with the CF team about that.

Local service bindings don't go through the registry since they run within the same Miniflare instance.

PR #710, opened 2025-07-18
feat(cli): add init command for existing projects

AmanVarshney01:
wip


alchemy/bin/commands/init.ts
```diff
@@ -0,0 +1,314 @@
...
+  try {
+    const packageJson = await fs.readJson(packageJsonPath);
+    const allDeps = {
+      ...packageJson.dependencies,
+      ...packageJson.devDependencies,
+      ...packageJson.peerDependencies,
+    };
+
+    if ("@redwoodjs/core" in allDeps || "@redwoodjs/web" in allDeps) {
+      return "rwsdk";
```
sam-goodwin: I think you're looking for `rwsdk`?

alchemy/bin/commands/init.ts
```diff
@@ -0,0 +1,314 @@
...
+      return "sveltekit";
+    }
+
+    if ("@tanstack/react-start" in allDeps) {
+      return "tanstack-start";
+    }
+
+    if ("vite" in allDeps) {
+      return "vite";
+    }
```
sam-goodwin: Would vite.config.ts be more reliable?

alchemy/bin/commands/init.ts
```diff
@@ -0,0 +1,314 @@
...
+      return "astro";
+    }
+
+    if ("nuxt" in allDeps) {
+      return "nuxt";
+    }
+
+    if ("react-router-dom" in allDeps || "react-router" in allDeps) {
+      return "react-router";
+    }
```
sam-goodwin: Why not just look for react-router?

Can you add a comment to each of these justifying it.

sam-goodwin (commented):
Looking awesome! Had some questions about our heuristics and also can you add a test init.test.ts that will run detection against each of the repository variants? Something like create.test.ts

PR #714, opened 2025-07-18
fix(cloudflare): certificate pack no longer returns undefined properties and can be replaced

Mkassabov:
- removes ability to update `cloudflareBranded` since it's not actually able to be updated (despite what the cloudflare api lets you do)
- replaces cert pack on any update
- fixed cert pack tests
- removed certain properties from cert pack that always returned nothing

### DIFF
--- a/alchemy/src/cloudflare/certificate-pack.ts
+++ b/alchemy/src/cloudflare/certificate-pack.ts
@@ -298,12 +298,17 @@ export const CertificatePack = Resource(
         const deleteResponse = await api.delete(
           `/zones/${zoneId}/ssl/certificate_packs/${this.output.id}`,
         );
+        const deleteResponseJson = (await deleteResponse.json()) as {
+          errors: Array<{ code: number }>;
+        };
+        const isCurrentlyDeleting =
+          deleteResponse.status === 400 &&
+          deleteResponseJson.errors[0].code === 1406;
 
         if (
           !deleteResponse.ok &&
           deleteResponse.status !== 404 &&
-          //* if certificate is mid deletion cloudflare errors with 400
-          deleteResponse.status !== 400
+          !isCurrentlyDeleting
         ) {
           await handleApiError(
             deleteResponse,


alchemy/src/cloudflare/certificate-pack.ts
```diff
@@ -299,7 +299,12 @@ export const CertificatePack = Resource(
           `/zones/${zoneId}/ssl/certificate_packs/${this.output.id}`,
         );
 
-        if (!deleteResponse.ok && deleteResponse.status !== 404) {
+        if (
+          !deleteResponse.ok &&
+          deleteResponse.status !== 404 &&
+          //* if certificate is mid deletion cloudflare errors with 400
+          deleteResponse.status !== 400
```
Mkassabov: I worry that this just isn't a great check, but IDK what else we can do here since the 400 error message is kind of bad

alchemy/src/cloudflare/certificate-pack.ts
```diff
@@ -314,96 +319,7 @@ export const CertificatePack = Resource(
...
-
-      const type = props.type || "advanced";
-      if (type !== currentPack.type) {
-        throw new Error(
-          `Cannot change type from '${currentPack.type}' to '${type}'. Type is immutable after creation. You must delete and recreate the certificate pack.`,
-        );
-      }
-
-      // Only cloudflareBranding can be updated
-      if (props.cloudflareBranding !== currentPack.cloudflareBranding) {
```
Mkassabov, line 355: cloudflare branding adds a cloudflare url to the certificate, so you can't actually update it despite what the api lets you do

alchemy/src/cloudflare/certificate-pack.ts
sam-goodwin: Have a look at the errors and the message data in the response. They will provide an error code you can narrow this on.

Mkassabov: this is the response
```
{
  "success": false,
  "errors": [
    {
      "code": 1406,
      "message": "Bad response certificate service"
    }
  ],
  "messages": [],
  "result": null
}
```

I'm like 95% sure I saw this message before but I'm not sure if it was with this error code. do you know if cloudflare error codes always align with the message?

will add the check here but I'm not sure thats enough to guarantee deletion

sam-goodwin: I don't know. Add a fat log here so we can see the response as we use it and adapt accordingly or escalate to Cloudflare.

sam-goodwin (approved):

alchemy/src/cloudflare/certificate-pack.ts
Mkassabov: looking at this:
https://cfapi.centminmod.com/#certificate-packs-delete-advanced-certificate-manager-certificate-pack

it seems that the message isn't unique so I think I'm okay just checking for the specific error code

sam-goodwin: Code should be more stable

PR #722, opened 2025-07-21
feat(cloudflare): local d1 migrations

### LINKED ISSUE #717: migrations not run in local db
When following the https://alchemy.run/guides/drizzle-d1/ tutorial, I tried to use it locally with `alchemy dev`

While the logs say a migration is applied
Image

In reality, the sqltie DB inside `.alchemy/miniflare/d1/miniflare-D1DatabaseObject` doesn't have them applied

john-royal:
Closes #717.

Question: How do we make sure migrations are run for existing databases (since the props haven't changed)? Should we run them outside of the resource handler instead?


sam-goodwin:
> How do we make sure migrations are run for existing databases (since the props haven't changed)? Should we run them outside of the resource handler instead?

Interesting question. Perhaps we need a way to "force" when in dev mode. We just added the feature `force` to apply updates even if there are no changes.

> Should we disable running the remote migration when running locally? Leaning towards no but not sure. Ideally we wouldn't create the remote D1 database at all if you're running locally.

If we're creating the remote DB even when running locally, that should be changed, yeah. Same for all Cloudflare resources like R2.

sam-goodwin:
When we compare properties to see if a resource should skip or not, maybe we should include Scope.dev in that comparison? 

It would cause a lot of resources to update that don't need to though.

john-royal:
We could also run the migration outside of the resource itself temporarily?

sam-goodwin:
> We could also run the migration outside of the resource itself temporarily?

Just for dev mode?

john-royal:
Yeah just for dev mode

sam-goodwin:
Does this problem only affect migrations? 

I'm fine with moving migrations out for dev mode. We can always change our mind. 

But I'm curious if this problem will be pervasive? Should we fix it at its core by ensuring resources that accept dev mode always trigger in dev mode?

Eg it's only not affecting Worker because we have alwaysUpdate: true

sam-goodwin:
> Should we disable running the remote migration when running locally? Leaning towards no but not sure. Ideally we wouldn't create the remote D1 database at all if you're running locally.

We should only create it if remote:true I assume.

PR #725, opened 2025-07-21
fix(cloudflare): don't require r2 access credentials to auto-empty bucket

### LINKED ISSUE #723: Destroy `R2Bucket` with `empty: true` fails without `R2_ACCESS_KEY_ID` and `R2_SECRET_ACCESS_KEY`
Config:
```ts
const r2Bucket = await R2Bucket("bucket", {
  name: `${project}-${app.stage}-bucket`,
  empty: true,
  adopt: true,
});
```

```
error: R2_ACCESS_KEY_ID and R2_SECRET_ACCESS_KEY environment variables are required
      at createR2Client (/Users/ianpascoe/Code/unnamed/node_modules/alchemy/src/cloudflare/bucket.ts:297:11)
      at <anonymous> (/Users/ianpascoe/Code/unnamed/node_modules/alchemy/src/cloudflare/bucket.ts:212:34)
```

The docs do not say that this is required:
https://alchemy.run/providers/cloudflare/bucket/#with-auto-emptying

john-royal:
We currently use the S3-compatible API to empty an R2 bucket before deleting it. This PR switches to the REST API so you don't need to provide specialized credentials. Closes #723.

Also:
- Cleaned up some dead code
- We weren't returning the R2.dev subdomain if enabled, so I added that
- We had a function to set CORS rules but it wasn't being called anywhere - fixed that

### DIFF
--- a/alchemy-web/src/content/docs/providers/cloudflare/bucket.md
+++ b/alchemy-web/src/content/docs/providers/cloudflare/bucket.md
@@ -42,6 +42,30 @@ const publicBucket = await R2Bucket("public-assets", {
   name: "public-assets",
   allowPublicAccess: true,
 });
+console.log(publicBucket.domain); // [random-id].r2.dev
+```
+
+This enables the `r2.dev` domain for the bucket. This URL is rate-limited and not recommended for production use.
+
+## With CORS
+
+Create a bucket with CORS rules:
+
+```ts
+import { R2Bucket } from "alchemy/cloudflare";
+
+const corsBucket = await R2Bucket("cors-bucket", {
+  name: "cors-bucket",
+  cors: [
+    {
+      allowed: {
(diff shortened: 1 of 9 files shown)


alchemy-web/src/content/docs/providers/cloudflare/bucket.md
```diff
@@ -42,6 +42,30 @@ const publicBucket = await R2Bucket("public-assets", {
   name: "public-assets",
```
sam-goodwin, line 42: Did you update the docs on clean?

sam-goodwin: Ok we neve had any lol. Might be worth adding it.

Also change the title of in the frontmatter to `Bucket` instead of `R2Bucket`

PR #729, opened 2025-07-22
fetch all zones in findZoneForHostname

eastlondoner:
Otherwise we don't find our zone.

The PR branch fixed issue for us :+1: thanks StackBlitz


sam-goodwin:
Thanks for the fix. Didn't consider this one! Have you tested it and it works?

PR #737, opened 2025-07-23
feat(vercel): add Storage resource for managing Vercel storage

oieduardorabelo:
Thanks for the support via Discord!

Support for Vercel's `blob` storage only

Happy to contribute more if there is demand

final API:

```ts
const acme_project = await Project("acme_project", {
  name: "acme_project",
  framework: "nextjs",
  gitRepository: {
    type: "github",
    repo: "acme_project",
  },
  rootDirectory: "acme_project",
  serverlessFunctionRegion: "syd1",
  environmentVariables: [],
});

const acme_storage = await Storage("acme_storage", {
  name: "acme_storage",
  region: "syd1",
  team: "acme_team",
  type: "blob",
  projects: [
    {
      projectId: acme_project.id,
      envVarEnvironments: ["production"],
      envVarPrefix: "ACME",
    },
    {
      projectId: "prj_acme_second_project",
      envVarEnvironments: ["development"],
      // by default `envVarPrefix` is BLOB
    },
  ],
});
```


alchemy/src/vercel/storage.ts
```diff
@@ -0,0 +1,184 @@
...
+export interface Storage extends Resource<"vercel::Storage">, StorageProps {
+  /**
+   * The ID of the storage
+   */
+  id: string;
+
+  /**
+   * The storage store information
+   */
+  store: {
```
sam-goodwin: Good to pull this out into its own interface since it's a nested object

alchemy/src/vercel/storage.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  /**
+   * The region where the storage will be deployed
+   */
+  region: string;
+
+  /**
+   * The team ID that the storage belongs to
+   */
+  teamId: string;
```
sam-goodwin: Do you think we'll eventually have a `Team` resource? If yes, then this should be `team: string` so that it can be later `team: string | Team` which is our convention for supporting both identifiers and references to alchemy resources.

alchemy/src/vercel/storage.ts
```diff
@@ -0,0 +1,184 @@
...
+export interface StorageProps {
+  /**
+   * The desired name for the storage
+   */
+  name: string;
+
+  /**
+   * The region where the storage will be deployed
+   */
+  region: string;
```
sam-goodwin: Is there a union type of allowed regions we can add to make this easier?

I see: https://vercel.com/docs/edge-network/regions

```ts
type VercelRegion =
  | "cpt1"  // Cape Town, South Africa
  | "cle1"  // Cleveland, USA
  | "dxb1"  // Dubai, UAE
  | "dub1"  // Dublin, Ireland
  | "fra1"  // Frankfurt, Germany
  | "hkg1"  // Hong Kong
  | "lhr1"  // London, UK
  | "bom1"  // Mumbai, India
  | "kix1"  // Osaka, Japan
  | "cdg1"  // Paris, France
  | "pdx1"  // Portland, USA
  | "sfo1"  // San Francisco, USA
  | "gru1"  // São Paulo, Brazil
  | "icn1"  // Seoul, South Korea
  | "sin1"  // Singapore
  | "arn1"  // Stockholm, Sweden
  | "syd1"  // Sydney, Australia
  | "hnd1"  // Tokyo, Japan
  | "iad1"; // Washington, D.C., USA
  | (string & {}) // special rune to maintain auto-suggestions without closing the type to new regions or ones we missed
```

alchemy/src/vercel/storage.ts
```diff
@@ -0,0 +1,184 @@
...
+type StorageType = "blob";
+
+/**
+ * Properties for creating or updating a Storage
+ */
+export interface StorageProps {
+  /**
+   * The desired name for the storage
+   */
+  name: string;
```
sam-goodwin: Another convention we have is to leave this as optional `name?: string` and "@default id`

alchemy/src/vercel/storage.ts
sam-goodwin: Or should we just flatten these properties into the `Storage` interface?

alchemy/src/vercel/storage.ts
```diff
@@ -0,0 +1,184 @@
...
+  return response.json() as Promise<{ store: Storage["store"] }>;
+}
+
+/**
+ * Delete a storage instance
+ */
+async function deleteStorage(api: VercelApi, output: Storage) {
+  await api.delete(`/storage/stores/${output.id}/connections?teamId=${output.teamId}`);
+  await api.delete(`/storage/stores/${output.type}/${output.id}?teamId=${output.teamId}`);
+}
```
sam-goodwin: Should be checking status code and throwing errors

alchemy/src/vercel/storage.ts
oieduardorabelo: i ended up merging them in the `Storage` interface, however, it does feel "messy", because we are merging the API response to it and some properties can overlap... but looking at other resources, they are all doing that (aka merging back to the main interface), so i'm just following it too

alchemy/src/vercel/storage.ts
oieduardorabelo: there's a `Team` entity on Vercel, not sure if we'll ended up using (depends on which resources are migrated to alchemy)

i added a basic type for `VercelTeam` in `vercel.types.ts`

alchemy/src/vercel/storage.ts
oieduardorabelo: great point, i was looking from a local experiment only. fixed it.

alchemy/src/vercel/storage.ts
oieduardorabelo: i checked `upstash` and `vercel/project`.... what you suggested is not happening! so i just followed their examples...

anyway, i added the case as you suggested (fallback to `id`) in the `Storage` class

alchemy/src/vercel/storage.ts
oieduardorabelo: the `api.*` client is returning a custom error if `fetch` is not `response.ok`

i ended up adding `neverthrow` on these functions, but there's no custom error, just bubbling up the returned error

sam-goodwin:
This is looking great. Should add some tests in alchemy/test/vercel/storage.test.ts and follow the same practices as the `project.test.ts` test suite.

sam-goodwin:
Hi @oieduardorabelo I tried pushing some changes to your Pr bur I don't have permission to push to your fork. Usually I do, not sure if you disabled that permission.

I opened this PR with the changes: https://github.com/sam-goodwin/alchemy/pull/822. I'd prefer to continue with your PR, but LMK if you'd prefer I took over and got this merged.

oieduardorabelo:
@sam-goodwin man, i have no idea why you don't have permission 😂 

i'm back today + i'll try to fix this permission issue + rebase your branch into this pr

oieduardorabelo:
the checkbox is ticked to allow edits by maintainers

CleanShot 2025-09-05 at 20 08 25

Mkassabov:
cleaning up inactive prs

oieduardorabelo:
it was not inactive @Mkassabov , but i need a position / answer from folks maintaining the project to merge it

Mkassabov:
> it was not inactive @Mkassabov , but i need a position / answer from folks maintaining the project to merge it

@oieduardorabelo my apologies, saw it hadn't been touched in a few weeks. What can I help with to get this over the line? Anything specific or just a review?

oieduardorabelo:
no apologies needed, you’re all good, don’t stress! @Mkassabov 😄

if the resource is still relevant for the project, yes? ahaha (i just synced it with the main branch)

i merged Sam's changes + double checked his access to the branch, nothing out of the ordinary

Mkassabov:
Uhhh, looks fine to me at first glance. Not at my desk rn, but will try and run tests tonight and if they pass I'll merge it in.

Thanks for the contribution, and sorry again for closing this pre-emptively! <3

Mkassabov:
test seems to be failing due to a permission issue, might take me a bit longer to fight with vercel to get proper credentials.

no action on your end just keeping you posted

Mkassabov:
@oieduardorabelo I'm not able to get the tests on this to pass even with a new vercel account.
I made some changes to the test to make sure it uses the correct team id (will add this to our env variables so its easier to test vercel resources).

but then I still get
```
Error: Project not found
    at VercelApi.fetch (D:\code\OSS\alchemy\alchemy\alchemy\src\vercel\api.ts:99:11)
    at processTicksAndRejections (node:internal/process/task_queues:105:5)
    at connectProject (D:\code\OSS\alchemy\alchemy\alchemy\src\vercel\storage.utils.ts:75:20)
    at createProjectsConnection (D:\code\OSS\alchemy\alchemy\alchemy\src\vercel\storage.utils.ts:62:5)
    at create.<anonymous> (D:\code\OSS\alchemy\alchemy\alchemy\src\vercel\storage.ts:168:11)
    at Scope.<anonymous> (D:\code\OSS\alchemy\alchemy\alchemy\src\apply.ts:278:15)
    at _alchemy.run (D:\code\OSS\alchemy\alchemy\alchemy\src\alchemy.ts:465:12)
    at _apply (D:\code\OSS\alchemy\alchemy\alchemy\src\apply.ts:269:16)
    at D:\code\OSS\alchemy\alchemy\alchemy\test\vercel\storage.test.ts:35:17
    at Scope.<anonymous> (D:\code\OSS\alchemy\alchemy\alchemy\src\test\vitest.ts:201:13) {
```

Am I doing something wrong in running the tests?

Mkassabov:
@oieduardorabelo 
following up here, is this work still active? if so could you repair the tests

PR #741, opened 2025-07-25
feat(cloudflare): wasm support + prisma example

john-royal:
- [x] Working Prisma example
- [x] Support wasm in esbuild bundles
- [x] Support wasm in Miniflare
- [x] Guide
- [x] Run tests


alchemy/src/cloudflare/worker.ts
```diff
@@ -1381,137 +1357,27 @@ async function createMiniflare(props: {
...
+  const run = async () => {
+    for await (const bundle of props.bundle.watch(controller.signal)) {
+      const server = await miniflareServer.push({
+        ...sharedOptions,
+        bundle,
+      });
+      if (startPromise.status === "pending") {
+        startPromise.resolve(server.url);
+      }
+      logger.task("", {
```
sam-goodwin, line 1371: What should this empty string be

john-royal: Great question. The string seems to be ignored by the logger. Here's the snippet from `src/util/cli.ts`:

```ts
    task: (_id: string, data: Task) => {
      const prefix = data.prefix ? `[${data.prefix}]` : "";

      // Pad the prefix to ensure consistent alignment (12 characters total including brackets)
      const paddedPrefix = prefix.padEnd(12);
      const prefixWithColor =
        data.prefixColor && prefix
          ? colorize(paddedPrefix, data.prefixColor as ColorName)
          : paddedPrefix;
```

I would like to overhaul the logger since it could be way more useful here.

alchemy/test/cloudflare/bundle-handler.ts
```diff
@@ -1,6 +1,7 @@
 import { initLogger } from "braintrust";
 // biome-ignore lint/style/useNodejsImportProtocol: we are testing `crypto` and `node:crypto`
 import crypto from "crypto";
+import debug from "debug";
```
john-royal, line 4: This test was failing if we did `import { debug }` instead of `import debug`. Wrangler exhibited the same behavior.

The Grammy test passes — I believe that's a separate issue since in that case, there's a `require` call that the `@cloudflare/unenv/npm` patch takes care of.

alchemy-web/src/content/docs/guides/prisma-d1.mdx
```diff
@@ -0,0 +1,445 @@
+---
+title: Prisma D1
+description: Build a full-stack application with Prisma ORM and Cloudflare D1 Database using Alchemy for type-safe database operations.
+sidebar:
+  order: 16
+---
+
+import { Tabs, TabItem, Steps } from '@astrojs/starlight/components';
+
+Build a full-stack application with Prisma ORM and Cloudflare D1 Database. This guide shows you how to set up a type-safe database layer with automated migrations and a web interface using Prisma's powerful schema and client generation.
```
sam-goodwin: remove all buzzwords like "powerful"

john-royal: Good catch. Cringe Sonnet 4

PR #753, opened 2025-07-28
fix(cloudflare): add back esbuild alias plugin

john-royal:
I removed [Wrangler's plugin for handling user-defined aliases](https://github.com/cloudflare/workers-sdk/blob/75640c15bd2e242fde3e091972fa6a198bdaac69/packages/wrangler/src/deployment-bundle/bundle.ts#L323) a while back. That was probably a mistake. This adds it back and includes the comments from Wrangler on why it's there.

### DIFF
--- a/alchemy/src/cloudflare/bundle/plugin-alias.ts
+++ b/alchemy/src/cloudflare/bundle/plugin-alias.ts
@@ -1,0 +1,49 @@
+import type esbuild from "esbuild";
+
+// esbuild's `alias` option is applied after each plugin's onResolve hook,
+// whereas we would like these user-defined aliases to take precedence over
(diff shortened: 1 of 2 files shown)


sam-goodwin:
Are we guessing it was a mistake or do we have a failing test case to warrant it?

john-royal:
The only reason I removed it is because I didn't see the comments in `workers-sdk` as to why it was there in the first place:

```ts
// esbuild's `alias` option is applied after each plugin's onResolve hook,
// whereas we would like these user-defined aliases to take precedence over
// the unenv polyfill aliases, so we reimplement the aliasing as a plugin
// to be applied before that plugin (earlier in the array of plugins)
```

No test case (should I add one?) but not quite a guess either.

sam-goodwin:
That's good to know, let's do it.

PR #764, opened 2025-07-29
fix(cloudflare): avoid creating or deleting remote resources in local mode

### LINKED ISSUE #748: Worker URL is deleted when switching to `dev`
Image

john-royal:
Will close #748.

I used the "noop subresource" approach you suggested. Not sure if this is better.

### DIFF
--- a/alchemy/src/cloudflare/miniflare/build-worker-options.ts
+++ b/alchemy/src/cloudflare/miniflare/build-worker-options.ts
@@ -1,6 +1,7 @@
 import * as miniflare from "miniflare";
 import { assertNever } from "../../util/assert-never.ts";
 import type { HTTPServer } from "../../util/http.ts";
+import type { CloudflareApi } from "../api.ts";
 import {
   Self,
   type Binding,
@@ -14,6 +15,7 @@ import type { AssetsConfig } from "../worker.ts";
 import { createRemoteProxyWorker } from "./remote-binding-proxy.ts";
 
 export interface MiniflareWorkerInput {
+  api: CloudflareApi;
   id: string;
   name: string;
   compatibilityDate: string | undefined;
@@ -165,7 +167,8 @@ export const buildWorkerOptions = async (
         break;
       }
       case "hyperdrive": {
-        throw new Error("Hyperdrive bindings are not supported in local mode");
+        (options.hyperdrives ??= {})[key] = binding.dev.origin.unencrypted;
+        break;
       }
       case "images": {
         if (isRemoteBinding(binding)) {
@@ -329,6 +332,7 @@ export const buildWorkerOptions = async (
   }
   if (remoteBindings.length > 0) {
     const remoteProxy = await createRemoteProxyWorker({
+      api: input.api,
       name: input.name,
       bindings: remoteBindings,
     });
--- a/alchemy/src/cloudflare/worker.ts
+++ b/alchemy/src/cloudflare/worker.ts
@@ -789,6 +789,7 @@ const _Worker = Resource(
       throw new Error(options.bundle.error);
     }
     const bundle = options.bundle.value;
+    const api = await createCloudflareApi(props);
 
     if (this.scope.local && !props.dev?.remote) {
       let url: string | undefined;
@@ -800,6 +801,7 @@ const _Worker = Resource(
         );
         const controller = MiniflareController.singleton;
         url = await controller.add({
+          api,
           id,
           name: options.name,
           compatibilityDate: options.compatibilityDate,
@@ -839,8 +841,6 @@ const _Worker = Resource(
         Env: undefined!,
       } as unknown as Worker<B>);
     }
-
-    const api = await createCloudflareApi(props);
 
     if (this.phase === "create" || this.output.dev?.hasRemote === false) {
       if (props.version) {
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -19,11 +19,21 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Cache Astro resources
+        uses: actions/cache@v4
+        with:
+          path: |
+            alchemy-web/.astro
+          key: ${{ runner.os }}-astro-cache
+
       - name: Setup Alchemy Environment
         uses: ./.github/actions/setup-alchemy
         with:
           aws-role-arn: ${{ secrets.AWS_ROLE_ARN }}
           aws-region: us-west-2
+
+      - name: Install Playwright Browsers
+        run: bunx playwright install --with-deps chromium
 
       - name: Get Astro version
         id: astro-version
@@ -54,4 +64,5 @@ jobs:
           CLOUDFLARE_ACCOUNT_ID: ${{ env.CLOUDFLARE_ACCOUNT_ID }}
           CLOUDFLARE_API_KEY: ${{ env.CLOUDFLARE_API_KEY }}
           CLOUDFLARE_EMAIL: ${{ env.CLOUDFLARE_EMAIL }}
+          GENERATE_OG_IMAGES: true
           STAGE: prod
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -62,6 +62,7 @@ jobs:
           R2_SECRET_ACCESS_KEY: ${{ env.R2_SECRET_ACCESS_KEY }}
           SECRET_PASSPHRASE: ${{ env.SECRET_PASSPHRASE }}
           SENTRY_AUTH_TOKEN: ${{ env.SENTRY_AUTH_TOKEN }}
+          SENTRY_DSN: ${{ env.SENTRY_DSN }}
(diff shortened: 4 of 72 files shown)


john-royal:
@sam-goodwin Let me know if this approach is what you were thinking.

I think it may be cleaner to have a clearer separation between local and remote resources rather than trying to handle both concurrently, but maybe I'm just doing it wrong?

Also curious if you have any ideas for the todos in worker.ts.

alchemy/src/cloudflare/container.ts
```diff
@@ -368,6 +368,15 @@ export interface ContainerApplicationProps extends CloudflareApiOptions {
...
   adopt?: boolean;
+
+  /**
+   * If true, the container application will not be created, but will be retained if it already exists.
+   * This is used for local development.
+   *
+   * @default false
+   * @internal
+   */
+  noop?: boolean;
```
sam-goodwin: instead of `noop`, i was expecting `dev?: boolean` and just set it. Have the resource do nothing if `scope.dev || props.dev`

john-royal: Wouldn't it need to be `scope.dev && !props.dev.remote` if we were going that route?

My thinking was for this to be distinct from the usual `props.dev` since it's governed by the parent `Worker` resource — the change applies only to `ContainerApplication`, `CustomDomain`, `Route`, and `WorkerSubdomain` — the rest have the usual `props.dev`. Not sure if that makes sense though, so happy to change it.

sam-goodwin: They aren't exclusively sub resources. We just use them that way. 

`noop` is a strange property that only makes sense from the perspective of a sub resource. 

The outer resource should pass

```ts
dev: this.scope.dev || props.dev
```

That will ensure that it updates appropriately. And leaves the door open to us doing more with these resources in dev mode later.

john-royal: Got it. I do want to clarify one thing.

Context:
1. The way `Worker` is set up, we check if `this.scope.local && !props.dev.remote` — that's how we know to run in Miniflare or not
2. The `Worker` can pass `{ dev: this.scope.local && !props.dev.remote }` to the inner resource
3. But then if the inner resource becomes a no-op if `this.scope.local || props.dev`, then it won't update even if `props.dev.remote` is `true` on the parent `Worker`

Do you want us to change 1 or 2? Or would you rather change 3 to `this.scope.local && props.dev`? I don't have a strong opinion either way — let me know what you think.

sam-goodwin: This looks right:
```ts
{ dev: this.scope.local && !props.dev.remote }
```

alchemy/src/cloudflare/miniflare/build-worker-options.ts
```diff
@@ -143,7 +143,7 @@ export const buildWorkerOptions = async (
             raw: true,
           });
         } else {
-          (options.d1Databases ??= {})[key] = binding.id;
+          (options.d1Databases ??= {})[key] = binding.dev.id;
```
sam-goodwin, line 148: this seems weird, why did it change?

john-royal: If you create a binding locally, this ensures the local ID for it is stable (i.e. doesn't change after you deploy it).

I did the same for KV and some of the others for that reason.

alchemy/src/cloudflare/bucket.ts
```diff
@@ -202,88 +225,119 @@ export function isBucket(resource: Resource): resource is R2Bucket {
...
+  props: BucketProps = {},
+): Promise<R2Bucket> {
+  return await _R2Bucket(id, {
+    ...props,
+    dev: {
+      ...(props.dev ?? {}),
+      force: Scope.current.local,
+    },
+  });
+}
```
sam-goodwin, line 239: As we duplicate this, I wonder if we can solve it more generally.

e.g.
```ts
export interface ResourceProps {
  dev?: {
    force?: boolean;
  }
}
```

And then have `Resource` do this automatically.

Problem is I am not sure how to identify when NOT to do it. Unless we support something like:
```ts
export const MyResource = Resource("foo::bar", { local: true }, async function(..) { .. });
```

sam-goodwin: Let's not solve it in this PR, but think about it.

alchemy/src/cloudflare/bucket.ts
```diff
@@ -202,88 +225,119 @@ export function isBucket(resource: Resource): resource is R2Bucket {
...
-              err instanceof CloudflareApiError &&
-              err.status === 409 &&
-              props.adopt
-            ) {
-              return await getBucket(api, bucketName, props);
-            }
-            throw err;
-          },
+    if (this.scope.local && !props.dev?.remote) {
+      return this({
```
sam-goodwin, line 256: Yeah, this is why I don't liek switch and never use it. Not expressive enough and awkward syntax.

alchemy/src/cloudflare/pipeline.ts
```diff
@@ -337,6 +339,20 @@ export const Pipeline = Resource("cloudflare::Pipeline", async function <
   const api = await createCloudflareApi(props);
   const pipelineName = props.name ?? id;
 
+  if (this.scope.local && !props.dev?.remote) {
+    return this({
+      type: "pipeline",
+      id: this.output?.id ?? "",
```
sam-goodwin, line 345: use the resource id?

alchemy/src/cloudflare/pipeline.ts
```diff
@@ -337,6 +339,20 @@ export const Pipeline = Resource("cloudflare::Pipeline", async function <
...
+      name: this.output?.name ?? pipelineName,
+      endpoint: this.output?.endpoint ?? "",
+      version: this.output?.version ?? 0,
+      source: this.output?.source ?? [],
+      destination: props.destination,
+      compression: props.compression,
+      accountId: this.output?.accountId ?? "",
```
sam-goodwin, line 352: Maybe use `api.accountId`?

alchemy/src/cloudflare/pipeline.ts
```diff
@@ -337,6 +339,20 @@ export const Pipeline = Resource("cloudflare::Pipeline", async function <
   const api = await createCloudflareApi(props);
   const pipelineName = props.name ?? id;
 
+  if (this.scope.local && !props.dev?.remote) {
+    return this({
+      type: "pipeline",
+      id: this.output?.id ?? "",
+      name: this.output?.name ?? pipelineName,
+      endpoint: this.output?.endpoint ?? "",
```
sam-goodwin, line 347: What does this break? Do we not have a local endpoint for a pipeline?

alchemy/src/cloudflare/queue-consumer.ts
```diff
@@ -148,16 +157,28 @@ export const QueueConsumer = Resource(
...
   ): Promise<QueueConsumer> {
-    const api = await createCloudflareApi(props);
-
     // Get queueId from either props.queue or props.queueId
     const queueId =
       typeof props.queue === "string" ? props.queue : props.queue.id;
 
+    if (this.scope.local && props.dev) {
+      return this({
+        id: this.output?.id ?? "",
```
sam-goodwin, line 166: resource `id`?

alchemy/src/cloudflare/queue-consumer.ts
```diff
@@ -148,16 +157,28 @@ export const QueueConsumer = Resource(
...
+        queueId,
+        queue: props.queue,
+        type: "worker",
+        scriptName: props.scriptName,
+        settings: props.settings,
+        accountId: this.output?.accountId ?? "",
```
sam-goodwin, line 172: These empty strings make me nervous.

alchemy/src/cloudflare/queue.ts
```diff
@@ -208,26 +231,57 @@ export interface Queue<Body = unknown>
...
-  const api = await createCloudflareApi(props);
   const queueName = props.name ?? id;
+  const dev = {
+    id: this.output?.dev?.id ?? this.output?.id ?? id,
+    remote: props.dev?.remote ?? false,
+  };
+  if (this.scope.local && !props.dev?.remote) {
+    return this({
+      type: "queue",
+      id: this.output?.id ?? "",
```
sam-goodwin, line 260: `?? id`?

sam-goodwin (approved):

alchemy/src/cloudflare/worker.ts
```diff
@@ -522,6 +524,11 @@ export type Worker<
     placement?: {
       mode: "smart";
     };
+
+    /**
+     * Whether the worker has a remote deployment
```
sam-goodwin, line 529: mark as @internal

sam-goodwin: Should it be in `dev?: { hasRemote: boolean }`

PR #770, opened 2025-07-29
fix: handle dangling processes

### LINKED ISSUE #747: Dangling processes
```
Port 5189 is in use, trying another one...
Port 5190 is in use, trying another one...
Port 5191 is in use, trying another one...
Port 5192 is in use, trying another one...
Port 5193 is in use, trying another one...
Port 5194 is in use, trying another one...
Port 5195 is in use, trying another one...
Port 5196 is in use, trying another one...
```

john-royal:
Closes #747:
- Added `this.spawn()` to `Scope` and `Context` to register a process with a teardown hook
- Refactored Miniflare controller
- Use `signal-exit` to add exit listener (same as Miniflare)

We don't intercept the Miniflare exit anymore since that was causing issues. However, this does mean when you exit with ctrl + C, you get a 130 exit code instead of 0.

I did try implementing the IPC solution ([branch](https://github.com/sam-goodwin/alchemy/tree/john/747-dangling-process-ipc)). It works on its own, but not with our CLI, so I didn't include it in this PR.


alchemy/src/scope.ts
```diff
@@ -466,6 +477,8 @@ export class Scope {
     });
 
     if (!this.parent && process.env.ALCHEMY_TEST_KILL_ON_FINALIZE) {
+      this.logger.log("[debug] Killing on finalize");
+      await Promise.allSettled(this.cleanups.map((cleanup) => cleanup()));
```
sam-goodwin: Should we provide a function that does this since it's done in multiple places?

`cleanup()`

alchemy/src/scope.ts
```diff
@@ -537,6 +550,49 @@ export class Scope {
...
+      return true;
+    });
+  }
+
+  /**
+   * Register a cleanup function that will be called when the scope is finalized.
+   * The provided function should create resources and return an object with a cleanup function.
+   * This should only be called on the root scope to ensure proper cleanup.
+   */
+  public async spawn<T extends { cleanup: () => Promise<void> }>(
```
sam-goodwin: That's not what this function does

alchemy/src/scope.ts
```diff
@@ -466,6 +477,8 @@ export class Scope {
     });
 
     if (!this.parent && process.env.ALCHEMY_TEST_KILL_ON_FINALIZE) {
```
sam-goodwin, line 472: This only runs during our tests? Of are you also catching a signal and running it?

alchemy/src/scope.ts
```diff
@@ -178,6 +183,12 @@ export class Scope {
     }
     this.phase = phase;
 
+    // Install exit handlers on first root scope creation
+    if (!this.parent && !Scope.exitHandlerInstalled) {
+      Scope.exitHandlerInstalled = true;
+      this.installExitHandlers();
+    }
```
sam-goodwin: Maybe do it in alchemy where the root is created instead of in he scope?

alchemy/src/scope.ts
john-royal: Yeah this only runs during tests. For some reason tests would fail without this

alchemy/src/scope.ts
john-royal: Agree. Any idea for what we should call it instead?

We need a function that can return a cleanup for use with Miniflare and the worker tail.

alchemy/src/scope.ts
john-royal: Sure

alchemy/src/scope.ts
sam-goodwin: It seems overly complicated. Why not just onCleanup?

It's also not clear why the user provides a function that we immediately call

john-royal: Fair enough

alchemy/src/cloudflare/miniflare/build-worker-options.ts
sam-goodwin, line 1: what changed in this file? Moving files around can make it harder to review code

alchemy/src/cloudflare/miniflare/controller.ts
sam-goodwin, line 1: we should make up our mind on naming conventions for the file, should this be miniflare-controller.ts? 

i see miniflare-worker-proxy.ts, but `miniflare-worker-options.ts` was just renamed to `build-worker-options.ts`

alchemy/src/scope.ts
```diff
@@ -537,6 +550,49 @@ export class Scope {
...
+   * Install global process exit handlers to ensure cleanup
+   */
+  private installExitHandlers() {
+    if (this.parent) return;
+    onExit((code) => {
+      if (this.cleanups.length === 0) return;
+      this.logger.log(kleur.gray("Exiting..."));
+      Promise.allSettled(this.cleanups.map((cleanup) => cleanup())).then(() => {
+        console.log("Exiting with code:", code); // this never gets logged, thanks Miniflare
+        process.exit(code);
```
sam-goodwin: why do we need to call this?

alchemy/src/cloudflare/miniflare/controller.ts
john-royal: I figured it doesn't need the prefix if it's already in the `miniflare` directory, is that ok with you or would you prefer to keep the prefix?

alchemy/src/scope.ts
john-royal: If you return `true` (as we do on line 565), that tells the `onExit` function that we'll call process.exit, so don't exit for us. This is important because otherwise, it'll exit before those cleanup promises can be awaited (`onExit` doesn't let you return a promise).

The debug log I added there isn't 100% true - it doesn't get logged if Miniflare is there intercepting the exit signal (which it shouldn't, I can make an issue for that tomorrow), but if we don't have Miniflare messing up our process (e.g. if your app is just a Vite website or something), then we do want to handle the exit ourselves.

alchemy/src/cloudflare/miniflare/build-worker-options.ts
john-royal: Before, the worker options were being assembled in multiple places:
1. The `Worker` resource would handle watching and piping the bundle output to Miniflare
2. There were multiple layers of back and forth between `miniflare-worker-options.ts`, `remote-binding-proxy.ts`, and `MiniflareServer` (`miniflare.ts`) to assemble bindings, provision the remote proxy if necessary, and finally push that updated code into Miniflare
3. This would be repeated every time the bundle updates, even when everything except the modules is the same

Now the options-wrangling is in one place, `buildWorkerOptions`, so it's easier to reason about. The function takes your bundle source and other config, the bindings are assembled once (instead of with every bundle update), and it returns an async iterator that we use to update the Miniflare instance when there's a change.

Figured this would be good to reduce indirection.

sam-goodwin: Makes sense.

alchemy/src/cloudflare/miniflare/controller.ts
sam-goodwin: I typically try to match the exported name and the file. 

```ts
export const MiniflareController
```

This would be `miniflare-controller.ts`

john-royal:
Just updated it to address your feedback and tests are all passing. Should be good to merge unless there's any other changes you want!

PR #779, opened 2025-07-31
Switch from xdg-app-paths to env-paths

### LINKED ISSUE #778: Alchemy can't be ran with deno
When trying to run alchemy with deno users will see the following error:

```bash
deno run alchemy.run.ts
error: [ERR_MODULE_NOT_FOUND] Cannot find module 'file:///Users/just-be/Library/Caches/deno/npm/registry.npmjs.org/xdg-app-paths/8.3.0/src/mod.deno.ts' imported from 'file:///Users/just-be/Library/Caches/deno/npm/registry.npmjs.org/alchemy/0.55.3/lib/cloudflare/oauth.js'
```

This is actually an issue with an upstream dependency which I've created an issue for here: https://github.com/rivy/js.xdg-app-paths/issues/1

There was a conversation back in 2021 that the author of the package was in around how to package for deno: https://github.com/denoland/deno/issues/3196#issuecomment-782789368. While well-intentioned, Deno just consumes node-like packages and it would've been better had they not tried to do something special for deno. 

I'm going to follow up with a PR to swap over to `env-paths`.

just-be-dev:
Fixes #778 

`xdg-app-paths` tries to do special handling for deno projects, but it did so before Deno 2.0 release. That special handling is broken now and causes alchemy to fail for anyone trying to run it in Deno. I've instead switched over to `env-paths` which hasn't been updated as recently but has significantly more downloads. It's a sindresorhus package.

### DIFF
--- a/alchemy/src/cloudflare/oauth.ts
+++ b/alchemy/src/cloudflare/oauth.ts
@@ -70,7 +70,7 @@ export const getRefreshedWranglerConfig = singleFlight(() =>
 );
 
 const getWranglerConfigPath = memoize(async () => {
-  const xdgConfigDir = envPaths("wrangler", { suffix: "" }).config;
+  const xdgConfigDir = envPaths(".wrangler", { suffix: "" }).config;
   const legacyConfigDir = path.join(os.homedir(), ".wrangler");
   const configDir = (await fs
     .stat(legacyConfigDir)


alchemy/src/cloudflare/oauth.ts
```diff
@@ -70,7 +70,7 @@ export const getRefreshedWranglerConfig = singleFlight(() =>
 );
 
 const getWranglerConfigPath = memoize(async () => {
-  const xdgConfigDir = xdgAppPaths(".wrangler").config();
+  const xdgConfigDir = envPaths("wrangler", { suffix: "" }).config;
```
just-be-dev: The `suffix` option is important here because otherwise `envPaths` will append a `-nodejs` suffix onto the config path. The original intent of this package is for storing new things, not necessarily looking up existing. Still, we can get the same behavior by passing in an empty string.

sam-goodwin:
@john-royal  any thoughts on this one?

john-royal:
Honestly, I've been wanting to switch to this because it's better architected (like a good sindresorhus package) and more popular than `xdg-app-paths`.

We should be aware that this may break compatibility with `wrangler login` in some edge cases.

For context, XDG paths are really only a Linux thing, so for other platforms, you use an analog like `AppData/Roaming` on Windows or `Library/Preferences` on Mac. These directories are fairly standardized, but there are rare edge cases where one library might have a different opinion from another. The reason we use `xdg-app-paths` right now isn't because it's the best option, but because we had a user hit an edge case a few weeks ago, so I figured it's best to keep our behavior fully compatible.

However, given that we have our own `alchemy login` command and we're moving away from relying on Wrangler auth anyway, I say merge it. Most users will be unaffected; the few that are can simply run `alchemy login` and they'll be taken care of.

alchemy/src/cloudflare/oauth.ts
just-be-dev:
```suggestion
  const xdgConfigDir = envPaths(".wrangler", { suffix: "" }).config;
```

PR #781, opened 2025-08-01
chore: more posthog logs

Mkassabov:
## What this does
- adds `cli.<events>` to cli
- removes dead legacy telemetry code
- ensure project id exists by setting the project id to `temp_${userId}` if it doesn't exist
- disable telemetry on our smoke tests
- adds project grouping


alchemy/bin/trpc.ts
```diff
@@ -1,3 +1,54 @@
...
+    telemetry.record({
+      event: "cli.success",
+      command: path,
+    });
+    return result;
+  } catch (error) {
+    telemetry.record({
+      event: "cli.error",
+      command: path,
+    });
```
sam-goodwin: Doesn't this mistakenly record an error when we exit with `0`?

alchemy/src/util/telemetry/client.ts
```diff
@@ -85,7 +85,28 @@ export class TelemetryClient implements ITelemetryClient {
...
+        properties: {
+          $group_type: "project",
+          $group_key: safeProjectId,
+        },
+      }),
+    });
+    if (!groupResponse.ok) {
+      throw new Error(
+        `Failed to send group identify: ${groupResponse.status} ${groupResponse.statusText} - ${await groupResponse.text()}`,
+      );
```
sam-goodwin, line 108: Will this crash alchemy?

alchemy/src/util/telemetry/context.ts
```diff
@@ -132,7 +132,7 @@ function environment(): Telemetry.Context["environment"] {
 
 export async function context(input: {
   sessionId: string;
-  phase: Phase;
+  phase?: Phase;
```
sam-goodwin, line 135: Why did this become optional?

alchemy/src/util/telemetry/client.ts
```diff
@@ -85,7 +85,28 @@ export class TelemetryClient implements ITelemetryClient {
     if (events.length === 0) {
       return;
     }
-    const { userId, ...context } = await this.context;
+    const { userId, projectId, ...context } = await this.context;
+    const safeProjectId = projectId ?? `temp_${userId}`;
```
sam-goodwin, line 89: will this double the number of projects if the user later runs `git init`?

Mkassabov: not double, just +1 because all the non-git projects would have the same id since `temp_${userId}` is consistent.

no matter what we would have ids changed between non-git and git when somebody runs `git init`, so most other solutions would be *2.

The other option to make it more stable is to have `alchemy init` run `git init` to reduce how often this happens

alchemy/src/util/telemetry/context.ts
Mkassabov: so we can use the same telemetry client in the cli, and the cli doesn't have a phase.

alchemy/bin/trpc.ts
Mkassabov: fixed

alchemy/src/util/telemetry/client.ts
Mkassabov: no. this is a private method thats always called with a catch in finalize
```
  async finalize() {
    await this.send(this.events).catch((error) => {
      if (!this.options.quiet) {
        console.warn(error);
      }
    });
  }
```

its the same way we handle normal send events failing

PR #782, opened 2025-08-01
chore: set up privacy policy for OAuth Client ID registration

sam-goodwin:


alchemy-web/src/pages/privacy.astro
```diff
@@ -0,0 +1,113 @@
...
+    margin-top: 0;
+  }
+  ul {
+    padding-left: 1.25rem;
+  }
+</style>
+
+<main>
+  <h1>Privacy&nbsp;Policy</h1>
+  <p>Last updated: <!-- Will be substituted at build time -->{new Date().toLocaleDateString()}</p>
```
Mkassabov: This makes me a little nervous because it updates every build not on changes so users might think something changed when it didn't. I try and read privacy policies for SaaS stuff by looking at the data and skimming for what I need (e.g is my user info being sold)

I think it's better to have this be a static value that we update manually

alchemy-web/src/pages/privacy.astro
```diff
@@ -0,0 +1,113 @@
...
+    runtime libraries, and associated websites (collectively, "Alchemy") collect and process information
+    when you install, run, or otherwise use Alchemy.
+  </p>
+
+  <h2>2.&nbsp;Information We Collect</h2>
+  <p>Alchemy is designed to collect <em>anonymous</em> usage analytics that help us understand how our tools are used and improve their reliability and performance. We do <strong>not</strong> collect names, email addresses, source code, or any directly identifying personal information. Specifically, the telemetry system built into the Alchemy CLI gathers the following data:</p>
+
+  <h3>2.1.&nbsp;Identifiers</h3>
+  <ul>
+    <li><strong>User&nbsp;ID&nbsp;(UUID)</strong>&nbsp;- generated once per machine and stored locally under your XDG&nbsp;config directory (e.g. <code>~/Library/Preferences/alchemy</code>). <em>Not collected when running in Continuous Integration (CI) environments.</em></li>
```
Mkassabov: Windows? Or are we fine with people assuming "some windows directory" because the actual directory here doesn't matter?

alchemy-web/src/pages/privacy.astro
sam-goodwin: Good point. will change

alchemy-web/src/pages/privacy.astro
sam-goodwin: No, it should cover windows. What should it say?

Mkassabov: for windows we store in `%LOCALAPPDATA%` defaults to `AppData/Local` so the final result it something like `C:\Users\michael\AppData\Local\alchemy`

PR #801, opened 2025-08-06
feat(cloudflare): allow local connection string for hyperdrive

john-royal:
Example usage:

```ts
const db = await Hyperdrive("my-postgres-db", {
  name: "my-postgres-db",
  origin: "postgresql://user:password@ep-example-host-1234.us-east-1.aws.neon.tech/mydb?sslmode=require",
  dev: {
    origin: "postgres://user:password@localhost:5432/postgres",
  },
});
```

### DIFF
--- a/alchemy/src/cloudflare/hyperdrive.ts
+++ b/alchemy/src/cloudflare/hyperdrive.ts
@@ -1,7 +1,8 @@
+import { alchemy } from "../alchemy.ts";
 import type { Context } from "../context.ts";
 import { Resource } from "../resource.ts";
 import { Scope } from "../scope.ts";
-import type { Secret } from "../secret.ts";
+import { Secret } from "../secret.ts";
 import { logger } from "../util/logger.ts";
 import { handleApiError } from "./api-error.ts";
 import { createCloudflareApi, type CloudflareApiOptions } from "./api.ts";
@@ -9,7 +10,7 @@ import { createCloudflareApi, type CloudflareApiOptions } from "./api.ts";
 /**
  * Origin configuration for a PostgreSQL or MySQL database connection
  */
-export interface HyperdriveOrigin {
+export interface HyperdrivePublicOrigin {
   /**
    * Database name
    */
@@ -24,7 +25,7 @@ export interface HyperdriveOrigin {
    * Database password
    * Use alchemy.secret() to securely store this value
    */
-  password: Secret;
+  password: string | Secret;
 
   /**
    * Database port
@@ -57,7 +58,7 @@ export interface HyperdriveOriginWithAccess {
    * Access client secret
    * Use alchemy.secret() to securely store this value
    */
-  access_client_secret: Secret;
+  access_client_secret: string | Secret;
 
   /**
    * Database host
@@ -119,6 +120,16 @@ export interface HyperdriveMtls {
   sslmode?: "verify-ca" | "verify-full";
 }
 
+export type HyperdriveOriginInput =
+  | string
+  | Secret
+  | HyperdrivePublicOrigin
+  | HyperdriveOriginWithAccess;
+
+export type HyperdriveOrigin =
+  | HyperdrivePublicOrigin
+  | HyperdriveOriginWithAccess;
+
 /**
  * Properties for creating or updating a Cloudflare Hyperdrive.
  */
@@ -131,7 +142,7 @@ export interface HyperdriveProps extends CloudflareApiOptions {
   /**
    * Database connection origin configuration
    */
-  origin: HyperdriveOrigin | HyperdriveOriginWithAccess;
+  origin: HyperdriveOriginInput;
 
   /**
    * Caching configuration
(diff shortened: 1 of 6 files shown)


tobimori:
I wonder if it would make sense to still create Hyperdrive when not specifying a localConnectionString? i.e. no localConnectionString = force remote binding

john-royal:
It would make sense, except remote bindings for Hyperdrive aren't supported by Cloudflare yet (see [unsupported remote bindings](https://developers.cloudflare.com/workers/development-testing/#unsupported-remote-bindings)). So, even if we create the remote Hyperdrive config, we wouldn't be able to use it.

They say this is being worked on though so would love to revisit when this is supported.

sam-goodwin:
Would have expected
```ts
const db = await Hyperdrive("my-postgres-db", {
  name: "my-postgres-db",
  origin: {
    database: "postgres",
    host: "database.example.com",
    password: alchemy.secret("your-password"),
    port: 5432,
    user: "postgres",
  },
  dev: {
    localConnectionString: alchemy.secret(
      "postgres://user:password@localhost:5432/postgres",
    ),
  }
});
```

But I also find it ugly that def is a connection string whereas prod is an `origin`. Maybe it could be done with overrides instead?
```ts
const db = await Hyperdrive("my-postgres-db", {
  name: "my-postgres-db",
  origin: {
    database: "postgres",
    host: "database.example.com",
    password: alchemy.secret("your-password"),
    port: 5432,
    user: "postgres",
  },
  dev: {
    host: "localhost",
    password: alchemy.secret("password")
  }
});
```

Or change everything to connection strings:
```ts
const db = await Hyperdrive("my-postgres-db", {
  name: "my-postgres-db",
  origin: alchemy.secret("postgres://user:password@database.example.com:5432/postgres"),
  dev: {
    origin: alchemy.secret("postgres://user:password@localhost:5432/postgres"),
  }
});
```

sam-goodwin:
We could also wrap the `_Hyperdrive` resource and do the secret wrapping automatically
```ts
const db = await Hyperdrive("my-postgres-db", {
  name: "my-postgres-db",
  origin: "postgres://user:password@database.example.com:5432/postgres",
  dev: {
    origin: "postgres://user:password@localhost:5432/postgres",
  }
});
```

alchemy/src/cloudflare/miniflare/build-worker-options.ts
```diff
@@ -165,7 +165,8 @@ export const buildWorkerOptions = async (
         break;
       }
       case "hyperdrive": {
-        throw new Error("Hyperdrive bindings are not supported in local mode");
+        (options.hyperdrives ??= {})[key] = binding.dev.origin.unencrypted;
```
sam-goodwin, line 168: 🔥

alchemy/src/cloudflare/hyperdrive.ts
```diff
@@ -170,7 +186,11 @@ export interface Hyperdrive
   /**
    * Database connection origin configuration
    */
-  origin: HyperdriveOrigin | HyperdriveOriginWithAccess;
+  origin: HyperdrivePublicOrigin | HyperdriveOriginWithAccess;
+
+  dev: {
+    origin: Secret;
+  };
```
sam-goodwin: document

alchemy/src/cloudflare/hyperdrive.ts
```diff
@@ -257,19 +277,63 @@ export interface Hyperdrive
  *   }
  * });
  */
-export const Hyperdrive = Resource(
+
+export async function Hyperdrive(
```
sam-goodwin: nit: no new line between comment and function

alchemy/src/cloudflare/hyperdrive.ts
```diff
@@ -375,16 +440,91 @@ function prepareRequestBody(props: HyperdriveProps): any {
...
+};
+
+const toConnectionString = (
+  origin: HyperdrivePublicOrigin | HyperdriveOriginWithAccess,
+) => {
+  const password = Secret.unwrap(
+    "password" in origin ? origin.password : origin.access_client_secret,
+  );
+  return new Secret(
+    `${origin.scheme || "postgres"}://${origin.user}:${password}@${origin.host}:${origin.port || 5432}/${origin.database}`,
```
sam-goodwin: Should the default port be determined based on schema? i think mysql and postgres have different default ports?

tobimori: mysql is 3306

john-royal: This is done by the normalizePort function:

```ts
const normalizePort = (
  scheme: "postgres" | "mysql",
  port: string | number | undefined,
) => {
  if (typeof port === "number") {
    return port;
  }
  if (port) {
    return Number.parseInt(port);
  }
  return scheme === "postgres" ? 5432 : 3306;
};
```

By the time it's passed into `toConnectionString`, the port has already been parsed. I'll remove the 5432 fallback to make this clearer.

sam-goodwin (approved):

PR #804, opened 2025-08-06
feat(cloudflare): clear miniflare data on destroy

### LINKED ISSUE #746: Clear miniflare stats on destroy

john-royal:
Closes #746. Depends on #764, which should be merged first.

Deletes data for the following binding types on destroy:
- D1
- Durable Object
- KV
- R2
- Workflow

Queues don't seem to be persisted locally anymore, so no need to delete their data.

Note: I didn't include scoping the Miniflare data by app/stage in this PR. There are some complications there since we'd need to pass the app/stage into the Vite plugin, so figured that can wait for another issue if we think it's necessary.


sam-goodwin:
> Note: I didn't include scoping the Miniflare data by app/stage in this PR. There are some complications there since we'd need to pass the app/stage into the Vite plugin, so figured that can wait for another issue if we think it's necessary.

That's fine. Do we have that issue?

sam-goodwin (approved):

alchemy/src/cloudflare/bucket.ts
```diff
@@ -271,6 +272,9 @@ const _R2Bucket = Resource(
 
     if (this.phase === "delete") {
       if (props.delete !== false) {
+        if (this.output.dev.id) {
```
sam-goodwin: dev?.id because old versions might be null

sam-goodwin (approved):

Issue #811, opened 2025-08-07
Remote Proxy Worker should use credentials passed in to the Worker


sam-goodwin:
```ts
export async function createRemoteProxyWorker(input: {
  name: string;
  bindings: WorkerBindingSpec[];
}): Promise<RemoteBindingProxy> {
  const api = await createCloudflareApi(); // <- missing options
```

This will break;

```ts
await Worker("..", {
  apiToken: alchemy.secret(..)
})
```

We should also check the codebase for other occurences of this. Ideally we'd pass the CloudflareApi client in as the first argument to avoid this problem.

I guess one day it'll be an Effect Service instead!

PR #812, opened 2025-08-07
fix(cloudflare): use explicit Worker credentials to create local proxy worker

sam-goodwin:
Closes https://github.com/sam-goodwin/alchemy/issues/811

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -62,6 +62,7 @@ jobs:
           R2_SECRET_ACCESS_KEY: ${{ env.R2_SECRET_ACCESS_KEY }}
           SECRET_PASSPHRASE: ${{ env.SECRET_PASSPHRASE }}
           SENTRY_AUTH_TOKEN: ${{ env.SENTRY_AUTH_TOKEN }}
+          SENTRY_DSN: ${{ env.SENTRY_DSN }}
           SENTRY_ORG: ${{ env.SENTRY_ORG }}
           STRIPE_API_KEY: ${{ env.STRIPE_API_KEY }}
           UPSTASH_API_KEY: ${{ env.UPSTASH_API_KEY }}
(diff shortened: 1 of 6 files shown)


alchemy/src/cloudflare/miniflare/remote-binding-proxy.ts
```diff
@@ -50,7 +51,7 @@ export async function createRemoteProxyWorker(input: {
         minimal_mode: true,
       },
     }),
-    import("../worker-subdomain.ts").then((m) => m.getAccountSubdomain(api)),
+    getAccountSubdomain(input.api),
```
john-royal, line 54: I remember doing a dynamic import here because of some circular dependency resolution error but if that's no longer an issue, might as well get rid of it.

alchemy/src/cloudflare/worker.ts
```diff
@@ -740,6 +740,8 @@ const _Worker = Resource(
     const local = this.scope.local && !props.dev?.remote;
     const watch = this.scope.watch;
 
+    const api = await createCloudflareApi(props);
```
john-royal, line 743: The reason I had this moved down originally was to improve startup times in cases where you don't need any remote bindings — perhaps we could pass through the API props instead so we still have the benefits of lazy initialization?

alchemy/src/cloudflare/miniflare/remote-binding-proxy.ts
sam-goodwin: Smoke tests pass and they all use dev. I added some explicit remote KV namespaces and it still works. THink we're good

alchemy/src/cloudflare/worker.ts
sam-goodwin: Fixing with https://github.com/sam-goodwin/alchemy/pull/821

PR #820, opened 2025-08-07
Dev tooling cdp

Mkassabov:


alchemy/src/alchemy.ts
```diff
@@ -189,6 +192,20 @@ async function _alchemy(
     try {
       Scope.storage.enterWith(root);
       Scope.storage.enterWith(stage);
+
+      // Add delay using stage.run() to preserve context
+      await stage.run(async () => {
```
sam-goodwin, line 197: Why does this need to be in scope

sam-goodwin (commented):
Looks like alchemy.js got checked in by mistake.

alchemy/bin/alchemy.js
john-royal, line 1: Looks like this was committed by mistake?

alchemy/src/tooling/cdp-server-remote.ts
```diff
@@ -0,0 +1,176 @@
+import { spawn, type ChildProcess } from "node:child_process";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+import { logger } from "../util/logger.ts";
+import { detectRuntime } from "../util/detect-node-runtime.ts";
+import { detectPackageManager } from "../util/detect-package-manager.ts";
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
```
sam-goodwin: can we use `import.meta.dir`

PR #829, opened 2025-08-09
feat(cloudflare): live proxy Worker

sam-goodwin:
This adds a "live proxy worker" to our dev mode to facilitate:
1. proxy push event hooks from a remote worker to a local worker (`queue`, `scheduled`, `email`, `tail`, `tailStream`, `test`, `trace`)
2. proxy HTTP requests received over the public internet (via a workers.dev subdomain) to a local worker

### DIFF
--- a/alchemy/src/cloudflare/miniflare/remote-binding-proxy.ts
+++ b/alchemy/src/cloudflare/miniflare/remote-binding-proxy.ts
@@ -41,6 +41,7 @@ export async function createRemoteProxyWorker(input: {
       name: input.name,
       metadata: {
         main_module: script.bundle.entrypoint,
+        // TODO(sam/john): should this be set to latest? Or match the user's worker that we are proxying?
         compatibility_date: "2025-06-16",
         bindings: input.bindings,
         observability: { enabled: false },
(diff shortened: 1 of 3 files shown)


alchemy/src/cloudflare/miniflare/remote-binding-proxy.ts
```diff
@@ -41,6 +41,7 @@ export async function createRemoteProxyWorker(input: {
       name: input.name,
       metadata: {
         main_module: script.bundle.entrypoint,
+        // TODO(sam/john): should this be set to latest? Or match the user's worker that we are proxying?
```
john-royal, line 44: The remote binding proxy doesn't contain any user code so not sure if it matters. My inclination would be to either match Wrangler's date (2025-04-28, per their [config](https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/templates/remoteBindings/proxyServerWorker/wrangler.jsonc)) or to keep it as is. AFAIK this setup isn't part of their public API, so my guess is if they made a breaking change, it would probably be associated with a compatibility date bump — keeping it static helps insulate us from there.

Not sure if I linked to the [original implementation](https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/api/remoteBindings/index.ts) in our code for reference, but definitely will if I haven't already.

PR #830, opened 2025-08-10
fix(core): instrumented state store should not swallot store errors

sam-goodwin:


alchemy/src/state/instrumented-state-store.ts
```diff
@@ -24,16 +24,19 @@ export class InstrumentedStateStore<T extends StateStore>
...
+      return await fn();
+    } catch (err) {
+      error = err;
+      throw err;
+    } finally {
+      this.telemetryClient.record({
+        event,
+        stateStoreClass: this.stateStoreClass,
+        elapsed: performance.now() - start,
+        error: error instanceof Error ? error : new Error(String(error)),
```
sam-goodwin: this looks weird, shouldn't it be `undefined` if no error?

PR #833, opened 2025-08-11
fix(cloudflare): dev mode vite process reload fix

Rhayxz:
What was causing it:

The dev runner tracked only a single PID in .alchemy/{id}.pid and tried to SIGTERM that one process on reload.
During rapid saves, a new dev process was spawned before the previous Vite process and its child tree had fully exited, leaving the old session alive.
Because the lingering Vite kept its port, subsequent spawns would pick new ports, resulting in accumulating dev sessions and port churn.

Current fix:

Implemented a per-run PID registry and process-group teardown in runDevCommand().
PID registry: .alchemy/pids/{id}-{pid}.pid (one file per dev session).
On each reload, sweep all pid files for the id, SIGTERM each, then delete their pid files. Also removes the legacy single .alchemy/{id}.pid for backward compatibility.
Spawn dev processes as detached process groups and prefer killing the whole group via process.kill(-pid) with a fallback to direct pid signals and SIGKILL if needed.
Left Vite/TanStack integrations untouched; they continue to call Website dev as before: Vite() and TanStackStart().

### DIFF
--- a/alchemy/src/cloudflare/website.ts
+++ b/alchemy/src/cloudflare/website.ts
@@ -353,21 +353,71 @@ async function runDevCommand(
     cwd?: string;
   },
 ) {
-  const persistFile = path.join(process.cwd(), ".alchemy", `${props.id}.pid`);
-  if (await exists(persistFile)) {
-    const pid = Number.parseInt(await fs.readFile(persistFile, "utf8"));
-    try {
-      // Actually kill the process if it's alive
-      process.kill(pid, "SIGTERM");
-    } catch {
-      // ignore
+  // Directory storing per-process PID files: .alchemy/pids/{id}-{pid}.pid
+  const pidDir = path.join(process.cwd(), ".alchemy", "pids");
+  const pidPrefix = `${props.id}-`;
+  await fs.mkdir(pidDir, { recursive: true });
+
+  // Sweep and terminate any previously tracked dev sessions for this id
+  try {
+    const entries = await fs.readdir(pidDir);
+    for (const entry of entries) {
+      if (entry.startsWith(pidPrefix) && entry.endsWith(".pid")) {
+        const full = path.join(pidDir, entry);
+        try {
+          const text = await fs.readFile(full, "utf8");
+          const oldPid = Number.parseInt(text);
+          if (!Number.isNaN(oldPid)) {
+            // Try to kill the whole process group first, then the PID
+            try {
+              process.kill(-oldPid, "SIGTERM");
+            } catch {
+              // ignore
+            }
+            try {
+              process.kill(oldPid, "SIGTERM");
+            } catch {
+              // ignore
+            }
+          }
+        } catch {
+          // ignore read issues
+        }
+        try {
+          await fs.unlink(full);
+        } catch {
+          // ignore unlink issues
+        }
+      }
+    }
+  } catch {
+    // ignore sweep issues
+  }
+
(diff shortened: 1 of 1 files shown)


alchemy/src/cloudflare/website.ts
```diff
@@ -353,21 +353,71 @@ async function runDevCommand(
...
 ) {
-  const persistFile = path.join(process.cwd(), ".alchemy", `${props.id}.pid`);
-  if (await exists(persistFile)) {
-    const pid = Number.parseInt(await fs.readFile(persistFile, "utf8"));
-    try {
-      // Actually kill the process if it's alive
-      process.kill(pid, "SIGTERM");
-    } catch {
-      // ignore
+  // Directory storing per-process PID files: .alchemy/pids/{id}-{pid}.pid
```
sam-goodwin, line 356: why do we need `-{pid}` prefix?

alchemy/src/cloudflare/website.ts
```diff
@@ -353,21 +353,71 @@ async function runDevCommand(
...
+  const pidDir = path.join(process.cwd(), ".alchemy", "pids");
+  const pidPrefix = `${props.id}-`;
+  await fs.mkdir(pidDir, { recursive: true });
+
+  // Sweep and terminate any previously tracked dev sessions for this id
+  try {
+    const entries = await fs.readdir(pidDir);
+    for (const entry of entries) {
```
sam-goodwin, line 364: Can use `await Promise.all` here instead of a `for` loop to avoid sequential IO

alchemy/src/cloudflare/website.ts
```diff
@@ -353,21 +353,71 @@ async function runDevCommand(
...
+      if (entry.startsWith(pidPrefix) && entry.endsWith(".pid")) {
```
sam-goodwin, line 365: why do we need to check `pidPrefix`?

alchemy/src/cloudflare/website.ts
```diff
@@ -353,21 +353,71 @@ async function runDevCommand(
...
-  if (await exists(persistFile)) {
-    const pid = Number.parseInt(await fs.readFile(persistFile, "utf8"));
-    try {
-      // Actually kill the process if it's alive
-      process.kill(pid, "SIGTERM");
-    } catch {
-      // ignore
+  // Directory storing per-process PID files: .alchemy/pids/{id}-{pid}.pid
+  const pidDir = path.join(process.cwd(), ".alchemy", "pids");
+  const pidPrefix = `${props.id}-`;
```
sam-goodwin, line 358: `props.id` is not globally unique if `id` is the Resource ID.

We could instead compute one from `resource.chain` (and remove any `/`, or store them as nested folders like we do for file system state store)

alchemy/src/cloudflare/website.ts
```diff
@@ -353,21 +353,71 @@ async function runDevCommand(
...
+      try {
+        process.kill(-legacyPid, "SIGTERM");
+      } catch {
+        // ignore
+      }
+      try {
+        process.kill(legacyPid, "SIGTERM");
+      } catch {
+        // ignore
+      }
```
sam-goodwin, line 411: Seeing this duplicated many times. Can we simplify the code with a few helpers?

john-royal:
We figured out a different solution in #839, #840, and #848, so closing this for now.

PR #839, opened 2025-08-12
fix(cli): improve exit signal handling

john-royal:
The CLI wasn't waiting for the child process to exit cleanly. This fixes that - but we'll want to make sure it doesn't cause any other issues.

### DIFF
--- a/alchemy/src/alchemy.ts
+++ b/alchemy/src/alchemy.ts
@@ -194,7 +194,6 @@ If this is a mistake, you can disable this check by setting the ALCHEMY_CI_STATE
     onExit((code) => {
       root.cleanup().then(() => {
         code = code === 130 ? 0 : (code ?? 0);
-        console.log("Exiting", code);
         process.exit(code);
       });
       return true;


alchemy/bin/services/execute-alchemy.ts
```diff
@@ -151,25 +152,41 @@ export async function execAlchemy(
...
-      console.error(error.stderr);
+  });
+  onExit(() => {
+    if (!controller.signal.aborted) {
+      controller.abort();
+      return true;
+    }
+  });
+  console.log(command);
+  const child = execa(command, {
```
sam-goodwin: we spoke about it in stand up - let's ditch `execa` and maintain control/transparency

alchemy/bin/services/execute-alchemy.ts
```diff
@@ -150,26 +151,23 @@ export async function execAlchemy(
...
-    if (error.stderr) {
-      console.error(error.stderr);
-    }
-    throw error;
-  }
+  console.log(command);
+  const [cmd, ...cmdArgs] = command.split(" ");
+  const child = spawn(cmd, cmdArgs, {
+    cwd,
+    shell: true,
```
sam-goodwin, line 163: if you set `shell: true` I don't think you ned to split the command

alchemy/bin/services/execute-alchemy.ts
```diff
@@ -150,26 +151,23 @@ export async function execAlchemy(
...
-      cwd,
-      shell: true,
-      stdio: "inherit",
-      env: {
-        ...process.env,
-        FORCE_COLOR: "1",
-      },
-    });
-  } catch (error: any) {
-    log.error(pc.red(`Deploy failed: ${error.message}`));
```
sam-goodwin, line 166: Looks like we lost this log?

alchemy/src/alchemy.ts
```diff
@@ -193,6 +193,7 @@ If this is a mistake, you can disable this check by setting the ALCHEMY_CI_STATE
     });
     onExit((code) => {
       root.cleanup().then(() => {
+        code = code === 130 ? 0 : (code ?? 0);
```
sam-goodwin, line 196: can this have any false positives?

PR #840, opened 2025-08-12
fix(cloudflare): make website exit handling more robust

john-royal:
This adds a "belt and suspenders" approach to make sure the website dev command is actually exited when it's supposed to, before trying to start another process.


sam-goodwin:
Does this replace? https://github.com/sam-goodwin/alchemy/pull/833

alchemy/src/cloudflare/website.ts
```diff
@@ -449,3 +439,34 @@ async function runDevCommand(
...
+
+const exitProcess = async (pid: number, signal: "SIGTERM" | "SIGKILL") => {
+  try {
+    process.kill(pid, signal);
+  } catch {
+    return;
+  }
+  for (let i = 0; i < 10; i++) {
+    try {
+      process.kill(pid, 0);
```
sam-goodwin: nit: i also did this in my change and found it confusing that `kill(pid, 0)` is how you check it's still alive. Worth a comment

sam-goodwin (approved):

alchemy/src/cloudflare/website.ts
```diff
@@ -352,22 +353,79 @@ async function writeMiniflareSymlink(cwd: string) {
...
     const pid = Number.parseInt(await fs.readFile(persistFile, "utf8"));
-    await exitProcess(pid, "SIGTERM");
-    await removePersistFile(persistFile);
+    try {
+      // Check if process is still alive
+      process.kill(pid, 0);
+      // Process is still running, reattach to the log stream
+      logger.log(
+        `Dev server for ${props.name} is already running (pid: ${pid}), reattaching to log stream...`,
+      );
```
sam-goodwin: Does this have potential to lose/miss logs?

alchemy/src/cloudflare/website.ts
```diff
@@ -352,22 +353,79 @@ async function writeMiniflareSymlink(cwd: string) {
...
+
+      // Read existing log content to find URL and display it
+      const logContent = await fs.readFile(logFile, "utf8");
+
+      // Stream new log content as it's written
+      const logStream = createReadStream(logFile, {
+        encoding: "utf8",
+        start: logContent.length,
+      });
```
sam-goodwin: This might duplicate data by replaying logs on every run

sam-goodwin: Think we might want to do this from the `alchemy` CLI process instead?

john-royal:
We'll want to merge #848 first. Let me know if I should close this one - not sure how to deal with the branches/diffs here since the changes are sort of intertwined.

PR #845, opened 2025-08-13
fix: use --env-file-if-exists for alchemy deploy

### LINKED ISSUE #844: Use `--env-file-if-exists` for alchemy deploy
@claude `alchemy deploy` uses `--env-file` flag under the hood which causes issues in CI env as there might not be a `.env` file in CI
by default the command tries to load `.env` file which is causing issues

ref:
https://nodejs.org/en/learn/command-line/how-to-read-environment-variables-from-nodejs
https://github.com/sam-goodwin/alchemy/blob/25f23ef44c755c20406889d6f3ea72882dbca168/alchemy/bin/services/execute-alchemy.ts#L86

BlankParticle:
Changes --env-file to --env-file-if-exists in execute-alchemy.ts to prevent errors in CI environments where .env files may not exist.

Fixes #844

### DIFF
--- a/alchemy/bin/services/execute-alchemy.ts
+++ b/alchemy/bin/services/execute-alchemy.ts
@@ -83,7 +83,7 @@ export async function execAlchemy(
     execArgs.push("--watch");
     args.push("--watch");
   }
-  if (envFile) execArgs.push(`--env-file ${envFile}`);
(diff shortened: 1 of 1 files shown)


BlankParticle:
also related, fix for the same issue #846 
probably should merge both

PR #846, opened 2025-08-13
Fix alchemy requiring env file

focux:


sam-goodwin:
What was the motivation for this change? is it because you don't want it to load it by default? We load it by default to mirror bun's behavior.

focux:
alchemy was crashing when the `.env` didn't exist, #845 fix this error but i thought the default was an error but it seems like it was purposely add it

PR #849, opened 2025-08-13
fix(cloudflare): r2 bucket empty fails if bucket does not exist

john-royal:
If an R2 bucket has `empty: true` but the bucket does not exist, the destroy will fail when trying to empty the bucket because of a not found error.

This handles the error by detecting the `10006` error code ("the specified bucket does not exist") when listing objects in the bucket.

### DIFF
--- a/alchemy/src/cloudflare/bucket.ts
+++ b/alchemy/src/cloudflare/bucket.ts
@@ -5,8 +5,8 @@ import { Scope } from "../scope.ts";
 import { withExponentialBackoff } from "../util/retry.ts";
 import { CloudflareApiError } from "./api-error.ts";
 import {
-  extractCloudflareError,
   extractCloudflareResult,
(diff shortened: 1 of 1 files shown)


alchemy/src/cloudflare/bucket.ts
```diff
@@ -518,19 +518,27 @@ export async function* listObjects(
...
       per_page: number;
     };
+    success: boolean;
+    errors: CloudflareApiErrorPayload[];
   } = await response.json();
+  if (!json.success) {
+    // 10006 indicates that the bucket does not exist, so there are no objects to list
+    if (json.errors.some((e) => e.code === 10006)) {
+      return;
+    }
```
sam-goodwin, line 535: it doesn't return `http 404`? Grr cloudflare.

sam-goodwin (approved):

Issue #814, opened 2025-08-07
Durable Object Websockets take time to start


Mkassabov:
from discord

> when using alchemy dev + DO with WS.
> If I try to send a WS message immediately after getting the open event on the client, it crashes the miniflare ws proxy as well.
> I need to add a small delay in dev before sending the event, otherwise it always crashes.
> This is the error:
>
>```
> /Users/[redacted]/node_modules/.pnpm/ws@8.18.3/node_modules/ws/lib/websocket.js:450
>      throw new Error('WebSocket is not open: readyState 0 (CONNECTING)');
>      ^
>
> Error: WebSocket is not open: readyState 0 (CONNECTING)
> ```

john-royal:
Right now we use a Node-based HTTP/WS proxy instead of another Miniflare service like Wrangler does. Might be time to revisit that so the behavior lines up. The reason I used a Node proxy is because of some runtime compatibility issues, specifically with Bun.

Or perhaps we can just queue up pending messages until the connection is fully open. That would be simpler, but not as robust — using a Miniflare proxy enables better handling of certain states.

Either way, I'll try and figure this out today.

sam-goodwin:
Why is a miniflare proxy more reliable than a node one? I don't understand.

PR #851, opened 2025-08-13
chore: add e2e test for durable-object-websocket example

john-royal:
This is related to #814, but doesn't reproduce the error.


john-royal:
this was merged separately thanks

PR #855, opened 2025-08-13
fix(cli): ensure exit code 0 for sigint

john-royal:
When you Ctrl + C out of the CLI, sometimes it shows an error even when it's a clean exit. This handles that by rewriting exit code 130 to 0.

### DIFF
--- a/alchemy/bin/services/execute-alchemy.ts
+++ b/alchemy/bin/services/execute-alchemy.ts
@@ -157,7 +157,7 @@ export async function execAlchemy(
   }
   process.on("SIGINT", async () => {
     await exitPromise;
-    process.exit(child.exitCode ?? 0);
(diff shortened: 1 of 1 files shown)


alchemy/bin/services/execute-alchemy.ts
```diff
@@ -173,5 +173,10 @@ export async function execAlchemy(
   });
   const exitPromise = once(child, "exit");
   await exitPromise.catch(() => {});
-  process.exit(child.exitCode ?? 0);
+  process.exit(formatExitCode(child.exitCode));
 }
+
+const formatExitCode = (exitCode: number | null) => {
```
sam-goodwin: nit: `sanitizeExitCode`

and add comment. it's a bit magical

PR #858, opened 2025-08-14
feat(cli): append .alchemy to .gitignore during init

AmanVarshney01:


alchemy/bin/commands/init.ts
```diff
@@ -482,6 +482,18 @@ async function updatePackageJson(context: InitContext): Promise<void> {
   }
 }
 
+async function updateGitignore(context: InitContext) {
+  try {
+    const gitignorePath = resolve(context.cwd, ".gitignore");
+
+    await fs.ensureFile(gitignorePath);
+
+    await fs.appendFile(gitignorePath, ".alchemy\n", "utf-8");
```
sam-goodwin: Should you check if it already exists?

AmanVarshney01: fs.ensureFile creates the file if it doesn’t exist, and does nothing if it already exists.


https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/ensureFile.md

sam-goodwin: I meant check if `.alchemy` is already in the `.gitignore`

alchemy/bin/commands/init.ts
```diff
@@ -488,7 +488,22 @@ async function updateGitignore(context: InitContext) {
...
+    const lines = gitignoreContent.split("\n").map((line) => line.trim());
+    const hasAlchemy = lines.some(
+      (line) => line === ".alchemy" || line === ".alchemy/",
+    );
+
+    if (!hasAlchemy) {
+      if (gitignoreContent && !gitignoreContent.endsWith("\n")) {
+        gitignoreContent += "\n";
+      }
+      await fs.appendFile(gitignorePath, ".alchemy\n", "utf-8");
```
sam-goodwin, line 505: nit: it's simpler:
```ts
await fs.writeFile(gitignorePath, [...lines, ".alchemy"].join("\n"))
```

PR #862, opened 2025-08-14
fix(cli): replace kleur with picocolors

AmanVarshney01:
- use single color library `picocolors` instead of two


sam-goodwin:
lol we just got rid of execa for a reason. Cc @john-royal 

Why do we need execa @AmanVarshney01 ? It was messing with exit codes and seems like an unnecessary layer

PR #869, opened 2025-08-15
chore: remove unused resources and dependencies

john-royal:
This is an experiment; no obligation to merge. I want to see what our download times look like with a thorough audit.

## Comparison

Did some anecdotal testing. Appears to reduce the number of packages for a new install down from 400 to 250, but install times were about the same. I suspected that `pkg.pr.new` was a little slower than NPM, so for an apples-to-apples comparison, I compared the preview package for #866 (minimal differences compared to `main`) to this branch and found a roughly 10 second difference between clean installs (including `pnpm store prune` to clear the local cache) with pnpm: from 26.043 seconds to 16.584.

## Changes

1. Removed the `alchemy/ai` resources (see #868).
2. Removed `alchemy/runtime` and `alchemy/web`.
3. Removed resources like `StaticJsonFile` and `StaticTypeScriptFile` from `alchemy/fs`.
4. Audited peer dependencies.
5. The TanStack Start example and the SvelteKit template both included extraneous `bun.lock` files, which I removed (found them while searching for references to dependencies).

## Notes on Peer Dependencies

Still working on this, but notes on peer dependencies thus far:
- `arktype`: only used as types aside from alchemy/ai and alchemy/web resources, so made it a dev dependency.
- `hono`: Only used in templates, not in Alchemy itself.
- `cloudflare`: Not imported anywhere.
- `@aws-sdk/client-ec2`: Only used in tests. Moved to dev dependencies.
- `@aws-sdk/client-sagemaker`: Not imported anywhere.
- `@aws-sdk/client-ses`: Not imported anywhere; all imports reference `@aws-sdk/client-sesv2`.
- other `@aws-sdk/*` peers:  Marked as optional, with a dynamic import at runtime that catches the error and reminds you to add the relevant peer dependency.
- `prettier`: Removed; was only used to format `cloudflare::WranglerJson`
- `@libsql/client` and `drizzle-orm`: Marked as optional; SQLiteStateStore generally handles missing peers gracefully but plugged the remaining holes
- `stripe`: marked as optional; added dynamic import with error handling; requires `stripe` to be a dynamic import which made `createStripeClient` async


alchemy/package.json
```diff
@@ -204,17 +184,63 @@
...
+    "prettier": {
+      "optional": true
+    },
+    "stripe": {
+      "optional": true
+    }
+  },
   "devDependencies": {
     "@asteasolutions/zod-to-openapi": "^7.3.0",
+    "@aws-sdk/client-ec2": "^3.868.0",
```
john-royal: If the peer dependency is marked as optional, then it won't be installed in our repo unless it's also here as a dev dependency, hence the duplication

alchemy/src/state/d1-state-store.ts
```diff
@@ -34,7 +34,12 @@ export class D1StateStore extends StateStoreProxy {
 }
 
 const createDatabaseClient = memoize(async (options: D1StateStoreOptions) => {
-  const { createCloudflareApi } = await import("../cloudflare/api.ts");
+  const { drizzle } = await import("drizzle-orm/sqlite-proxy").catch(() => {
+    throw new Error(
+      "Drizzle not found. Please add drizzle-orm to your project dependencies.",
+    );
+  });
+  const { createCloudflareApi } = await import("../cloudflare/api.js");
```
sam-goodwin: this doesn't need to be lazy does it?

alchemy/src/state/sqlite-state-store.ts
```diff
@@ -131,6 +130,7 @@ async function createBunSQLiteDatabase(
     );
   });
   const { migrate } = await import("drizzle-orm/bun-sqlite/migrator");
+  const schema = await import("./schema.js");
```
sam-goodwin, line 133: not sure these need to be lazy

john-royal: The way I have this set up is anything that imports Drizzle is lazy so you get a better runtime error if you don't have Drizzle installed. `schema.js` is imported lazily because that imports Drizzle at the top level.

I'm thinking it might be good to add a `utils/import-peer-dependency.ts` to consolidate this sort of "dynamic import/catch/throw friendly error" behavior since it's all over the place now.

alchemy/src/state/d1-state-store.ts
john-royal: Good catch. IIRC `createCloudflareApi` was lazily imported because `D1StateStore` was imported from `scope.ts`, which caused a nasty circular dependency without the lazy import. That shouldn't be a problem anymore so I'll fix it.

alchemy/package.json
sam-goodwin: consistently use `^3` or at least align all of the client-* packages to the same version

PR #874, opened 2025-08-15
fix(core): clean up processes created with idempotentSpawn

john-royal:


alchemy/src/util/idempotent-spawn.ts
```diff
@@ -69,7 +72,10 @@ export async function idempotentSpawn({
     write,
   });
 
-  return extracted;
+  return {
+    extracted,
+    stop: async () => (pid ? await kill(pid) : undefined),
```
sam-goodwin, line 77: Are we calling this somewhere

john-royal: Scroll down?

https://github.com/sam-goodwin/alchemy/pull/874/files#diff-5860b8c8986f4f840a2754e1513d0b35b9604b06637e2d1d9840a43cdd171c3cR306

sam-goodwin: Didn't render properly on phone lol

PR #881, opened 2025-08-17
feat(cloudflare): flatten Website into Worker.assets

sam-goodwin:
```ts
export const website = await Worker("website", {
  name: `${app.name}-${app.stage}-website`,
  assets: ViteAssets(),
  adopt: true,
  bindings: {
    KV: kv,
    ALCHEMY_TEST_VALUE: alchemy.secret("Hello from Alchemy!"),
  },
});
```

### DIFF
--- a/alchemy/src/cloudflare/vite/vite.ts
+++ b/alchemy/src/cloudflare/vite/vite.ts
@@ -1,6 +1,7 @@
 import { getPackageManagerRunner } from "../../util/detect-package-manager.ts";
 import type { Assets } from "../assets.ts";
 import type { Bindings } from "../bindings.ts";
+import { WebsitePlugin, type WebsitePluginProps } from "../website-plugin.ts";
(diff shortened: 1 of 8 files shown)


john-royal:
Not sure how I feel about including the commands as part of `assets` because for SSR frameworks, you're not just configuring assets — you're typically also configuring the built entrypoint. Wonder if something like `preset:` would be better.

sam-goodwin:
Hmm I closed this one. Did I close the wrong PR? Lol

sam-goodwin:
> Not sure how I feel about including the commands as part of `assets` because for SSR frameworks, you're not just configuring assets — you're typically also configuring the built entrypoint. Wonder if something like `preset:` would be better.

I don't like this design either. Had a long discussion in discord that may have been missed.

Going to stick with wrapping but with:
- ViteProps.server and WorkerProps.main instead of entrypoint. 
- Remove the nested Scope and move wrangler json into Worker

PR #883, opened 2025-08-17
fix(cloudflare): allow React Router apps to be deployed in SPA mode

john-royal:
React Router doesn't currently work in SPA mode because the `ReactRouter` resource always expects a server entrypoint. To fix this:

- [x] Read `react-router.config.ts` to detect `ssr: false` (React Router v7 defaults to `ssr: true` unless the file is present and `ssr` is set to `false`; see [React Router docs](https://reactrouter.com/api/framework-conventions/react-router.config.ts))
- [x] If SSR is disabled, enable SPA mode in the Vite resource and avoid autofilling `worker.entrypoint` or `wrangler.main`

### DIFF
--- a/alchemy/src/cloudflare/react-router/react-router.ts
+++ b/alchemy/src/cloudflare/react-router/react-router.ts
@@ -1,4 +1,6 @@
+import path from "node:path";
 import { getPackageManagerRunner } from "../../util/detect-package-manager.ts";
+import { logger } from "../../util/logger.ts";
 import type { Assets } from "../assets.ts";
 import type { Bindings } from "../bindings.ts";
 import { Vite, type ViteProps } from "../vite/vite.ts";
@@ -21,6 +23,8 @@ export async function ReactRouter<B extends Bindings>(
   props: ReactRouterProps<B> = {},
 ): Promise<ReactRouter<B>> {
   const runner = await getPackageManagerRunner();
+  const cwd = path.resolve(props.cwd ?? process.cwd());
+  const ssr = await detectSSREnabled(cwd);
   return await Vite(id, {
     ...props,
     build:
@@ -29,13 +33,50 @@ export async function ReactRouter<B extends Bindings>(
     dev:
       props.dev ??
       `${runner} react-router typegen && ${runner} react-router dev`,
+    spa: !ssr,
     compatibility: "node",
-    entrypoint: props.entrypoint ?? "build/server/index.js",
+    entrypoint: props.entrypoint ?? (ssr ? "build/server/index.js" : undefined),
     noBundle: props.noBundle ?? true,
     assets: props.assets ?? "build/client",
     wrangler: {
-      main: props.wrangler?.main ?? props.main ?? "workers/app.ts",
+      main:
+        props.wrangler?.main ??
+        props.main ??
+        (ssr ? "workers/app.ts" : undefined),
       transform: props.wrangler?.transform,
     },
   });
+}
+
+/**
+ * Detect if SSR is enabled by checking for an `ssr` property in `react-router.config.ts`.
+ * If no config is found or if there is no `ssr` property, default to true in line with React Router's default behavior.
+ * @see https://reactrouter.com/api/framework-conventions/react-router.config.ts
+ */
+async function detectSSREnabled(cwd: string): Promise<boolean> {
+  const candidates = [
+    "react-router.config.mjs",
+    "react-router.config.js",
+    "react-router.config.ts",
+    "react-router.config.mts",
+  ];
+  try {
+    const config = await Promise.any(
+      candidates.map((candidate) => import(path.join(cwd, candidate))),
+    );
+    if (
+      typeof config.default === "object" &&
+      config.default &&
+      "ssr" in config.default &&
+      typeof config.default.ssr === "boolean"
+    ) {
+      return config.default.ssr;
+    }
+    return true;
+  } catch {
+    logger.warn(
+      "[ReactRouter] No react-router.config.{ts,js,mts,mjs} found, assuming SSR is enabled",
+    );
+    return true;
+  }
 }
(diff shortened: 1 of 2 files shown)


alchemy/src/cloudflare/react-router/react-router.ts
```diff
@@ -29,13 +33,50 @@ export async function ReactRouter<B extends Bindings>(
...
+ */
+async function detectSSREnabled(cwd: string): Promise<boolean> {
+  const candidates = [
+    "react-router.config.mjs",
+    "react-router.config.js",
+    "react-router.config.ts",
+    "react-router.config.mts",
+  ];
+  try {
+    const config = await Promise.any(
```
john-royal: `Promise.any` returns the first resolved promise or rejects with an AggregateError if they all reject, good for matching multiple files.

alchemy/src/cloudflare/react-router/react-router.ts
```diff
@@ -29,13 +33,50 @@ export async function ReactRouter<B extends Bindings>(
...
+      config.default &&
+      "ssr" in config.default &&
+      typeof config.default.ssr === "boolean"
+    ) {
+      return config.default.ssr;
+    }
+    return true;
+  } catch {
+    logger.warn(
+      "[ReactRouter] No react-router.config.{ts,js,mts,mjs} found, assuming SSR is enabled",
```
john-royal: Not sure if this warning is really necessary because the config file isn't required.

alchemy/src/cloudflare/react-router/react-router.ts
```diff
@@ -29,13 +33,50 @@ export async function ReactRouter<B extends Bindings>(
...
 }
+
+/**
+ * Detect if SSR is enabled by checking for an `ssr` property in `react-router.config.ts`.
+ * If no config is found or if there is no `ssr` property, default to true in line with React Router's default behavior.
+ * @see https://reactrouter.com/api/framework-conventions/react-router.config.ts
+ */
+async function detectSSREnabled(cwd: string): Promise<boolean> {
+  const candidates = [
+    "react-router.config.ts",
```
john-royal, line 57: Just realized — we have a similar pattern for the `Astro` resource, but if you're not using Bun, dynamic imports of `.ts` files probably won't work.

@sam-goodwin Any thoughts on the right solution? Is it [jiti](https://github.com/unjs/jiti) time?

sam-goodwin: What would that look like to use here?

I'm also not against just saying we are node 24 and up (modern JS runtime only)

alchemy/src/cloudflare/react-router/react-router.ts
sam-goodwin: We need a verbose option to put these kinds of logs behind. 

I tend to agree it would be annoying/noisy.

alchemy/src/cloudflare/react-router/react-router.ts
```diff
@@ -29,13 +33,50 @@ export async function ReactRouter<B extends Bindings>(
...
-    entrypoint: props.entrypoint ?? "build/server/index.js",
+    entrypoint: props.entrypoint ?? (ssr ? "build/server/index.js" : undefined),
     noBundle: props.noBundle ?? true,
     assets: props.assets ?? "build/client",
     wrangler: {
-      main: props.wrangler?.main ?? props.main ?? "workers/app.ts",
+      main:
+        props.wrangler?.main ??
+        props.main ??
+        (ssr ? "workers/app.ts" : undefined),
```
sam-goodwin, line 44: How does undefined affect the behavior? Do we just inject the no-op worker implementation?

sam-goodwin: Should we add a smoke test for this?

alchemy/src/cloudflare/react-router/react-router.ts
sam-goodwin: What if multiple exist? Then it would be a race condition?

I suspect it's rare and a user error, but maybe it's better to have deterministic behavior by scanning .ts to .mjs to .js?

alchemy/src/cloudflare/react-router/react-router.ts
```diff
@@ -29,13 +33,50 @@ export async function ReactRouter<B extends Bindings>(
...
+  ];
+  try {
+    const config = await Promise.any(
+      candidates.map((candidate) => import(path.join(cwd, candidate))),
+    );
+    if (
+      typeof config.default === "object" &&
+      config.default &&
+      "ssr" in config.default &&
+      typeof config.default.ssr === "boolean"
```
sam-goodwin: Could we do it simpler with:

```ts
config?.default?.ssr
```

alchemy/src/cloudflare/react-router/react-router.ts
john-royal: Instead of an `import()` call it would be `jiti.import()` which handles TS.

Honestly, I love the idea of dropping support for older Node versions because it would make our job a lot easier (native `Promise.withResolvers`!) but not sure if that's the right approach. We do have a decent chunk of users on 20 and 22, so my hunch is it's probably best to keep supporting LTS versions.

The other catch is that IIRC, you still need that `--experimental-strip-types` option for TS to work in Node, so the behavior could be unpredictable even if you're on the newest version. Feel free to correct me if I'm wrong.

alchemy/src/cloudflare/react-router/react-router.ts
john-royal: Good catch, let me do that.

alchemy/src/cloudflare/react-router/react-router.ts
john-royal: Good catch. Made a change so it's deterministic.

alchemy/src/cloudflare/react-router/react-router.ts
john-royal: Yes, the behavior for a `Website` with no entrypoint is we use a stub worker for serving the SPA. Smoke test is a good idea, I can add that now.

PR #885, opened 2025-08-17
feat(cloudflare): flatten Website scope hierarchy

sam-goodwin:
The `Website` wrapper used to create the Worker under a nested scope with `alchemy.run`, but this caused user friction when moving from a Website to a plain Worker.

This new change removes the nested scope and changes `WranglerJson` to just a function instead of a Resource.

This should be a seamless transition for users, provided my heuristic works:
1. during deletion of a Worker, check for a `sibling` resource of `id=wrangler.jsonc, kind="cloudflare::WranglerJson"` and skip deletion
2. during creation of a Worker, check for a `child` resource of `id=wrangler.jsonc, kind="cloudflare::WranglerJson"` and set `adopt=true`


Mkassabov:
2 things:
- its not caused by this PR, so it seems we have a regression in main somewhere: #911 
- this causes issues when downgrading from the flat website to the legacy website resource (IDK if we care about this, sine nobody should really be downgrading alchemy)

sam-goodwin:
I don't think we care about downgrading and there's not much we can do because we can't change old versions

Issue #886, opened 2025-08-17
local d1Database migrations apply to a secondary empty .sqlite db instead of applying to the one the worker actively queries against


Rhayxz:
When running a Worker locally with a local D1 database, migrations are not being applied to the same .sqlite database file that the Worker actually queries against. Instead, they seem to run on an empty copy of the database, leaving the active one without any migrations.

As a result, after the first local run:

The Worker’s database remains stuck on the initial migration. -> All subsequent schema changes (new tables, dropped tables, etc.) are never reflected in the Worker’s queries.

Meanwhile, the .alchemy/miniflare/v3 folder contains a new SQLite db with the correct latest migrations, but that db is empty and never used by the Worker.

Steps to Reproduce:

1- Create a Worker with a D1 database.

2- Add migrations to create 5 tables.
3- Run the Worker locally → Worker db has 5 tables.
4-Modify the migrations folder and add a .sql file to drop 4 tables.

Run the Worker again → Worker db still has all 5 tables, stuck with a db that has the schema from the first migration run.

Inspect .alchemy/miniflare/v3 → a new SQLite file exists with the correct schema (only 1 table), but the Worker never queries it. And you will find the a

Expected Behavior:
Migrations should be applied to the same SQLite db file that the Worker queries against, so that schema changes are reflected locally.

Actual Behavior:
Migrations are applied to a fresh db file each time under .alchemy/miniflare/v3, not to the Worker’s used .sqlite file.

Command ran: `bun --watch --env-file .env.dev ./apps/infra/alchemy.run.ts --stage dev --dev`

PR #889, opened 2025-08-18
fix(d1): use stable dev.id for applyLocalD1Migrations databaseId in local dev

Rhayxz:
### Summary
Fix local D1 migrations applying to a secondary empty SQLite DB instead of the DB the Worker actually uses.

Fixes sam-goodwin/alchemy#886.

### Root Cause
In local dev, subsequent runs passed databaseId: "" into applyLocalD1Migrations due to:
  databaseId: this.output?.id ?? id

After the first run, this.output?.id becomes an empty string (""), which is truthy in TS, so databaseId is always "".  
Miniflare then creates a fresh DB under .alchemy/miniflare/v3 and applies migrations there, while the Worker continues querying the original DB whose schema never updates.

### The Change
Use the stable dev-phase id instead:
  databaseId: dev.id

where:
  const dev = { id: this.output?.dev?.id ?? this.output?.id ?? id, ... }

This keeps the same identifier across phases so migrations always target the exact DB the Worker binds to.

### Results
- Migrations apply to the same SQLite file the Worker queries.
- No extra fresh DBs are created under .alchemy/miniflare/v3 on subsequent runs.
- Verified by:
  1. Creating 5 tables via migrations
  2. Dropping 4 tables in a new migration
  3. Observing d1_migrations and table list through a test endpoint: schema updates now reflect in the active DB.

### Notes
- No behavior change in non-dev modes.
- Minimal surface area: only swaps the databaseId source.

### DIFF
--- a/alchemy/src/cloudflare/d1-database.ts
+++ b/alchemy/src/cloudflare/d1-database.ts
@@ -273,7 +273,7 @@ const _D1Database = Resource(
     if (local) {
       if (props.migrationsFiles && props.migrationsFiles.length > 0) {
         await applyLocalD1Migrations({
-          databaseId: this.output?.id ?? id,
+          databaseId: dev.id,
           migrationsTable: props.migrationsTable ?? DEFAULT_MIGRATIONS_TABLE,
           migrations: props.migrationsFiles,
         });


john-royal:
Excellent job pinning this down. I think this is good to merge.

sam-goodwin:
Released in `0.62.3`

PR #892, opened 2025-08-18
fix(cloudflare): skip path validation during astro output type check

### LINKED ISSUE #860: Unable to resolve alchemy server type if using the alchemy astro vite adapter
When running `alchemy deploy` we are unable to infer to correct output type for the cloudflare astro resource.

This is the function we use to infer the astro output type.

```ts
async function resolveOutputType(cwd: string): Promise<"server" | "static"> {
  const candidates = [
    "astro.config.mjs",
    "astro.config.js",
    "astro.config.ts",
    "astro.config.mts",
  ];
  for (const candidate of candidates) {
    try {
      const config = await import(join(cwd, candidate));
      if (
        typeof config.default === "object" &&
        config.default &&
        "output" in config.default &&
        typeof config.default.output === "string"
      ) {
        return config.default.output;
      }
      return "static";
    } catch {
      // ignore
    }
  }
  return "static";
}

```

The problem is when we try and import an astro config that references the alchemy vite adapter it will error with
```
41 |
42 | const warned = new Set<string>();
43 |
44 | const warnOrThrow = (message: string, throws: boolean) => {
45 |   if (throws) {
46 |     throw new Error(message);
               ^
error: The Wrangler config path, ".alchemy/local/wrangler.jsonc", could not be found. Please run `alchemy dev` or `alchemy deploy` to create it.
      at warnOrThrow (D:\code\OSS\alchemy\alchemy\alchemy\src\cloudflare\miniflare\paths.ts:46:11)
      at validateConfigPath (D:\code\OSS\alchemy\alchemy\alchemy\src\cloudflare\miniflare\paths.ts:11:5)
      at getPlatformProxyOptions (D:\code\OSS\alchemy\alchemy\alchemy\src\cloudflare\cloudflare-env-proxy.ts:42:17)
      at alchemy (D:\code\OSS\alchemy\alchemy\alchemy\src\cloudflare\astro\plugin.ts:6:20)
      at D:\worktemp\alchemy\astro-mono-test\deployments\astro\astro.config.mjs:8:11
      at requestImportModule (2:1)
```

alchemy dev works fine.

john-royal:
Closes #860.

Uses `AsyncLocalStorage` to detect when Miniflare path validation calls should not cause an error to be thrown.

### DIFF
--- a/alchemy/src/cloudflare/miniflare/paths.ts
+++ b/alchemy/src/cloudflare/miniflare/paths.ts
@@ -1,4 +1,14 @@
+import { AsyncLocalStorage } from "node:async_hooks";
 import { existsSync, readlinkSync, statSync } from "node:fs";
+
+const dynamicImportContext = new AsyncLocalStorage<boolean>();
+
+/**
+ * Used to disable path validation during dynamic imports.
+ */
+export const withSkipPathValidation = <T>(callback: () => T) => {
+  return dynamicImportContext.run(true, callback);
+};
 
 export const DEFAULT_CONFIG_PATH = ".alchemy/local/wrangler.jsonc";
 export const DEFAULT_PERSIST_PATH = ".alchemy/miniflare/v3";
@@ -42,6 +52,9 @@ export const validatePersistPath = (
 const warned = new Set<string>();
 
 const warnOrThrow = (message: string, throws: boolean) => {
+  if (dynamicImportContext.getStore()) {
+    return;
+  }
   if (throws) {
     throw new Error(message);
   }
--- a/alchemy/src/cloudflare/astro/astro.ts
+++ b/alchemy/src/cloudflare/astro/astro.ts
@@ -2,6 +2,7 @@ import { join, resolve } from "node:path";
 import { getPackageManagerRunner } from "../../util/detect-package-manager.ts";
 import type { Assets } from "../assets.ts";
 import type { Bindings } from "../bindings.ts";
+import { withSkipPathValidation } from "../miniflare/paths.ts";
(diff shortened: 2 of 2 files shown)


alchemy/src/cloudflare/miniflare/paths.ts
```diff
@@ -1,5 +1,15 @@
+import { AsyncLocalStorage } from "node:async_hooks";
 import { existsSync, readlinkSync, statSync } from "node:fs";
 
+const dynamicImportContext = new AsyncLocalStorage<boolean>();
```
sam-goodwin, line 4: is async local storage necessary? what was the justification?

Mkassabov:
can confirm this does fix my issue on #860

alchemy/src/cloudflare/miniflare/paths.ts
john-royal: I was split between 3 ways of doing this:
1. environment variable
2. `globalThis`
3. async local storage

Figured #3 was the least leaky - no additional global, keeps it isolated.

Happy to change it if you want

Mkassabov: isn't asl more leaky? I'd expect this to break if I ran two astro resources at the same time (e.g. with `promise.all([astro1, astro2])`)

john-royal: Why would that break?

We're using `.run()` instead of `.enter()` so it's scoped to the Astro dynamic import. Running multiple in parallel shouldn't be an issue.

PR #893, opened 2025-08-18
feat: generate physical names from app, stage and resource ID

sam-goodwin:
Now, when you create a Worker without specifying a name:
```ts
const app = await alchemy("my-app");

await Worker("worker")
```

And run it with `--stage`:
```sh
alchemy deploy --stage prod
```

The name will be automatically derived:
```
my-app-prod-worker
```

If you create within a nested scope, the scope chain will be included in the physical name:
```ts
await alchemy.run("nested', async () => {
  await Worker("worker")
})
```

The id will be:
```sh
my-app-prod-nested-worker
```

Question: how do we feel about the stage not being a suffix? e.g. we could also do it as `${app}-${id}-${stage}`
```sh
my-app-worker-prod
my-app-worker-foo-prod
```
or `${stage}-${app}-${id}`:
```sh
prod-my-app-worker
prod-my-app-foo-worker # nested scope
```


ericclemmons:
For the nested worker should the output be “nested” and not “foo”?

alchemy/src/cloudflare/kv-namespace.ts
```diff
@@ -209,7 +211,13 @@ const _KVNamespace = Resource(
     id: string,
     props: KVNamespaceProps,
   ): Promise<KVNamespace> {
-    const title = props.title ?? id;
+    const title =
+      props.title ?? this.output?.title ?? this.scope.createPhysicalName(id);
+
+    if (this.phase === "update" && this.output?.title !== title) {
+      this.replace();
+    }
```
sam-goodwin: Might be able to rename

alchemy/src/cloudflare/ai-gateway.ts
```diff
@@ -211,44 +231,45 @@ const AiGatewayResource = Resource(
...
-      logger.error(`Error ${this.phase} AI Gateway '${id}':`, error);
+      logger.error(`Error ${this.phase} AI Gateway '${gatewayName}':`, error);
       throw error; // Re-throw the error to fail the deployment
     }
 
     // Construct the output object from API response and merged props
     return this({
       ...mergedProps, // Start with the input props (including defaults)
       id: apiResource.id,
+      gatewayName,
```
sam-goodwin, line 254: We forgot to include this, so we should try and resolve it by ID

alchemy/src/cloudflare/ruleset.ts
```diff
@@ -172,10 +174,12 @@ export const Ruleset = Resource("cloudflare::Ruleset", async function <
     return this.destroy();
   }
 
+  const rulesetName = props.name ?? this.scope.createPhysicalName(id);
```
sam-goodwin: retain old name

sam-goodwin:
> For the nested worker should the output be “nested” and not “foo”?

Oops that was a typo - yes it should.

PR #897, opened 2025-08-20
fix(cloudflare): ensure D1 migrations run on fresh installs in local dev mode

acoyfellow:
## Problem
D1 migrations don't run on fresh installs because Alchemy skips the resource when props haven't changed, even though migrations need to execute.

## Solution
Force resource update when migrations exist in local dev mode:

```typescript
// Force resource update when migrations exist in local dev mode
if (migrationsFiles.length > 0 && Scope.current.local) {
  (Scope.current as any).force = true;
}
```

Fixes the issue where users had to manually save migration files to trigger migrations on fresh `alchemy dev` runs.


sam-goodwin:
Is this happening because your state store is the CloudflareStateStore (remote) and then you blow away all the local miniflare state (but keep the remote state) and then expect it to detect that the state is clear and to re-run migrations?

sam-goodwin:
Pushed up a change that does this in a more supported way

acoyfellow:
> Is this happening because your state store is the CloudflareStateStore (remote) and then you blow away all the local miniflare state (but keep the remote state) and then expect it to detect that the state is clear and to re-run migrations?

potentially - i didn't think about that. i will poke around

sam-goodwin:
We can merge this change by the way, i don't see any reason not to do it? Migrations in dev mode are cheap so it's fine to run them like this

acoyfellow:
> Is this happening because your state store is the CloudflareStateStore (remote) and then you blow away all the local miniflare state (but keep the remote state) and then expect it to detect that the state is clear and to re-run migrations?

I can confirm now, yes.. This was leading to some other confusing things for me too:

```
stateStore: process.env.NODE_ENV === 'production' || process.env.CI
    ? (scope: any) => new CloudflareStateStore(scope, {
        scriptName: `${projectName}-ci-state`,
      })
    : undefined
```

Clears it up.

Mkassabov (approved):

PR #902, opened 2025-08-21
feat(planetscale): add postgres support

john-royal:
Haven't had a chance to test this yet due to PlanetScale account difficulties.

BTW: I noticed that the create API call includes fields that aren't documented as part of create but are available as part of update. Not sure if this was intentional but might be worth evaluating in a separate issue.


alchemy/src/planetscale/database.ts
```diff
@@ -89,13 +90,24 @@ export interface DatabaseProps {
...
    */
   clusterSize: PlanetScaleClusterSize;
+
+  /**
+   * The CPU architecture of the database (create only).
+   * Only available for PostgreSQL databases.
+   * ARM provides superior cost-to-performance, while x86 is more mature and provides maximum extension compatibility.
+   * @default "x86"
+   */
+  arch?: "arm" | "x86";
```
sam-goodwin, line 110: should we use discriminated unions to enforce this?
```ts
& ({ kind: "mysql" } | { kind: "postgres", arch?: "arm" | "x86" })
```

john-royal:
Replacing with #918

PR #903, opened 2025-08-21
fix(cloudflare): set `spa: false` for astro resource

john-royal:
The `Website` resource defaults to `spa: true` and is not overridden by the `Astro` resource. This may cause Astro websites to mistakenly be interpreted as single-page apps.


sam-goodwin:
There's nothing interesting in our sitemap? 
image

PR #904, opened 2025-08-21
chore: add robots.txt to docs

john-royal:
Our website doesn't have a robots.txt, which may prevent LLMs from reading our docs.

### DIFF
--- a/alchemy-web/public/robots.txt
+++ b/alchemy-web/public/robots.txt
@@ -1,0 +1,4 @@
+User-agent: *
+Allow: /
+
+Sitemap: https://alchemy.run/sitemap-index.xml


alchemy-web/public/robots.txt
```diff
@@ -0,0 +1,4 @@
+User-agent: *
+Allow: /
+
+Sitemap: https://alchemy.run/sitemap-index.xml
```
sam-goodwin, line 4: Do we have a sitemap?

john-royal: Yes, Starlight generates it automatically: [link](https://alchemy.run/sitemap-index.xml).

PR #916, opened 2025-08-27
fix(cloudflare): sync node compat plugin with workers-sdk

### LINKED ISSUE #913: Dynamic require of "X" is not supported
I have a Turbo + PNPM monorepo, with an `apps/api` that contains a basic Hono app.

Wrangler dev/deploy (v4.33.0) both work and deploy to CF Workers successfully, but I'm having trouble setting up a working example with Alchemy.

I've tried a straight `Worker` setup (toggling the `compatibility` and `format` flags), bundling with `Esbuild`, and running via both Bun and Node.

In all cases, I get errors like the following:

```
[creating]  api Creating Resource...
service core:user:api-test: Uncaught Error: Dynamic require of "buffer/index.js" is not supported
  at .alchemy/out/api-test/server.js:12:9
  at .alchemy/out/api-test/server.js:2941:20 in node_modules/.pnpm/iron-webcrypto@0.2.8/node_modules/iron-webcrypto/dist/index.js
  at .alchemy/out/api-test/server.js:18:50 in __require2
  at .alchemy/out/api-test/server.js:3245:25 in node_modules/.pnpm/iron-session@6.3.1_express@5.1.0_next@15.5.0_@opentelemetry+api@1.9.0_react-dom@19.1.1_react@19.1.1__react@19.1.1_/node_modules/iron-session/edge/dist/index.js
  at .alchemy/out/api-test/server.js:18:50 in __require2
  at .alchemy/out/api-test/server.js:3511:18 in node_modules/.pnpm/@workos-inc+node@7.62.0_express@5.1.0_next@15.5.0_@opentelemetry+api@1.9.0_react-dom@19.1.1_react@19.1.1__react@19.1.1_/node_modules/@workos-inc/node/lib/common/iron-session/edge-iron-session-provider.js
  at .alchemy/out/api-test/server.js:18:50 in __require2
  at .alchemy/out/api-test/server.js:17505:40 in node_modules/.pnpm/@workos-inc+node@7.62.0_express@5.1.0_next@15.5.0_@opentelemetry+api@1.9.0_react-dom@19.1.1_react@19.1.1__react@19.1.1_/node_modules/@workos-inc/node/lib/index.worker.js
  at .alchemy/out/api-test/server.js:18:50 in __require2
  at .alchemy/out/api-test/server.js:38936:28
```

As you can see from the stack trace, there are various NPM imports -- it fails every time it hits a dynamic `require`.

Using Esbuild, I'm able to get past some of them by setting `external`, but I'm also not sure that this is the right approach since Wrangler works fine without any additional config.

The dir structure is fairly straightforward:

```
├── package.json
├── src
│   └── server.ts
├── tsconfig.json
├── worker-configuration.d.ts
└── wrangler.jsonc
```

**package.json:**

```json
{
  "name": "@myorg/api",
  "version": "0.1.0",
  "type": "module",
  "private": true,
  "scripts": {
    "dev": "wrangler dev",
    "deploy:cf": "wrangler deploy",
    "check-types": "tsc --noEmit",
    "cf-typegen": "wrangler types --include-runtime=false"
  },
  "dependencies": {
    "hono": "catalog:",
    "zod": "catalog:"
  },
  "devDependencies": {
    "@myorg/ts-config": "workspace:*",
    "@myorg/rest": "workspace:*",
    "wrangler": "catalog:"
  }
}
```

**wrangler.jsonc:**

```json
{
  "$schema": "../../node_modules/wrangler/config-schema.json",
  "account_id": "xxx",
  "compatibility_date": "2025-08-16",
  "compatibility_flags": ["nodejs_compat"],
  "main": "src/server.ts",
  "name": "api",
  "route": {
    "custom_domain": true,
    "pattern": "api.my-org.dev"
  },
  "observability": {
    "enabled": true,
    "logs": { "enabled": true }
  }
}
```

**alchemy.run.ts** in the project root:

```ts

const app = await alchemy("my-app", {
  stage: "dev",
  password: "some totally secure password here",
});

  accountId: "xxx",
  name: "api",
  entrypoint: "apps/api/src/server.ts",
  domains: ["api.my-org.dev"],
  compatibility: "node",
  format: "esm",
  observability: {
    enabled: true,
  },
  bindings: {
    // a fair amount of env vrs
});

console.log(`Worker deployed at: ${worker.url}`);
await app.finalize();
```

---

Any ideas where I could be going wrong? Thanks in advance.

john-royal:
Updates our ESBuild plugins to match Wrangler. Closes #913. Bundle tests are passing.

Changes:
- Copied the default Node compat and ALS plugins from Wrangler.
- Removed hard-coded externals and used the Wrangler plugins instead.
- Use `import { getCloudflarePreset } from "@cloudflare/unenv-preset"` instead of `import { cloudflare }`. This smarter implementation takes compatibility dates/flags into account. The old `import { cloudflare }` is deprecated. From the `index.d.ts` for the package.
- Added a WorkOS bundle test to reproduce the issue from 913.


alchemy/src/cloudflare/bundle/plugin-hybrid-node-compat.ts
```diff
@@ -144,17 +152,12 @@ function handleUnenvAliasedPackages(
...
     // Convert `require()` calls for NPM packages to a virtual ES Module that can be imported avoiding the require calls.
     // Note: Does not apply to Node.js packages that are handled in `handleRequireCallsToNodeJSBuiltins`
     if (
       args.kind === "require-call" &&
       (unresolvedAlias.startsWith("unenv/npm/") ||
-        unresolvedAlias.startsWith("unenv/mock/") ||
-        // this does not exist in the original wrangler code but does seem to resolve the problem
-        // TODO(sam): is this the right solution? Why does the same code work in wrangler?
-        unresolvedAlias.startsWith("@cloudflare/unenv-preset/npm/") ||
-        unresolvedAlias.startsWith("@cloudflare/unenv-preset/mock/"))
```
sam-goodwin, line 157: Ah does your change explain why I needed to do this?

alchemy/src/cloudflare/bundle/plugin-hybrid-node-compat.ts
```diff
@@ -259,8 +262,9 @@ function handleNodeJSGlobals(
 
     return {
       contents: dedent`
-        import { ${imports.join(", ")} } from "${module}";
-        ${injects.map(({ injectedName, importName }) => `globalThis.${injectedName} = ${importName};`).join("\n")}`,
+				import { ${imports.join(", ")} } from "${module}";
+				${injects.map(({ injectedName, importName }) => `globalThis.${injectedName} = ${importName};`).join("\n")}
+			`,
```
sam-goodwin: Unnecessary indentation change. You should still maintain 2-space indentation like a code block for multi line string literals.

alchemy/src/cloudflare/bundle/plugin-node-compat.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    // Prevent multiple warnings per package
+    const warnedPackages = new Map<string, string[]>();
+
+    pluginBuild.onStart(() => {
+      seen.clear();
+      warnedPackages.clear();
+    });
+    pluginBuild.onResolve(
+      { filter: /node:.*/ },
```
sam-goodwin: what about `"fs/promises"` instead of `node:fs/promises`

alchemy/src/cloudflare/bundle/plugin-hybrid-node-compat.ts
john-royal: Good catch. That was just from copy/pasting. Will fix.

alchemy/src/cloudflare/bundle/plugin-node-compat.ts
john-royal: Ah, so the reason this wasn't needed in Wrangler is because esbuild catches it and treats it as a `node:` import. But then it doesn't trigger our warning. Fixed.

alchemy/src/cloudflare/bundle/plugin-hybrid-node-compat.ts
john-royal: Found it! We made that change because of the `debug` module ([commit](https://github.com/sam-goodwin/alchemy/pull/741/commits/a947a2dcbae8de74a65fb9cdbcc050b872775c1c)).

Wrangler handles the `debug` module's interaction with unenv differently, so we're using that now instead:
```ts
const { alias, inject, external, polyfill } = defineEnv({
  presets: [
    getCloudflarePreset({
      compatibilityDate,
      compatibilityFlags,
    }),
    {
      alias: {
        // Force esbuild to use the node implementation of debug instead of unenv's no-op stub.
        // The alias is processed by handleUnenvAliasedPackages which uses require.resolve().
        debug: "debug",
      },
    },
  ],
  npmShims: true,
}).env;
```

alchemy/src/cloudflare/bundle/plugin-node-compat.ts
```diff
@@ -1,10 +1,13 @@
 import type { Plugin } from "esbuild";
 import type { NodeJSCompatMode } from "miniflare";
+import { builtinModules } from "node:module";
 import { relative } from "node:path";
 import chalk from "picocolors";
 import { dedent } from "../../util/dedent.ts";
 import { logger } from "../../util/logger.ts";
 
+const NODEJS_MODULES_RE = new RegExp(`^(node:)?(${builtinModules.join("|")})$`);
```
sam-goodwin: By the way, this can have differing behavior on `bun` because `builtinModules` is different. Not sure what I had before, but worth checking we didn't regress anything there.

john-royal: good catch. added that back

PR #917, opened 2025-08-27
chore(planetscale): use generated api client

john-royal:
Replaces the current API client with one generated from the PlanetScale API. Includes types for development and much more useful error messages.

Very few logic changes, mainly to fix failing tests:
- `Password` resource replace logic was a little wonky; tests weren't passing.
- There were one or two API calls to undocumented endpoints to verify that some settings were deleted by checking for a 404. Turns out the reason why they were returning 404 is because the endpoints didn't exist anyway — if you try calling them, you get a blank response instead of the JSON-formatted error PlanetScale uses to indicate a resource doesn't exist. These were in addition to other checks that do work as expected, so I removed these redundant ones.

### DIFF
--- a/alchemy/package.json
+++ b/alchemy/package.json
@@ -182,7 +182,7 @@
     "@aws-sdk/client-ssm": "^3.0.0",
     "@aws-sdk/client-sts": "^3.0.0",
     "@cloudflare/vite-plugin": "^1.10.1",
-    "@cloudflare/workers-types": "*",
+    "@cloudflare/workers-types": "catalog:",
     "@libsql/client": "^0.15.12",
     "@octokit/rest": "^21.1.1",
     "astro": "^5.13.2",
--- a/bun.lock
+++ b/bun.lock
@@ -116,7 +116,7 @@
         "@aws-sdk/client-ssm": "^3.0.0",
         "@aws-sdk/client-sts": "^3.0.0",
         "@cloudflare/vite-plugin": "^1.10.1",
-        "@cloudflare/workers-types": "*",
+        "@cloudflare/workers-types": "catalog:",
         "@libsql/client": "^0.15.12",
         "@octokit/rest": "^21.1.1",
         "astro": "^5.13.2",


alchemy/src/planetscale/utils.ts
```diff
@@ -25,71 +25,80 @@ export type PlanetScaleClusterSize =
...
-  db: string,
+  api: PlanetScaleClient,
+  organization: string,
+  database: string,
   branch: string,
   keyspace: string,
 ): Promise<void> {
   const start = Date.now();
-  let delay = 100;
+  let delay = 1000;
```
Mkassabov: we do we need to increase this delay?

alchemy/src/planetscale/utils.ts
```diff
@@ -98,92 +107,88 @@ export async function waitForKeyspaceReady(
     }
 
     await new Promise((r) => setTimeout(r, delay));
-    delay = Math.min(delay * 2, 1_000);
+    delay = Math.min(delay * 2, 5_000);
```
Mkassabov: also why are we increasing this delay?

scripts/planetscale/template.ts
```diff
@@ -0,0 +1,175 @@
+// This file is generated by scripts/planetscale/generate.ts. Do not edit manually.
```
Mkassabov: can we rename this file to `template.gen.ts`? I like the convention of making it clear this is a generated file

alchemy/src/planetscale/utils.ts
john-royal: Great question — must've copy/pasted the wrong thing. Postgres takes forever but keyspaces are a Vitess thing. Will revert.

scripts/planetscale/template.ts
john-royal: The actual generated client is `client.gen.ts`. In the Postgres PR I removed the banner and added it manually at build time. Thoughts?

alchemy/src/planetscale/utils.ts
john-royal: Good question. Reverted.

I think I just frantically increased a bunch of timeouts because I was hitting them in tests.

scripts/planetscale/template.ts
Mkassabov: Not sure I follow? You removed the banner, and have the build step add it every time?

What's the point of the banner over the gen.ts name? Or both?

alchemy/src/planetscale/api/client.gen.ts
```diff
@@ -0,0 +1,2882 @@
+// This file is auto-generated by scripts/planetscale/generate.ts. Do not edit manually.
```
john-royal, line 1: @Mkassabov The actual generated file is `.gen.ts` and has the banner.

Removed it from the template to make it clear that the template isn't actually generated.

I guess I find it a little weird to put `.gen.ts` on a file that isn't actually generated, but I don't have too strong of an opinion.

Mkassabov: Ohhhhhh, this file isn't generated. I missunderstood that part.
Then yeah, removing the banner and doing nothing else is fine

PR #918, opened 2025-08-27
feat(planetscale): postgres support

### LINKED ISSUE #525: Planetscale Postgres support
https://x.com/planetscale/status/1940065701156532352?s=46&t=DPy1AX-U06PrVdZ4pYROKg

https://planetscale.com/blog/planetscale-for-postgres

john-royal:
Closes #525 (this time for real).

Changes:
- [x] Updated `Database` to set `kind` and `arch`
  - [x] Updated `DatabaseProps` to use a discriminated union so you can only set `arch` for postgres databases
- [x] Created `Role` resource (similar to `Password`, but the endpoints and fields are completely different so wasn't able to reuse it) - included tests
- [x] Updated `Branch` resource and cluster resize logic to detect and handle Postgres resizes
- [x] Updated tests for `Database` and `Branch` to check both MySQL and Postgres
- [x] Got tests passing - had to increase timeouts because Postgres is slow to initialize
- [x] Update docs

# Notes
- The PlanetScale tests now take about 20-25 minutes to run — Postgres clusters take a few minutes to initialize, and 10-15 minutes to resize.
- Assuming we don't need a smoke test for this since the integration tests are passing. Waiting for these to initialize would be inconvenient anyway.

### DIFF
--- a/alchemy/src/planetscale/role.ts
+++ b/alchemy/src/planetscale/role.ts
@@ -206,7 +206,8 @@ export const Role = Resource(
                 // Essentially, the role is still linked to a branch, so we can't delete it here, but it'll be deleted when the branch is deleted.
                 logger.warn(
                   [
-                    `Delete role "${id}" (name: "${this.output.name}") failed with status ${res.error.status} (${res.error.code}): ${res.error.message}`,
+                    `Failed to delete role "${id}" (name: "${this.output.name}").`,
+                    `Error: ${res.error.message}`,
                     `If you are deleting database "${database}" or branch "${branch}", the role will be deleted automatically.`,
                     `Otherwise, consider manually deleting the role at: https://app.planetscale.com/${organization}/${database}/settings/roles`,
                   ].join("\n"),


alchemy/src/test/bun.ts
```diff
@@ -53,9 +53,9 @@ type test = {
    */
   skipIf(condition: boolean): test;
 
-  beforeAll(fn: (scope: Scope) => Promise<void>): void;
+  beforeAll(fn: (scope: Scope) => Promise<void>, timeout?: number): void;
```
john-royal, line 56: Note on this: Bun doesn't seem to support timeouts for `beforeAll`, but had to add this here as a type, otherwise I got a type error from adding this only to Vitest.

Had to add a timeout option to `beforeAll` because of Postgres tests taking forever.

john-royal:
Weird type error showing up in tests. Is it safe to remove the `@ts-expect-error` directives?

I'd love to get to the bottom of why this is showing up here but not in #918.

```bash
$ bun run build && bun tsc -b && biome check
$ cd ./alchemy && bunx tsc -b && bunx tsc -p tsconfig.test.json && bun run --filter alchemy build
Error: src/cloudflare/durable-object-namespace.ts(22,3): error TS2578: Unused '@ts-expect-error' directive.
Error: src/neon/project.ts(639,9): error TS2578: Unused '@ts-expect-error' directive.
error: script "build" exited with code 1
error: script "check" exited with code 1
Error: Process completed with exit code 1.
```

alchemy/src/planetscale/role.ts
```diff
@@ -0,0 +1,240 @@
...
+          const res = await api.organizations.databases.branches.roles.delete({
+            path: {
+              organization,
+              database,
+              branch,
+              id: this.output.id,
+            },
+            result: "full",
+          });
+          // TODO: (422 unprocessable): Role is still referenced and cannot be dropped
```
john-royal: Any idea what to do about this error?

If you delete the database or branch then the role is deleted automatically so this shouldn't technically be blocking.

alchemy/src/planetscale/role.ts
```diff
@@ -0,0 +1,240 @@
...
+
+  /**
+   * Time to live in seconds
+   */
+  ttl?: number;
+
+  /**
+   * Roles to inherit from
+   */
+  inheritedRoles?: Array<
```
john-royal: There are some complications here. This should probably either be required or have a default value, otherwise you can't read or write anything.

The `postgres` role is the default role with all permissions — skeptical about the security of making that the default, plus you tend to get a 422 error when deleting it because you can't delete the default role. Then again, I'd be worried about footguns if we don't allow all permissions by default.

alchemy/src/test/bun.ts
sam-goodwin: We can probably delete bun since we don't need it and would only use it after they add parallel tests.

sam-goodwin:
> Weird type error showing up in tests. Is it safe to remove the `@ts-expect-error` directives?
> 
> I'd love to get to the bottom of why this is showing up here but not in #918.
> 
> ```shell
> $ bun run build && bun tsc -b && biome check
> $ cd ./alchemy && bunx tsc -b && bunx tsc -p tsconfig.test.json && bun run --filter alchemy build
> Error: src/cloudflare/durable-object-namespace.ts(22,3): error TS2578: Unused '@ts-expect-error' directive.
> Error: src/neon/project.ts(639,9): error TS2578: Unused '@ts-expect-error' directive.
> error: script "build" exited with code 1
> error: script "check" exited with code 1
> Error: Process completed with exit code 1.
> ```

Merge with main, i think i fixed these.

sam-goodwin:
Still waiting on docs?

john-royal:
> Still waiting on docs?

Sorry, I forgot to ask. Can you let me know your thoughts on:
- [this comment on role permissions](https://github.com/sam-goodwin/alchemy/pull/918#discussion_r2304557538)
- [this comment on role deletion](https://github.com/sam-goodwin/alchemy/pull/918#discussion_r2304557538)

I have some docs ready but feel like we should address the role problem before release. Current behavior isn't especially intuitive.

alchemy/src/planetscale/role.ts
sam-goodwin: Maybe make it required and use this rune to suggest `"postgres" | (string & {})` in the auto complete, but also make sure to document that it is admin? 

Should we also support referencing another `Role`?

john-royal: I don't see why not! Yeah let's do it.

alchemy/src/planetscale/role.ts
sam-goodwin: That looks fatal. Or is it something we need to wait for?

For now let's just crash and let users report issues.

PR #921, opened 2025-08-28
feat(cloudflare): debugging support for cf workers

Mkassabov:
## What this does
- CDPServer now allows multiple clients to be connected, with each client having their own domains
- CDPServer now does message playback on domain enablement
- workers managed by miniflare now create a CDP proxy and get added to the CDP manager for debugging

### DIFF
--- a/alchemy/bin/services/cdp-manager/cdp-proxy.ts
+++ b/alchemy/bin/services/cdp-manager/cdp-proxy.ts
@@ -1,5 +1,9 @@
 import type { WebSocket as WsWebSocket } from "ws";
-import { CDPServer, type ClientCDPMessage } from "./server.ts";
+import {
+  CDPServer,
+  type ClientCDPMessage,
+  type ServerCDPMessage,
+} from "./server.ts";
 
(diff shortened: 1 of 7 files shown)


Mkassabov:
Enough changes are here that this warrants testing debugging for:
- [ ] cloudflare workers
  - [x] cf web debugger
  - [x] chrome native debugger
- [ ] alchemy.run.ts
  - [ ] via tsx 
    - [ ] vscode debugging
    - [ ] cf web debugger
    - ~~[ ] vscode debug terminal~~
    - [ ] chrome native debugger
  - [ ] via bun 
    - [x] vscode debugging
    - [ ] bun web debugger
    - ~~[ ] vscode debug terminal~~
    - ~~[ ] chrome native debugger~~

Issues
- CF breakpoints not working
- playback not working
- debug terminal doesn't work because we launch multiple inspectors. IT MIGHT work under a model where everything is unified in one server, but I don't think thats what we want to do here

Mkassabov:
closing this cause its outdated but will keep the branch around since its still a good reference

PR #927, opened 2025-08-29
fix(cloudflare): enforce stateToken requirement in CloudflareStateSto…

### LINKED ISSUE #926: create vite flow fails with `Secret cannot be undefined`
See the screenshots below, this is happening for me every single time I try to bootstrap a vite application from the tutorial. https://alchemy.run/guides/cloudflare-vitejs/#init

Flow is:
```
bunx alchemy create my-react-app --template=vite
cd my-react-app
bun alchemy login // follow steps to login with CF
bun run dev/deploy
```

Either `dev` or `deploy` both fail with the same issue as below.

Image

Image

MattieTK:
This fixes #926 which is occurring on all template starters, or anything that requires this `ALCHEMY_STATE_TOKEN` env var but doesn't specify it in the instructions. This just makes it a bit clearer when the error occurs by throwing before it gets to the secrets error handler.

### DIFF
--- a/alchemy/src/state/cloudflare-state-store.ts
+++ b/alchemy/src/state/cloudflare-state-store.ts
@@ -49,16 +49,11 @@ export class CloudflareStateStore extends StateStoreProxy {
     super(scope);
     if (!options.stateToken && !process.env.ALCHEMY_STATE_TOKEN) {
       throw new Error(
-        "Missing token for DOStateStore. Please set ALCHEMY_STATE_TOKEN in the environment or set the `stateToken` option in the DOStateStore constructor. See https://alchemy.run/guides/cloudflare-state-store/",
+        "Missing token for CloudflareStateStore. Please set ALCHEMY_STATE_TOKEN in the environment or set the `stateToken` option in the CloudflareStateStore constructor. See https://alchemy.run/guides/cloudflare-state-store/",
       );
     }
     const stateToken =
       options.stateToken ?? alchemy.secret(process.env.ALCHEMY_STATE_TOKEN);
-    if (!stateToken) {
-      throw new Error(
-        "Missing token for DOStateStore. Please set ALCHEMY_STATE_TOKEN in the environment or set the `stateToken` option in the DOStateStore constructor.",
-      );
-    }
     this.options = {
       ...options,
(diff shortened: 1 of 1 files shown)


sam-goodwin:
Thank you @MattieTK, this is a helpful change.

FYI - we could have also done `alchemy.env.ALCHEMY_STATE_TOKEN` and it would have at least complained that `ALCHEMY_STATE_TOKEN is undefined`, but your error message is more helpful.

alchemy/src/state/cloudflare-state-store.ts
```diff
@@ -47,6 +47,11 @@ export class CloudflareStateStore extends StateStoreProxy {
   options: CloudflareStateStoreOptions & { stateToken: Secret<string> };
   constructor(scope: Scope, options: CloudflareStateStoreOptions = {}) {
     super(scope);
+    if (!options.stateToken && !process.env.ALCHEMY_STATE_TOKEN) {
+      throw new Error(
+        "Missing token for DOStateStore. Please set ALCHEMY_STATE_TOKEN in the environment or set the `stateToken` option in the DOStateStore constructor. See https://alchemy.run/guides/cloudflare-state-store/",
```
sam-goodwin: Should say `CloudflareStateStore`

alchemy/src/state/cloudflare-state-store.ts
```diff
@@ -47,6 +47,11 @@ export class CloudflareStateStore extends StateStoreProxy {
...
+      );
+    }
     const stateToken =
       options.stateToken ?? alchemy.secret(process.env.ALCHEMY_STATE_TOKEN);
```
sam-goodwin, line 56: can be `alchemy.secret.env.ALCHEMY_STATE_TOKEN`, but is not important if we have the error message above (which is better)

MattieTK:
Thanks Sam, I can make those changes, was just copying what was already there to get myself unblocked, but happy to contribute :)

alchemy/src/state/cloudflare-state-store.ts
MattieTK: Updated in latest commit

PR #928, opened 2025-08-30
feat(cloudflare): add Workers.DevUrl and Workers.DevDomain bindings

sam-goodwin:
A Worker by default is allocated a `workers.dev` subdomain computed from its name and your account's worker subdomain.

E.g.
```sh
${workerName}.${subdomain}.workers.dev
```

Use the `Worker.DevDomain` and `Worker.DevUrl` bindings to inject a Worker's own subdomain or URL into its script.

```ts

  name: "api-worker",
  entrypoint: "./src/api.ts",
  // workers.dev URL must be enabled (is by default)
  // url: true,
  bindings: {
    // ${workerName}.${subdomain}.workers.dev
    DEV_DOMAIN: Worker.DevDomain,
    // https://${workerName}.${subdomain}.workers.dev
    DEV_URL: Worker.DevUrl,
  },
});
```

### DIFF
--- a/alchemy/src/cloudflare/miniflare/build-worker-options.ts
+++ b/alchemy/src/cloudflare/miniflare/build-worker-options.ts
@@ -59,6 +59,7 @@ type BaseWorkerOptions = {
 };
 
 export const buildWorkerOptions = async (
+  url: string,
   input: MiniflareWorkerInput,
 ): Promise<{
   watch: (signal: AbortSignal) => AsyncGenerator<miniflare.WorkerOptions>;
@@ -97,7 +98,7 @@ export const buildWorkerOptions = async (
       continue;
     }
     if (binding === Worker.DevDomain || binding === Worker.DevUrl) {
-      (options.bindings ??= {})[key] = "TODO";
+      (options.bindings ??= {})[key] = binding === Worker.DevUrl ? url : url.replace("http://", "").replace("https://", "");
       continue;
     }
     switch (binding.type) {
--- a/alchemy/src/cloudflare/miniflare/miniflare-controller.ts
+++ b/alchemy/src/cloudflare/miniflare/miniflare-controller.ts
@@ -30,7 +30,13 @@ export class MiniflareController {
   }
 
   async add(input: MiniflareWorkerInput) {
-    const { watch, remoteProxy } = await buildWorkerOptions(input);
+    const proxy = new MiniflareWorkerProxy({
+      name: input.name,
+      // TODO(sam): this is a race condition. We can let it select it on server.listen(), but then it switches on each re-run
+      port: input.port ?? (await findOpenPort()),
+      miniflare: this.update(),
+    });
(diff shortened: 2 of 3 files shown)


john-royal:
A few questions:
- Why do we have both `DevDomain` and `DevUrl`? How common is it to need the domain without the http(s) prefix?
- Should this also take worker versions into account? Would definitely make it more complicated since I'm not sure if there's a way to determine the version ID without injecting something into the worker runtime that can read a version metadata binding.

sam-goodwin:
Good idea from @Mkassabov, why not just bundle it with other data into a single always-available binding.
```ts
type MEtadata = {
  url: string;
  devDomain: string;
  commit: string;
  appName: string;
  stage: string;
  scope: string[];
  // .. what else?
}
```

ericclemmons:
From [Discord](https://discord.com/channels/1359694195782320389/1433922744436916324), I'd like to have a stable URL across all environments to rely on.

`import.meta.url` is a thing, so I wouldn't be surprised to have `import.meta.env.*` to match.

Screenshot 2025-10-31 at 6 09 44 PM
Screenshot 2025-10-31 at 6 09 52 PM

I naturally expected that https://alchemy.run/providers/cloudflare/worker/#self-binding would give me a reference back to `SELF.url` like `worker.url` is in my `alchemy.run.ts`.

But, now I understand that it's a _binding_ to itself (`.fetch | .connect`).

I **cannot** `import { worker } from ".../alchemy.run.ts"` because that has compilation issues in Vite (`@libsql`, `sqlite`, Bun, all showed up in the error stack).  **If it had worked, I would've just used `worker.url` directly**.

vrabota:
@ericclemmons @sam-goodwin can we use `SELF.url` like `worker.url` is in my alchemy.run.ts, trying to something like this
CleanShot 2025-12-13 at 08 38 19@2x

sam-goodwin:
@vrabota you can use `SELF_URL: Worker.DevUrl`

vrabota:
@sam-goodwin seems like not working 
CleanShot 2025-12-13 at 08 45 15@2x

sam-goodwin:
@vrabota :
1. are yu on latest alchemy?
2. are you importing `{ Worker } from "alchemy/cloudflare"`?

vrabota:
@sam-goodwin package update fixed , here is my setup 
CleanShot 2025-12-13 at 09 00 01@2x
now getting this 
CleanShot 2025-12-13 at 09 00 39@2x

vrabota:
CleanShot 2025-12-13 at 09 03 55@2x
added `dev.domain` like this, but still throwing an error

sam-goodwin:
Ok, found a bug: https://github.com/alchemy-run/alchemy/pull/1264

sam-goodwin:
@vrabota can you try `https://pkg.pr.new/alchemy-run/alchemy@d4545cd` and see if it works. I also added logic to infer the port so you should hopefully be able to drop `dev.domain` too.

vrabota:
@sam-goodwin yeah tested all good, thanks :)

PR #930, opened 2025-09-01
feat(stripe): bump stripe dependency, remove coupon from Customer

Mkassabov:
Seems stripe dropped support or some stuff
1. [deprecate coupon and promotion code on customer](https://docs.stripe.com/changelog/basil/2025-03-31/deprecate-singular-coupon-promotion-code?changelog-sdk=node) => this seems to just have been removed; they replaced it with `discount` but not everywhere and the customer api is one of the places where it was just removed
2. [deprecate aggregate usage on prices](https://docs.stripe.com/changelog/basil/2025-03-31/deprecate-legacy-usage-based-billing
) it seems that we already support meters so we just drop support for the legacy method

### DIFF
--- a/alchemy-web/src/content/docs/providers/stripe/price.md
+++ b/alchemy-web/src/content/docs/providers/stripe/price.md
@@ -42,7 +42,21 @@ const subscriptionPrice = await Price("pro-monthly", {
 Create a metered price for usage-based billing:
 
 ```ts
-import { Price } from "alchemy/stripe";
+import { Price, Meter } from "alchemy/stripe";
+
+const storageUsageMeter = await Meter("storageUsageMeter", {
+  displayName: "Storage Usage",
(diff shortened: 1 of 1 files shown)


alchemy/src/stripe/customer.ts
```diff
@@ -92,10 +92,6 @@ export interface CustomerProps {
    * An integer amount in cents that represents the customer's current balance
    */
   balance?: number;
-  /**
-   * If you provide a coupon code, the customer will have a discount applied on all recurring charges
-   */
-  coupon?: string;
```
sam-goodwin, line 98: This change looks like a breaking change, not a chore?

Mkassabov: yeah called it a chore before I realized bumping stripe was breaking

sam-goodwin:
Need to change the docs?

PR #931, opened 2025-09-01
fix(cloudflare): use "text/javascript" mime type for assets

john-royal:
This fixes the following error:

```
Failed to load module script: Expected a JavaScript-or-Wasm module script but the server responded with a MIME type of "application/javascript+module". Strict MIME type checking is enforced for module scripts per HTML spec.
```

## Explanation

Our `getContentType` function returns `application/javascript` or `application/javascript+module` for JS files. This is necessary for Workers scripts since Cloudflare expects those specific MIME types. However, the most widely accepted MIME type for JS in the browser is `text/javascript`, so we should use that for assets instead.

It's safe to make this change to the `getContentType` function directly because it's used exclusively for assets.


alchemy/src/cloudflare/worker-assets.ts
john-royal, line 1: My first attempt to fix the content type issue didn't work because it turned out we were computing the content type twice: once in `assets.ts` and again here.

Figured I'd deduplicate.

Issue #933, opened 2025-09-02
Queue bindings not included in development for producer workers


Lustyn:
## Problem Description

Alchemy omits queue bindings from the generated development `wrangler.jsonc` when a worker is configured as a **producer-only** for a queue, even when the queue is explicitly defined in the worker's bindings.

## Root Cause

Alchemy only includes queue bindings in the generated development `wrangler.jsonc` [if the worker is considered a consumer](https://github.com/sam-goodwin/alchemy/blob/822847f7ede859c0efd375ce8a8d5abd3d43fe34/alchemy/src/cloudflare/wrangler.json.ts#L847-L849) of the queue. If a worker only produces messages to a queue (has the queue in `bindings` but not in `eventSources`), the queue binding is omitted from the generated configuration.

## Expected Behavior

Queue bindings should be included in `wrangler.jsonc` for any worker that has the queue defined in its `bindings`, regardless of whether it's a consumer or producer.

## Actual Behavior

Queue bindings are only included when the worker is also a consumer (has the queue in `eventSources`).

## Impact

- Producer-only workers cannot send messages to queues at runtime
- Forces unnecessary architectural coupling (workers must be both producers AND consumers)

## Minimal Reproduction

```typescript
import alchemy from "alchemy";
import { ReactRouter, Queue, Worker } from "alchemy/cloudflare";

const app = await alchemy("queue-bug-demo");

// Create a queue for processing tasks
export const taskQueue = await Queue<{
  taskId: string;
  data: any;
}>("task-queue");

// Website worker - PRODUCER ONLY
// This worker will have queue bindings OMITTED from wrangler.jsonc
export const website = await ReactRouter("website", {
  name: `${app.name}-website`,
  bindings: {
    TASK_QUEUE: taskQueue, // This binding will be missing in wrangler.jsonc!
  },
  // No eventSources - not a consumer, only producer
});

// Queue processor worker - CONSUMER ONLY
// This worker will have the queue properly configured as event source
export const queueProcessor = await Worker("queue-processor", {
  entrypoint: "./workers/queue-processor.ts",
  bindings: {
    // Other bindings...
  },
  // This worker consumes from the queue
  eventSources: [taskQueue],
});

await app.finalize();
```

The website worker will have missing queue bindings in the generated `wrangler.jsonc` despite having `TASK_QUEUE` in its bindings.

## Development Mode Considerations

[...]

PR #934, opened 2025-09-02
fix(cloudflare): fix support for producer only queue bindings for wranglerjson

Mkassabov:
should fix: https://github.com/sam-goodwin/alchemy/issues/933

### DIFF
--- a/alchemy/src/cloudflare/wrangler.json.ts
+++ b/alchemy/src/cloudflare/wrangler.json.ts
@@ -844,9 +844,7 @@ function processBindings(
     spec.d1_databases = d1Databases;
   }
 
-  if (queues.consumers.length > 0) {
-    spec.queues = queues;
-  }
+  spec.queues = queues;
 
   if (vectorizeIndexes.length > 0) {
     spec.vectorize = vectorizeIndexes;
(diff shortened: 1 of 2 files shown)


alchemy/src/cloudflare/wrangler.json.ts
```diff
@@ -844,9 +844,7 @@ function processBindings(
     spec.d1_databases = d1Databases;
   }
 
-  if (queues.consumers.length > 0) {
-    spec.queues = queues;
-  }
+  spec.queues = queues;
```
sam-goodwin: Should it be this?
```ts
if (spec.queues.length > 0) {
  spec.queues = queues;
}
```

sam-goodwin (approved):

alchemy/src/cloudflare/wrangler.json.ts
Mkassabov: a little more complicated than that but yeah, fixed

alchemy/src/cloudflare/wrangler.json.ts
```diff
@@ -845,7 +845,10 @@ function processBindings(
   }
 
   if (queues.consumers.length > 0) {
-    spec.queues = queues;
+    (spec.queues ??= {}).consumers = queues.consumers;
+  }
+  if (queues.producers.length > 0) {
+    (spec.queues ??= {}).producers = queues.producers;
   }
```
sam-goodwin, line 852: The original change was this. Was that wrong?

image

Mkassabov: It would always create both consumers and producers asday regardless of what the queue bindings are.

Technically there is no change, it just no follows the convention to only have something in wrangler.jsonc if it needs to be there

sam-goodwin (approved):

Issue #935, opened 2025-09-02
Binding a Worker to a Worker fails if @cloudflare/workers-types not in tsconfig


sam-goodwin:
We currently get a type instantiation error when binding a Worker to a Worker without `@cloudflare/workers-types` in `tsconfig.json`. This should not be required.

```ts
export const worker1 = await Worker("worker1", {
  entrypoint: "src/worker1.ts",
  adopt: true,
  bindings: {
    KV: kv,
    D1: d1,
    R2: r2,
    DO: doNamespace,
  },
  compatibilityFlags: ["nodejs_compat"],
});

export const worker2 = await Worker("worker2", {
  entrypoint: "src/worker2.ts",
  adopt: true,
  bindings: {
    // ERROR: will fail if tsconfig mis configured
    WORKER: worker1,
    DO: worker1.bindings.DO,
  },
  compatibilityFlags: ["nodejs_compat"],
});
```

sam-goodwin:
Ah, tracked it down to:
```ts
import type { Pipeline } from "cloudflare:pipelines";
```

This virtual module comes from `@cloudflare/workers-types`

PR #936, opened 2025-09-02
fix(cloudflare): vendor-in Pipeline Binding type to solve type instantiation error

sam-goodwin:
Our `Bound` type imported the virtual module `cloudflare:pipelines` which I think was causing #935's type instantiation error.

Fixes https://github.com/sam-goodwin/alchemy/issues/935


JacobMGEvans:
Does this need to be revived? Seems to fix a bug?

JacobMGEvans:
Didnt mean to close it -_-

PR #938, opened 2025-09-02
Adding itty-based AWS resources

kmitchener:


alchemy/src/aws/itty/iam/policy.ts
```diff
@@ -0,0 +1,328 @@
...
+            .pipe(
+              Effect.catchTag("NoSuchEntityException", () =>
+                Effect.succeed("Policy doesn't exist"),
+              ),
+            );
+        });
+
+        // FIXME: how should we be handling errors? this can fail for example if the policy is attached to a user or role
+        await Effect.runPromise(deleteEffect);
+        return this.destroy();
```
sam-goodwin: I think just do one Effect for all phases

Mkassabov:
not happening until after effect

PR #942, opened 2025-09-03
fix(core): commands run by scope.spawn are properly mirrored to stdio

### LINKED ISSUE #911: Astro logs hang in dev mode
Whenever I run `alchemy dev` and have an astro resource I get the following logs
```
$ astro dev
```

then I have astro getting served on `localhost:4321` but astro doesn't log that url like it normally does and astro logs don't appear in the console anywhere.

interesting enough `alchemy deploy` shows astro's build logs just fine.

Mkassabov:
fixes #911

long term scope.spawn should probably be farmed out to something we can listen to via our debug server (since its realistically just mirroring `console.log` and `console.error` so CDP works fine or we can have an entire different set of messages for it if we want)

### DIFF
--- a/alchemy/src/util/idempotent-spawn.ts
+++ b/alchemy/src/util/idempotent-spawn.ts
@@ -108,14 +108,31 @@ export async function idempotentSpawn({
   async function spawnLoggedChild() {
     const out = await fsp.open(outPath, "a");
 
-    // shell:true, NO args — pass a single command string
     const child = spawn(cmd, {
       shell: true,
       cwd,
-      stdio: ["ignore", out.fd, out.fd], // stdout/stderr -> files (OS-level)
+      stdio: ["ignore", "pipe", "pipe"],
       env,
       detached: false,
     });
+
+    // Tee stdout/stderr to both file and parent stdio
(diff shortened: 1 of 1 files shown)


alchemy/src/util/idempotent-spawn.ts
```diff
@@ -108,15 +108,32 @@ export async function idempotentSpawn({
   async function spawnLoggedChild() {
     const out = await fsp.open(outPath, "a");
 
-    // shell:true, NO args — pass a single command string
     const child = spawn(cmd, {
       shell: true,
       cwd,
-      stdio: ["ignore", out.fd, out.fd], // stdout/stderr -> files (OS-level)
+      stdio: ["ignore", "pipe", "pipe"],
```
sam-goodwin: This was done on purpose because we can't rely on the in-memory processing. dev mode can kill it at any time, leading to data loss. Using file descriptors puts that in the OS's hands with stronger guarnatees.

sam-goodwin: If you want to read the data, use the `out` file, not the in memory stream.

PR #943, opened 2025-09-04
Durable Object Guide Code Snippet Cleanup

harrysolovay:
An extremely meaningful PR

### DIFF
--- a/alchemy-web/src/content/docs/guides/cloudflare-durable-objects.mdx
+++ b/alchemy-web/src/content/docs/guides/cloudflare-durable-objects.mdx
@@ -50,27 +50,28 @@ This is a step-by-step guide to deploy a Durable Object, for a complete API refe
 
    ```ts
    import type { worker } from "./alchemy.run";
+   import type { DurableObjectState } from "@cloudflare/workers-types";
 
    export class Counter {
      declare env: typeof worker.Env;
-     private count: number;
+     state;
+     count;
 
-     constructor(state, env: typeof worker.Env) {
+     constructor(state: DurableObjectState, env: typeof worker.Env) {
        this.state = state;
        this.count = 0;
      }
 
-     async fetch(request) {
-       const url = new URL(request.url);
-       const path = url.pathname;
+     async fetch(request: Request) {
+       const { pathname } = new URL(request.url);
 
        // Retrieve current count
-       this.count = (await this.state.storage.get("count")) || 0;
+       this.count = (await this.state.storage.get("count") as number) ?? 0;
 
-       if (path === "/increment") {
+       if (pathname === "/increment") {
          this.count++;
          await this.state.storage.put("count", this.count);
-       } else if (path === "/decrement") {
+       } else if (pathname === "/decrement") {
          this.count--;
          await this.state.storage.put("count", this.count);
        }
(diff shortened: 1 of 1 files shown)


alchemy-web/src/content/docs/guides/cloudflare-durable-objects.mdx
```diff
@@ -50,27 +50,28 @@ This is a step-by-step guide to deploy a Durable Object, for a complete API refe
 
    ```ts
    import type { worker } from "./alchemy.run";
+   import type { DurableObjectState } from "@cloudflare/workers-types";
```
john-royal, line 53: Do we want people directly importing `DurableObjectState` from `@cloudflare/workers-types`? AFAIK, you should have `@cloudflare/workers-types` configured in your tsconfig, making `DurableObjectState` a global.

alchemy-web/src/content/docs/guides/cloudflare-durable-objects.mdx
```diff
@@ -50,27 +50,28 @@ This is a step-by-step guide to deploy a Durable Object, for a complete API refe
...
 
    export class Counter {
```
john-royal, line 55: Apologies for the pedantry but IMO, the correct way to do this is:

```ts
// The `DurableObject` class:
// - includes `this.ctx` and `this.env` by default; no need to type manually
// - ensures that the RPC type is correct
// - generic env type is not required, but is used to type `this.env`
export class Counter extends DurableObject<typeof worker.Env> {
  private count!: number;

  constructor(ctx: DurableObjectState, env: typeof worker.Env) {
    super(ctx, env);
    // `ctx.blockConcurrencyWhile` callback runs before any RPC or fetch methods,
    // so `this.count` will be defined before it is used
    ctx.blockConcurrencyWhile(async () => {
      this.count = (await ctx.storage.get("count")) ?? 0;
    });
  }

  override async fetch(request: Request) {
    const { pathname } = new URL(request.url);
    if (pathname === "/increment") {
      this.count++;
      await this.ctx.storage.put("count", this.count);
    } else if (pathname === "/decrement") {
      this.count--;
      await this.ctx.storage.put("count", this.count);
    }
    return Response.json({ count: this.count });
  }
}
```

harrysolovay: The example stores the count in both the DO's memory and the SQLite store. I think this makes it a bit confusing. Might it be worthwhile to use just one?

```ts
import type { worker } from "./alchemy.run";
import { DurableObject } from "cloudflare:workers"

export class Counter extends DurableObject<typeof worker.Env>  {
  override async fetch(request: Request) {
    const { pathname } = new URL(request.url);

    // Retrieve current count
    let count = (await this.ctx.storage.get("count") as number) ?? 0;

    if (pathname === "/increment") {
      count++;
      await this.ctx.storage.put("count", count);
    } else if (pathname === "/decrement") {
      count--;
      await this.ctx.storage.put("count", count);
    }

    return Response.json({ count });
  }
}
```

Also, why would the `private` modifier be necessary?

sam-goodwin:
Lol did I fix the same problem? https://github.com/sam-goodwin/alchemy/pull/949

PR #944, opened 2025-09-04
feat(cloudflare): update oauth client id and storage path

john-royal:
- [x] Use our Cloudflare OAuth client ID
- [x] Store credentials in `${os.homedir()}/.alchemy`
- [x] Update login command and Cloudflare API to use updated implementation

Still testing to figure out the minimum required scopes. So far, it seems that even if we request all available scopes, we still can't create an R2 bucket. I'll keep going through and running tests to identify any other glaring regressions compared to the Wrangler-based implementation.

### DIFF
--- a/alchemy/bin/commands/login.ts
+++ b/alchemy/bin/commands/login.ts
@@ -1,9 +1,9 @@
 import { cancel, intro, log, outro } from "@clack/prompts";
 import pc from "picocolors";
 import { zod as z } from "trpc-cli";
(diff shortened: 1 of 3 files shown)


sam-goodwin:
> So far, it seems that even if we request all available scopes, we still can't create an R2 bucket

Don't hesitate to ask CF for help. They are very responsive.

john-royal:
Replaced by #956

PR #946, opened 2025-09-04
Prevent Pkg New Publishing of Drafts

harrysolovay:


harrysolovay:
Was a WIP to communicate an idea about short-circuit publishing of drafts. Deemed not necessary.

PR #956, opened 2025-09-10
feat(cli): alchemy auth command for cloudflare

john-royal:


alchemy/src/util/neverthrow.ts
john-royal, line 1: These functions were only used for the old Wrangler OAuth implementation. Neverthrow is still used in one other place, so we can remove in separate PR.

alchemy/src/cloudflare/api.ts
```diff
@@ -42,56 +45,119 @@ export interface CloudflareApiOptions {
...
-    authOptions,
-  }));
-}
+    const baseUrl = options.baseUrl ?? process.env.CLOUDFLARE_BASE_URL;
+    const profile = options.profile ?? process.env.CLOUDFLARE_PROFILE;
+
+    if (profile) {
+      const item = await Profile.get({ provider: "cloudflare", profile });
+      if (!item) {
+        throw new Error(`The Cloudflare profile "${profile}" was not found.`);
```
john-royal: Should we handle this better? Any ideas?

alchemy/bin/commands/auth.ts
```diff
@@ -0,0 +1,315 @@
...
+                { label: "API Key", value: "api-key", hint: "Legacy" },
+              ],
+              initialValue: "oauth" as const,
+            }),
+          )),
+        profile:
+          // TODO: should we
+          // - prompt for this by default, or no?
+          // - slightly improve the UX (e.g. "would you like to set a profile name")?
+          // - warn if the profile already exists?
```
sam-goodwin: maybe confirm overwrite

alchemy/bin/commands/auth.ts
```diff
@@ -0,0 +1,315 @@
...
+                { label: "OAuth", value: "oauth", hint: "Recommended" },
+                { label: "Token", value: "token" },
+                { label: "API Key", value: "api-key", hint: "Legacy" },
+              ],
+              initialValue: "oauth" as const,
+            }),
+          )),
+        profile:
+          // TODO: should we
+          // - prompt for this by default, or no?
```
sam-goodwin: I don't follow, looks like we prompt if undefined

alchemy/bin/alchemy.ts
```diff
@@ -29,4 +29,26 @@ const cli = createCli({
     "🧪 Welcome to Alchemy! Creating infrastructure as code with JavaScript and TypeScript.",
 });
 
-cli.run();
+cli.run({
+  argv: rewriteLoginArgs(process.argv.slice(2)),
+});
```
sam-goodwin: Is this a promise?

alchemy/bin/alchemy.ts
```diff
@@ -29,4 +29,26 @@ const cli = createCli({
...
+
+// TODO: is there a better way to alias "alchemy login" to "alchemy auth login"?
```
sam-goodwin: Don't do it, let's just update login?

alchemy/src/cloudflare/api.ts
sam-goodwin: Looks fatal

alchemy/src/cloudflare/api.ts
```diff
@@ -42,56 +45,119 @@ export interface CloudflareApiOptions {
...
-    process.env.CF_ACCOUNT_ID ??
-    (await getCloudflareAccountId(authOptions));
-  return (cloudflareApiCache[cacheKey] = new CloudflareApi({
-    baseUrl: options.baseUrl,
-    accountId,
-    authOptions,
-  }));
-}
+    const baseUrl = options.baseUrl ?? process.env.CLOUDFLARE_BASE_URL;
+    const profile = options.profile ?? process.env.CLOUDFLARE_PROFILE;
```
sam-goodwin: `options.profile ?? CLOUDFLARE_PROFILE ?? ALCHEMY_PROFILE`

alchemy/src/auth.ts
```diff
@@ -0,0 +1,122 @@
...
+   * Update the profile data.
+   */
+  const update = async (updater: (data: Data) => void) => {
+    await mutex.lock(async () => {
+      const data = await list();
+      updater(data);
+      await fs.mkdir(path.dirname(FILE_PATH), { recursive: true });
+      await fs.writeFile(FILE_PATH, JSON.stringify(data, null, 2));
+      await fs.chmod(FILE_PATH, 0o600); // 0o600 means only the current user can read/write
+    });
```
sam-goodwin: This is not safe in a multi-process environment.

Is this only called when we need to refresh? Do we store the access/refresh token in the profile?

How do we minimize the contention here?

alchemy/src/auth.ts
```diff
@@ -0,0 +1,122 @@
...
+   * Set a profile by provider and profile name.
+   */
+  export const set = async (input: Input, profile: Profile) => {
+    await update((data) => {
+      data[input.provider] = {
+        ...(data[input.provider] ?? {}),
+        [input.profile]: profile,
+      };
+    });
+  };
```
sam-goodwin: Is there ever a need to get and set atomically? E.g. "check if i need to refresh, update tokens if i have to, update the stored tokens"?

alchemy/src/auth.ts
john-royal: I updated it so the credentials are stored separately from the profile with a lockfile mechanism to prevent concurrent refreshes.

Profiles are stored in a single `config.json` for simplicity — I don't think those need to be thread-safe since they're unlikely to be modified concurrently.

alchemy/src/auth.ts
john-royal: Great question. This was originally handled in `cloudflare/auth.ts`, but it wasn't safe in a multi-process environment. So I added a `getRefreshedCredentials` function that handles this better.

Just a note: I created the lockfile mechanism myself instead of using `proper-lockfile` like we talked about because:
- I was having trouble getting `proper-lockfile` to work as expected
- The package is CJS only, which as we know can cause complications depending on runtime

It works well so far but I'll keep testing it.

john-royal:
I think I implemented everything we talked about... but honestly I'm a little iffy on whether I got the CLI part correct. If we don't love it, maybe we can hash out the details during standup today so we can finally get this merged

alchemy/bin/commands/configure.ts
```diff
@@ -0,0 +1,194 @@
...
+      ) {
+        throw new CancelSignal();
+      }
+    }
+    const method = await select({
+      message: `Select a login method for ${pc.bold("Cloudflare")}`,
+      options: [
+        { label: "OAuth", value: "oauth", hint: "Recommended" },
+        { label: "API Token", value: "api-token" },
+        { label: "API Key", value: "api-key", hint: "Legacy" },
```
sam-goodwin: Should we update this to `Global API Key`, I think a lot of people will get confused by API Token vs Key

alchemy/src/cloudflare/api.ts
```diff
@@ -42,56 +51,109 @@ export interface CloudflareApiOptions {
...
-    process.env.CLOUDFLARE_ACCOUNT_ID ??
-    process.env.CF_ACCOUNT_ID ??
-    (await getCloudflareAccountId(authOptions));
-  return (cloudflareApiCache[cacheKey] = new CloudflareApi({
-    baseUrl: options.baseUrl,
-    accountId,
-    authOptions,
-  }));
-}
+    if (options.profile) {
```
sam-goodwin: I think we mentioned `API Key > API Token > Explicit Profile > Default Profile`

alchemy/src/auth.ts
```diff
@@ -0,0 +1,326 @@
...
+  /**
+   * Acquires a lock on a resource.
+   *
+   * WARNING: This function uses async io, so it is not thread-safe within a single process.
+   * Use a `singleFlight` or mutex to ensure thread-safety.
+   *
+   * @param name The name of the resource to lock.
+   * @returns True if the lock was acquired, false if it was already held by another process.
+   */
+  export const acquire = async (name: string): Promise<boolean> => {
```
sam-goodwin: Why don't we lock internally?

sam-goodwin: I am having a hard time finding the lock implementation?

I have used https://www.npmjs.com/package/proper-lockfile and it's pretty useful if you need multi-process locking. 

Do we need to worry about turborepos trying to access profiles at the same time?

john-royal:
- [ ] Cloudflare auth guide
- [ ] CLI commands: configure, login, logout
- [ ] Guide for profiles?

alchemy/src/auth.ts
john-royal: I tried using `proper-lockfile` but:
- I had trouble getting it to work
- I noticed that it's CJS only — other CJS-only packages have caused issues for us

To be clear, this does work across processes. But I'll clarify the implementation and comments.

Brief explanation: The mechanism works *across* processes but because it uses async IO, it has to be used in tandem with `singleFlight` so it works within the same process.

I'll make the implementation more succint.

john-royal: Here's how it's used:

https://github.com/sam-goodwin/alchemy/pull/956/files#diff-3f8f32dccc166808066dbb4b172417239d789686509c3a7497327b359691567fR194-R208

Here's how it's implemented:

https://github.com/sam-goodwin/alchemy/blob/72fbc41cbd87cf19657863d6c5b759a602e485c7/alchemy/src/util/lock.ts

Happy to share my test scripts for it - I would've checked them in but frankly wasn't sure where to put them because it involved transpiling and spawning to verify it works in Node.

alchemy/src/cloudflare/api.ts
john-royal: Made that update but I feel like `options.profile` should probably take precedence over an API key or token in your environment? I don't have too strong an option on this though.

PR #959, opened 2025-09-12
feat(cloudflare): tunnel

john-royal:


sam-goodwin:
Docs are missing

alchemy/src/cloudflare/miniflare/miniflare-controller.ts
```diff
@@ -40,21 +45,32 @@ export class MiniflareController {
...
+        api: input.api,
+        name: input.name,
+      });
+    } else {
+      const proxy = await createMiniflareWorkerProxy({
+        port: input.port ?? (await findOpenPort()),
+        getWorkerName: () => input.name,
+        miniflare,
+        mode: "local",
+      });
```
sam-goodwin, line 61: Curious, why don't we always create the miniflare worker proxy?

alchemy/src/cloudflare/miniflare/miniflare-worker-proxy.ts
```diff
@@ -1,126 +1,108 @@
...
-interface MiniflareWorkerProxyOptions {
-  name: string;
-  port: number;
-  miniflare: miniflare.Miniflare;
+export interface MiniflareWorkerProxy {
+  url: URL;
+  close: () => Promise<void>;
 }
 
-export class MiniflareWorkerProxy {
```
sam-goodwin, line 15: I swear you told me, but what was the reason for this major change? Are we sure it works well (nervous to break dev mode)

alchemy/src/cloudflare/miniflare/miniflare-controller.ts
sam-goodwin: Nevermind, I see it also creates it inside createTunnel

sam-goodwin (approved):
I added the docs for this change. Merge once you are ready.

alchemy/src/cloudflare/miniflare/miniflare-worker-proxy.ts
john-royal: We essentially need two versions of the Miniflare worker proxy:
- For tunnel: maps to one or more workers based on request headers (hence `getWorkerName` prop); must rewrite host header to localhost otherwise Miniflare rejects the connection (hence `transformRequest` prop); stack traces are logged locally but not exposed over HTTP for security.
- For localhost: maps to one worker, no rewrites, stack traces are fine.

However aside from those two differences, the actual request handling logic is identical. I tried accommodating these differences without modifying the original class-based structure and... the code made no sense.

It looks like a bigger change than it actually is — the logic is the same, just with different structure. Works just as well in my testing both automated and manual.

PR #961, opened 2025-09-14
Fix workspace root detection in nested monorepos

ghostwriternr:
## Summary

Fixes an issue with workspace root detection that causes apps to hang when using the new `--app` flag introduced in #953.

## Context

PR #953 added support for multi-app, interconnected monorepos with the `--app` flag. However, when apps have their own `turbo.json` files (common in Turborepo), the workspace root detection fails, causing module resolution to hang.

## The Problem

When using `--app frontend`, the `findWorkspaceRoot` function:
1. Starts searching from the app directory
2. Finds `turbo.json` in `apps/frontend/` 
3. **Incorrectly stops there** instead of continuing to the monorepo root
4. Passes wrong `--root-dir` to the alchemy process
5. Module imports like `import { backend } from "backend/alchemy"` fail to resolve
6. Process hangs indefinitely

This breaks the interconnected app pattern.

## The Solution

This PR fixes `findWorkspaceRoot` to properly handle nested monorepos by:
- Prioritizing authoritative markers (`.git`, `package.json` with workspaces)
- Treating tool configs (`turbo.json`, `nx.json`) as fallbacks only
- Continuing to search upward even after finding tool configs

## Reproduction

I've created a minimal reproduction with testing steps here: https://github.com/ghostwriternr/alchemy-root-detection-repro

## Impact

This fix enables the monorepo support to work with Turborepo projects where apps have their own `turbo.json` files extending the root config.

## AI usage
Claude Code was used to create this PR. Unfortunately don't have the prompts because I spun the conversation off from a session on my personal project, starting with debugging the issue.

### DIFF
--- a/alchemy/bin/services/find-workspace-root.ts
+++ b/alchemy/bin/services/find-workspace-root.ts
@@ -1,48 +1,105 @@
 import fs from "node:fs/promises";
 import path from "node:path";
 
-export async function findWorkspaceRoot(dir: string = process.cwd()) {
-  if ((await fs.stat(dir)).isDirectory()) {
-    if (await exists(dir, ".git")) {
-      // the root of the git repo is usually the workspace root and we should always stop here
-      return dir;
-    } else if ((await read(dir, "package.json"))?.workspaces) {
-      // package.json with workspaces (bun, npm, etc.)
-      return dir;
-    } else if (await anyExists(dir, ...rootFiles)) {
-      return dir;
-    }
-  }
-  return findWorkspaceRoot(path.resolve(dir, ".."));
-}
-
-const read = (...p: string[]): Promise<any> =>
-  fs
-    .readFile(path.join(...p), "utf8")
-    .then(JSON.parse)
-    .catch(() => undefined);
-
-const exists = (...p: string[]) =>
-  fs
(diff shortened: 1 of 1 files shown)


sam-goodwin:
This change is bigger than expected. Should we just not look for `turbo.json` files and instead rely on finding the package workspace vis `.git` `package.json#workspaces` or `pnpm-workspace.yaml`. Probably enough

sam-goodwin:
Does https://github.com/sam-goodwin/alchemy/pull/965 work for your use-case?

You can try by installing `https://pkg.pr.new/sam-goodwin/alchemy@965` (see the `pkg.pr.new` comment on the PR [here](https://github.com/sam-goodwin/alchemy/pull/965#issuecomment-3290798873))

ghostwriternr:
> This change is bigger than expected. Should we just not look for `turbo.json` files and instead rely on finding the package workspace vis `.git` `package.json#workspaces` or `pnpm-workspace.yaml`. Probably enough

I agree, that should work too! In my PR I kept those as fallbacks, but it does feel cleaner to skip the other files.

> Does https://github.com/sam-goodwin/alchemy/pull/965 work for your use-case?

Just tried it on my repo, and it does!

PR #963, opened 2025-09-14
fix(cloudflare): remove ai gateway binding type

### LINKED ISSUE #962: AI Gateway resource and documentation fix
A user had a problem with AI bindings and I traced it back to this: The documentation and typing for our `AiGateway` resource appears to be incorrect.

It seems you can't actually bind to an AI gateway directly, so our types and docs are incorrect. [Cloudflare's documentation](https://developers.cloudflare.com/workers/configuration/multipart-upload-metadata/#bindings) makes no mention of this binding type, and our handling of it is inconsistent: 

https://github.com/sam-goodwin/alchemy/blob/3aac9e0c5621d20aeaa4d50bfd288a8c945199d0/alchemy/src/cloudflare/worker-metadata.ts#L494-L499

https://github.com/sam-goodwin/alchemy/blob/3aac9e0c5621d20aeaa4d50bfd288a8c945199d0/alchemy/src/cloudflare/miniflare/build-worker-options.ts#L100-L108

https://github.com/sam-goodwin/alchemy/blob/3aac9e0c5621d20aeaa4d50bfd288a8c945199d0/alchemy/src/cloudflare/wrangler.json.ts#L762-L764

The correct way to access an AI gateway is to use a standard `Ai` binding, which lets you pass in a gateway ID:

```ts
const response = await env.AI.run(
  "@cf/meta/llama-3.1-8b-instruct-fast",
  {
    prompt: "What is the origin of the phrase Hello, World",
  },
  {
    gateway: {
      id: env.GATEWAY_ID,
    },
  },
);
```

You can also use `env.AI.gateway("<id>")` (this returns the `AiGateway` type that we incorrectly treated as a binding), but [Cloudflare's docs](https://developers.cloudflare.com/ai-gateway/integrations/aig-workers-ai-binding/#3-run-an-inference-task-containing-ai-gateway-in-your-worker) recommend using `env.AI.run()`, and I can see why — the `AiGateway` type is a little unintuitive and seems to be an internal thing.

john-royal:
We had a hallucinated binding type for our `AiGateway` resource. This removes the mistaken type, updates the docs to explain how to use it correctly, and adds some checks to ensure you don't mistakenly try to add more than one AI binding (this seems to be unsupported).

Tested this manually but should I update `ai-gateway.test.ts` too?

Closes #962.

### DIFF
--- a/alchemy/src/cloudflare/worker-metadata.ts
+++ b/alchemy/src/cloudflare/worker-metadata.ts
@@ -491,12 +491,6 @@ export async function prepareWorkerMetadata(
         name: bindingName,
         index_name: binding.name,
       });
-    } else if (binding.type === "ai_gateway") {
-      // AI Gateway binding - just needs the name property
-      meta.bindings.push({
-        type: "ai",
-        name: bindingName,
-      });
     } else if (binding.type === "hyperdrive") {
       // Hyperdrive binding
       meta.bindings.push({
@@ -510,6 +504,12 @@ export async function prepareWorkerMetadata(
         name: bindingName,
       });
     } else if (binding.type === "ai") {
+      const existing = meta.bindings.find((b) => b.type === "ai");
+      if (existing) {
+        throw new Error(
+          `Workers cannot have multiple AI bindings. Binding "${bindingName}" conflicts with "${existing.name}".`,
+        );
+      }
       meta.bindings.push({
         type: "ai",
         name: bindingName,
--- a/alchemy-web/src/content/docs/providers/cloudflare/ai-gateway.md
+++ b/alchemy-web/src/content/docs/providers/cloudflare/ai-gateway.md
@@ -95,15 +95,19 @@ await Worker("chat-worker", {
 // src/worker.ts
 export default {
   async fetch(request, env) {
-    const { message } = await request.json();
+    const { prompt } = await request.json();
 
-    const response = await env.AI.run("@cf/meta/llama-3.1-8b-instruct", {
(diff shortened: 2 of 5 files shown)


alchemy/src/cloudflare/worker-metadata.ts
```diff
@@ -510,6 +504,12 @@ export async function prepareWorkerMetadata(
         name: bindingName,
       });
     } else if (binding.type === "ai") {
+      const existing = meta.bindings.find((b) => b.type === "ai");
```
sam-goodwin, line 507: Won't this find itself? Do you need to filter by key or just count?

john-royal: Not sure what exactly you mean here?

The reason I added this check is because the user who encountered this issue yesterday hit an interesting edge case: if more than one AI binding was defined, Cloudflare didn't throw an error — it just silently made one of the bindings undefined.

The reason we're grabbing the object here instead of just counting is so that way we can throw an error that says `Workers cannot have multiple AI bindings. Binding "${bindingName}" conflicts with "${existing.name}".`, which is probably more helpful than just `More than one AI binding detected` or something like that.

Let me know if this answers your question or if you have a different approach in mind!

sam-goodwin: Oh I got confused. I now see that it is checking if the `meta.bindings` contains an existing AI binding before adding to it. I thought `meta.bindings` was the input, not the output metadata.

PR #969, opened 2025-09-16
feat(cloudflare): service binding named entrypoints

Mkassabov:


alchemy/src/cloudflare/bindings.ts
```diff
@@ -68,17 +68,25 @@ export type Binding =
   | SecretKey
   | string
   | VectorizeIndex
-  | Worker
-  | WorkerStub
-  | WorkerRef
+  | (Worker & { __entrypoint__?: string })
+  | (WorkerStub & { __entrypoint__?: string })
+  | (WorkerRef & { __entrypoint__?: string })
```
sam-goodwin: Can we instead add a new type `WorkerEntrypoint` or something like that?

alchemy/src/cloudflare/bindings.ts
```diff
@@ -68,17 +68,25 @@ export type Binding =
...
-export const Self = Symbol.for("Self");
+export type Self<
+  RPC extends Rpc.WorkerEntrypointBranded = Rpc.WorkerEntrypointBranded,
+> = {
+  type: "_alchemy_self";
+  __entrypoint__?: string;
+  __rpc__?: RPC;
+};
+export const Self = {
+  type: "_alchemy_self",
```
sam-goodwin: let's use the same naming convention we use in resources? Something like `cloudflare::Worker::Self`?

PR #974, opened 2025-09-17
feat(cloudflare): add head, get, list, put and delete methods to R2Bucket

sam-goodwin:
```ts

const bucket = await R2Bucket(..)

const obj = await bucket.get("key");
await bucket.put("key", value)
await bucket.delete("key")
```


focux:
can we add `bucket.head` to this PR? 👀

sam-goodwin:
> can we add `bucket.head` to this PR? 👀

Also added `list`

PR #979, opened 2025-09-19
fix(cloudflare): local workers fail to initialize concurrently

john-royal:
## Changes

- [x] In the `Worker` resource, change the `MiniflareController` import from a dynamic import to a standard one.
- [x] In `createMiniflareWorkerProxy`, retry `server.listen()` on the next port if an `EADDRINUSE` error is encountered.
- [ ] Smoke tests are getting stuck... will verify no regressions before marking ready for review.

## Background

If you initialize multiple `Worker` resources concurrently in dev mode, this may create a race condition that prevents the dev server from starting. This seems to be because the `MiniflareController` is imported dynamically, which prevents the mutex from working as expected. In Node, this shows as an `EADDRINUSE` error; in Bun, it shows as something more cryptic:

```zsh
869 |       let url: string | undefined;
870 |       if (props.dev?.url) {
871 |         url = props.dev.url;
872 |       } else {
873 |         const { MiniflareController } = await import(
874 |           "./miniflare/miniflare-controller.js"
                ^
ReferenceError: Cannot access 'MiniflareController' before initialization.
      at /Users/johnroyal/Developer/Alchemy/alchemy/alchemy/src/cloudflare/worker.ts:874:11
      at processTicksAndRejections (native:7:39)

Bun v1.2.22 (macOS arm64)
```

This is exacerbated when the port is hard-coded, as there is no fallback mechanism for binding to the next available port.

### Reproduction

```ts title="alchemy.run.ts"
// bun run alchemy.run.ts --dev

const app = await alchemy("concurrent-worker-test");

const workers = await Promise.all(
  Array.from({ length: 10 }).map((_, i) =>
    Worker(`worker-${i}`, {
      name: `worker-${i}`,
      script: `export default {
    async fetch(request, env, ctx) {
      return new Response("Hello world!");
    }
  }`,
    }),
  ),
);

console.log(workers.map((worker) => worker.url));

await app.finalize();

```

### DIFF
--- a/alchemy/src/cloudflare/miniflare/miniflare-worker-proxy.ts
+++ b/alchemy/src/cloudflare/miniflare/miniflare-worker-proxy.ts
@@ -1,8 +1,10 @@
 import * as miniflare from "miniflare";
 import { once } from "node:events";
 import http from "node:http";
+import type { AddressInfo } from "node:net";
 import { Readable } from "node:stream";
 import { WebSocket, WebSocketServer } from "ws";
+import { logger } from "../../util/logger.ts";
 
 export interface MiniflareWorkerProxy {
   url: URL;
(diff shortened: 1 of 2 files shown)


JacobMGEvans:
Seemed to work locally for me

john-royal:
Replacing with #1132

PR #980, opened 2025-09-19
fix(cloudflare): prevent websocket connections in remote binding proxy

john-royal:


alchemy/src/cloudflare/miniflare/remote-binding-proxy.ts
```diff
@@ -60,10 +60,19 @@ export async function createRemoteProxyWorker(input: {
...
+      });
+      req.headers.forEach((value, key) => {
+        // `mf-` headers (e.g. `mf-binding`) indicate which binding to use.
+        // Including other headers may cause unexpected behavior.
+        // TODO(john): On some runtimes, it seems to be trying to create a websocket connection.
+        // We should figure out why and handle it, but simply ignoring them seems to work for now.
+        if (key.match(/^mf-/i)) {
+          headers.append(key, value);
+        }
+      });
```
sam-goodwin, line 75: Not sure I follow. Why would stripping headers be a good thing?

john-royal: Apologies- I linked to the Discord messages but should've summarized here. A user was having a problem with a remote binding because of a recent Miniflare update.

In essence, it was trying to make a websocket request through here to the remote binding proxy. This is part of a new remote binding protocol that we don't support yet. Been wanting to add support for it for a week or two, but haven't had a chance and the JSRPC implementation is unstable.

So, right now we only support the "v1" protocol. The only relevant request headers here are the ones that start with `mf-`. Any headers required for remote bindings have that prefix. The others were causing errors. When we only allow the "v1" protocol, Miniflare falls back gracefully.

This did fix the problem for that user, but happy to try a different approach if you'd prefer.

sam-goodwin: Can you explain the architecture? 

Local worker is trying to execute a remote binding. That goes through a proxy to a remote worker that then executes the binding and  returns the data?

That makes sense now why we would strip headers.

sam-goodwin: Can we add a test to prevent regressions?

john-royal: You're absolutely right.™️ That's how it works.

Sure thing, I'll add some tests.

john-royal:
This is a workaround. #990 is the real solution, so closing and replacing with that.

PR #988, opened 2025-09-22
docs: fix the example Durable Object in DO guide

BlankParticle:
The Durable Object code was wrong, no `super` call, missing types
fixed the example code along with better example counter using the sync KV apis


Mkassabov (approved):
Lgtm. Thank you!

JacobMGEvans:
Omg so fast!

PR #989, opened 2025-09-22
feat(clickhouse): Clickhouse Service resource

Mkassabov:
working example with a cf worker reading and writing to a clickhouse service here: https://github.com/Mkassabov/playground/tree/clickhouse

^ example is also in the docs

### DIFF
--- a/alchemy/src/aws/utils.ts
+++ b/alchemy/src/aws/utils.ts
@@ -1,12 +1,13 @@
 import { fromNodeProviderChain } from "@aws-sdk/credential-providers";
 import { loadConfig } from "@smithy/node-config-provider";
+import type { Provider } from "@smithy/types";
 import { AwsClient } from "aws4fetch";
 import { flattenParams } from "../util/params.ts";
 
 /**
  * Get AWS region from configuration
  */
-export const getRegion = loadConfig({
+export const getRegion: Provider<string> = loadConfig({
   environmentVariableSelector: (env: any) =>
     env.AWS_REGION || env.AWS_DEFAULT_REGION,
   configFileSelector: (profile: any) => profile.region,


alchemy/src/clickhouse/api-key.ts
```diff
@@ -0,0 +1,130 @@
...
+      return this({
+        ...this.output,
+        ...updatedKey,
+        name,
+      });
+    }
+
+    const key = await api.v1.organizations(organizationId).keys.post({
+      name,
+      //todo(pr): this is the default for "never expires" that clickhouses uses in their dashboard. do we want to keep it?
```
Mkassabov: This is what clickhouse does with their api; they just set expiration date to 2999. idk if we want to do the same thing; I also don't know if this is a sane default; especially since we can just have it be like N days after deployment and just update it every deployment

alchemy/src/clickhouse/organization.ts
```diff
@@ -0,0 +1,31 @@
...
+ * @param options The api credentials to use.
+ * @returns The organization.
+ */
+export async function getOrganizationByName(
+  name: string,
+  options?: {
+    keyId?: string | Secret<string>;
+    secret?: string | Secret<string>;
+  },
+) {
```
Mkassabov: do we want this to be an organization resource? because technically you can't create or update it so it would be adopt only.

kind of similar to Zones with cloudflare (except you can make a zone and not have NS servers for it)

alchemy/src/clickhouse/service.ts
```diff
@@ -0,0 +1,433 @@
...
+   * @default 356
+   */
+  maxReplicaMemoryGb?: ApiService["maxReplicaMemoryGb"];
+
+  /**
+   * The number of replicas to create the service with.
+   *
+   * @default 3
+   */
+  numReplicas?: ApiService["numReplicas"];
```
Mkassabov, line 68: these are the defaults that clickhouse uses, idk if we want to have different defaults because honestly these seem a little easy to make costs explode out of control

alchemy/src/clickhouse/service.ts
```diff
@@ -0,0 +1,433 @@
...
+  encryptionKey?: ApiService["encryptionKey"];
+  encryptionAssumedRoleIdentifier?: ApiService["encryptionAssumedRoleIdentifier"];
+  releaseChannel: ApiService["releaseChannel"];
+  byocId?: ApiService["byocId"];
+  hasTransparentDataEncryption: ApiService["hasTransparentDataEncryption"];
+  profile: ApiService["profile"];
+  complianceType?: ApiService["complianceType"];
+  backupId?: string;
+  //todo(pr): do we want to split this out into separate properties so we dont have to use find to access?
+  endpoints: ApiService["endpoints"];
```
Mkassabov: the type right now is this:
```ts
  endpoints: [
    {
      protocol: "nativesecure",
      host: "bxbjyogpoo.us-east-1.aws.clickhouse.cloud",
      port: 9440,
    }, {
      protocol: "https",
      host: "bxbjyogpoo.us-east-1.aws.clickhouse.cloud",
      port: 8443,
    }, {
      protocol: "mysql",
      host: "bxbjyogpoo.us-east-1.aws.clickhouse.cloud",
      port: 3306,
      username: "mysql4bxbjyogpoo",
    }
  ],
```

which feels a little gross since if you want to access it you would need to do
```ts
const serviceEndpoint = service.endpoints.find(
	(endpoint) => endpoint.protocol === "https",
)!;
```

and error if it doesn't exist. which IMO isn't great

alchemy/src/clickhouse/service.ts
```diff
@@ -0,0 +1,431 @@
...
+      await checkState();
+      return;
+    } catch (error) {
+      if (attempt === maxRetries - 1) {
+        throw error;
+      }
+      await new Promise((resolve) => setTimeout(resolve, 5000));
+    }
+  }
+}
```
Mkassabov, line 668: I hate this. For deletion the service state needs to be in one of `['provisioning','starting','awaking','idle','stopped','degraded','failed']`.

The only thing we can do is stop the service, wait for it to be stopped on clickhouse's side, then kill it.

This is the exact same thing their terraform provider does, but it will slow deletion down A LOT (because this is not a fast process).

the other option is we just shove it in `pendingDeletions` like replace works and hope it gets taken care of in the future

Mkassabov:
Also not sure what we want to do RE migrations; there don't seem to be any good ORMs that offer typesafety.

we could pull in `https://www.npmjs.com/package/clickhouse-migrations` which is one they recommend [here](https://clickhouse.com/docs/knowledgebase/schema_migration_tools). there is also waddler from the drizzle team: https://waddler.drizzle.team/docs/clickhouse/overview

but I haven't had a chance to look into that too much

alchemy/src/clickhouse/api.ts
```diff
@@ -0,0 +1,916 @@
...
+async function errorHandler<T>(res: Response) {
+  if (res.status !== 200) {
+    const errorMessage = await res.json<{ error: string }>();
+    throw new Error(`Clickhouse API error: ${errorMessage.error}`);
+  }
+  const response = await res.json<{ result: T }>();
+  return response.result;
+}
+
+export function createClickhouseApi(options?: ClickhouseApiOptions) {
```
sam-goodwin: Do they have a SDK we could use in typescript? I've been regretting hand-rolling them all with fetch, hopefully we can avoid doing that?

https://clickhouse.com/docs/integrations/javascript

alchemy/src/clickhouse/organization.ts
sam-goodwin: If you can't create or update it, then maybe not? Could expose a function+type that looks like an organization but all it does it look it up?

```ts
export type Organization = { .. }

export async function Organization(..)
```

alchemy/src/clickhouse/service.ts
sam-goodwin: Can there be more than one per protocol? Maybe a key per protocol would be better?

alchemy/src/clickhouse/service.ts
sam-goodwin: > the other option is we just shove it in pendingDeletions like replace works and hope it gets taken care of in the future

If a resource is slow to delete, there's not much we can do. Going to have to just poll I think. Shame it's so slow.

alchemy/src/clickhouse/organization.ts
Mkassabov: That's what the current getOrgByName function does. Happy to rename it just to `Organization` if it would fit better

alchemy/src/clickhouse/service.ts
Mkassabov: I don't think so? Will try and find out more; but that's why I'm leaning towards keys as well

alchemy/src/clickhouse/organization.ts
sam-goodwin: That or OrganizationRef

sam-goodwin: Our convention is noun based, not verb because it's declarative.

alchemy/src/clickhouse/api.ts
Mkassabov: no their sdk is for interfacing with clickhouse instances (querying and inserting data), not interfacing with their API

alchemy/src/clickhouse/api-key.ts
```diff
@@ -0,0 +1,127 @@
+import type { Context } from "../context.ts";
+import { Resource } from "../resource.ts";
+import { secret, type Secret } from "../secret.ts";
```
Mkassabov, line 3: Test

alchemy/src/clickhouse/organization.ts
sam-goodwin: Are we going to rename this to `OrganizationRef`?

alchemy/src/clickhouse/service.ts
```diff
@@ -0,0 +1,548 @@
...
+   * @default 8
+   */
+  minReplicaMemoryGb?: ApiService["minReplicaMemoryGb"];
+
+  /**
+   * The maximum replica memory to create the service with.
+   *
+   * @default 356
+   */
+  maxReplicaMemoryGb?: ApiService["maxReplicaMemoryGb"];
```
sam-goodwin, line 61: Should we just be doing `extends Omit<ApIService, ..>` instead of duplicating with structural references?

alchemy/src/clickhouse/service.ts
```diff
@@ -0,0 +1,548 @@
...
+  provider: NonNullable<ApiService["provider"]>;
+  region: NonNullable<ApiService["region"]>;
+  ipAccessList: NonNullable<ApiService["ipAccessList"]>;
+  minReplicaMemoryGb: NonNullable<ApiService["minReplicaMemoryGb"]>;
+  maxReplicaMemoryGb: NonNullable<ApiService["maxReplicaMemoryGb"]>;
+  numReplicas: NonNullable<ApiService["numReplicas"]>;
+  idleScaling: NonNullable<ApiService["idleScaling"]>;
+  idleTimeoutMinutes: NonNullable<ApiService["idleTimeoutMinutes"]>;
+  isReadonly: NonNullable<ApiService["isReadonly"]>;
+  dataWarehouseId: NonNullable<ApiService["dataWarehouseId"]>;
```
sam-goodwin, line 212: These types are complicated. Can we extend `ApiService`?

Also need to document all properties.

alchemy/src/neon/api/types.gen.ts
```diff
@@ -2879,6 +2886,24 @@ export type ListApiKeysResponses = {
...
+   *
+   * The following HTTP methods are considered non-idempotent: `POST`, `PATCH`, `DELETE`, and `PUT`. Retrying these methods is generally **not safe**.
+   * The following methods are considered idempotent: `GET`, `HEAD`, and `OPTIONS`. Retrying these methods is **safe** in the event of a network error or timeout.
+   *
+   * Any request that returns a `503 Service Unavailable` response is always safe to retry.
+   *
+   * Any request that returns a `423 Locked` response is safe to retry. `423 Locked` indicates that the resource is temporarily locked, for example, due to another operation in progress.
+   *
+   */
+  default: GeneralError;
```
sam-goodwin: Is this a bug? Does our code handle `default` status code?

alchemy/src/neon/api/types.gen.ts
```diff
@@ -3082,6 +3161,24 @@ export type ListProjectsResponses = {
...
```
sam-goodwin: This looks awkward :)

alchemy/src/clickhouse/service.ts
Mkassabov: there can't so we moved to keys

PR #991, opened 2025-09-22
fix(cloudflare): avoid storing assets in state

john-royal:
Turns assets from a resource into a function so file names aren't stored in state. This should fix the problem with our website deployments.

```diff
- const assets = await Assets("versioned-static-assets", {
+ const assets = Assets({
  path: tempDir,
});
```

A no-op `cloudflare::Asset` resource is included, similar to what was done with the `WranglerJson` resource, so the old resource is removed from state as expected.

As an added bonus, the analysis is done in one pass instead of two.


sam-goodwin:
Should it still compute a hash of all the files? Could even just hash the file metadata and not the data as an optimization. We alwaysUpdate so there's no risk of breaking workers as it is today but does allow it to still be used to trigger state changes later?

john-royal:
> Should it still compute a hash of all the files? Could even just hash the file metadata and not the data as an optimization. We alwaysUpdate so there's no risk of breaking workers as it is today but does allow it to still be used to trigger state changes later?

I see what you're going for but is it worth it? The `Worker` resource would have to read the files and compute the hashes regardless since it no longer has the state to read that from.

Happy to make the change if you'd like but wondering if the added I/O operations would be worth it.

sam-goodwin:
I was thinking of a metadata-only hash (doesn't `read` file, only `stat`). Not sure if that's worth it either though ... It just makes the whole `Assets` resource pretty useless. 

I guess it's just a type-safe box for the path and a `Bound<T>` mapping. 

Maybe just keep it as an `async` function so that it's not a breaking change in that respect and we can always change our mind again without changing the function's color.

john-royal:
Done, do you want to keep the ID as well so it's not a breaking change at all?

sam-goodwin:
> Done, do you want to keep the ID as well so it's not a breaking change at all?

Probably safe to assume it'll never be a resource again since it has no backing API outside of being bound to a worker?

alchemy/src/cloudflare/website.ts
```diff
@@ -342,13 +342,9 @@ export async function Website<B extends Bindings>(
...
-      ...(!scope.local
-        ? {
-            ASSETS: await Assets("assets", {
-              path: path.relative(process.cwd(), paths.assets),
-            }),
-          }
-        : {}),
+      ASSETS: await Assets({
+        path: paths.assets,
+      }),
```
sam-goodwin: This is a logic change, is there a reason why we weren't adding them for `scope.local` before?

PR #992, opened 2025-09-22
fix(cloudflare): minify internal workers

john-royal:


alchemy/scripts/build-workers.ts
```diff
@@ -13,6 +13,8 @@ await esbuild.build({
     .map((worker) => path.join(WORKERS_DIR, worker)),
   outdir: WORKERS_DIR,
   bundle: true,
+  minify: true,
+  keepNames: true,
```
sam-goodwin: Is this necessary?

john-royal: I thought it was for durable objects and I thought that was why we use that same option for bundling user workers... but I removed it and it still seems to work. Makes sense since the exports are still the same.

sam-goodwin: yeah, I think esbuild will make sure to not rename exports.

PR #995, opened 2025-09-23
feat(neon): branch resource & rewrite project

john-royal:


alchemy/src/neon/project.ts
```diff
@@ -43,400 +49,20 @@ export interface NeonProjectProps extends NeonApiOptions {
    * PostgreSQL version to use
    * @default 16
    */
-  pg_version?: 14 | 15 | 16 | 17;
-
-  /**
-   * Whether to create a default branch and endpoint
-   * @default true
-   */
-  default_endpoint?: boolean;
```
john-royal, line 52: This prop isn't part of the spec and has no effect

alchemy/src/neon/project.ts
```diff
@@ -43,400 +49,20 @@ export interface NeonProjectProps extends NeonApiOptions {
...
    * @default "main"
    */
   default_branch_name?: string;
-
-  /**
-   * Existing project ID to update
-   * Used internally during update operations
-   * @internal
-   */
-  existing_project_id?: string;
```
john-royal, line 65: Do we want to keep this? Implementation is way simpler without, but I can see why it might be useful

alchemy/src/neon/project.ts
```diff
@@ -519,367 +155,100 @@ export const NeonProject = Resource(
...
+  ) {
     const api = createNeonApi(props);
-    const projectId = props.existing_project_id || this.output?.id;
-    const projectName =
+    const name =
       props.name ?? this.output?.name ?? this.scope.createPhysicalName(id);
 
-    if (this.phase === "update" && this.output.name !== projectName) {
-      this.replace();
-    }
```
john-royal, line 529: Neon projects can be renamed, so this wasn't necessary

PR #996, opened 2025-09-23
chore: improve api client generation

john-royal:
Uses the default `@hey-api/sdk` client generator instead of a custom one so it's trivial to add new APIs.


scripts/clients/generate.ts
```diff
@@ -0,0 +1,65 @@
+import {
+  createClient,
+  defineConfig,
+  type UserConfig,
+} from "@hey-api/openapi-ts";
+import path from "node:path";
+
+export const clients = ["planetscale"] as const;
```
john-royal, line 8: @Mkassabov If you want to try this out for Clickhouse, add `"clickhouse"` to this array and run `bun run scripts/clients/generate.ts`

scripts/clients/clickhouse.ts
```diff
@@ -0,0 +1,55 @@
...
+  (res) => res.json() as Promise<OpenApi.V3_0_X>,
+);
+
+const createOperationId = (summary: string) => {
+  return toCamelCase(
+    summary
+      .replace(/(\b(?:a|from|to|the|given|of|an)\b)/gi, "")
+      .replace("Get list", "list"),
+  );
+};
```
john-royal, line 18: This looks dumb because it is but mostly good results (e.g. listAvailableOrganizations, getOrganizationDetails, updateOrganizationDetails, getOrganizationMetrics, listOrganizationServices, createNewService, getServiceDetails, updateServiceBasicDetails, deleteService)

scripts/clients/clickhouse.ts
```diff
@@ -0,0 +1,55 @@
...
+export default {
+  input: spec as any,
+  output: {
+    path: "alchemy/src/clickhouse/api",
+    format: "biome",
+  },
+  plugins: [
+    {
+      name: "@hey-api/typescript",
+      exportFromIndex: false,
```
mrlubos, line 42: @john-royal you can also do `output.indexFile: false` to avoid specifying this inside plugins. Apart from that, let me know if you have any feedback on the library!

PR #997, opened 2025-09-24
draft: update tanstack start resource for rc

john-royal:


alchemy/package.json
```diff
@@ -60,6 +60,10 @@
       "bun": "./src/cloudflare/astro/plugin.ts",
       "import": "./lib/cloudflare/astro/plugin.js"
     },
+    "./cloudflare/env": {
+      "bun": "./src/cloudflare/cloudflare-env-proxy.ts",
+      "import": "./lib/cloudflare/cloudflare-env-proxy.js"
+    },
```
john-royal, line 66: Vite was complaining when we imported `getCloudflareEnvProxy` from `alchemy/cloudflare` - the error was coming from some completely unrelated resources that happened to be part of the barrel import. Importing only the env proxy fixes that.

alchemy/src/cloudflare/tanstack-start/plugin.ts
```diff
@@ -9,22 +10,41 @@ import { dedent } from "../../util/dedent.ts";
...
+              "cloudflare:workers is not supported in the client",
+            );
+          }
+          return id;
+        }
+      },
+      load: (id) => {
+        if (id === "cloudflare:workers") {
+          return dedent`
+          import { getCloudflareEnvProxy } from "alchemy/cloudflare/env";
```
john-royal, line 31: See https://github.com/sam-goodwin/alchemy/pull/997/files#r2376895079

alchemy/src/cloudflare/tanstack-start/plugin.ts
```diff
@@ -9,22 +10,41 @@ import { dedent } from "../../util/dedent.ts";
...
+    },
+    {
+      name: "alchemy:workers-build-external",
+      config: () => ({
+        build: {
+          rollupOptions: {
+            external: ["cloudflare:workers"],
+          },
+        },
+      }),
```
john-royal, line 47: This means you don't have to set it as an external module in your `vite.config.ts`.

We might want to include the other unenv externals? Not sure.

examples/cloudflare-tanstack-start/vite.config.ts
```diff
@@ -1,25 +1,18 @@
...
 import { defineConfig } from "vite";
 import tsConfigPaths from "vite-tsconfig-paths";
 
 export default defineConfig({
-  build: {
-    target: "esnext",
-    rollupOptions: {
-      external: ["node:async_hooks", "cloudflare:workers"],
-    },
-  },
```
john-royal, line 13: This is no longer necessary; it's part of our Vite plugin instead.

examples/cloudflare-tanstack-start/package.json
```diff
@@ -11,23 +11,25 @@
     "start": "vite start"
   },
   "dependencies": {
-    "@tanstack/react-router": "^1.128.3",
-    "@tanstack/react-router-devtools": "^1.128.3",
-    "@tanstack/react-start": "^1.128.3",
+    "@tanstack/react-router": "^1.132.2",
+    "@tanstack/react-router-devtools": "^1.132.2",
+    "@tanstack/react-start": "^1.132.2",
+    "@tanstack/router-plugin": "^1.132.2",
```
utopyin, line 17: I believe `@tanstack/router-plugin` is not needed

alchemy/src/cloudflare/tanstack-start/plugin.ts
sam-goodwin: Does the cloudflare plugin not do that? Or are we doing this because TSS is not properly using the CF plugin (because of a regression)

john-royal: TSS never seemed to work with the Cloudflare plugin and the update hasn't changed that but double checking now in case it's changed

john-royal:
Replacing with #1028

PR #999, opened 2025-09-24
WIP: no public Symbols

eastlondoner:
Plan: Remove Symbols keys from public Alchemy Resources

So far:
- Cloudflare Workers ✅ 

Putting up a draft PR to show how I will do this, pls let me know if this seems OK

How does it work?

`ExportedResource<T extends Resource>` is a generic wrapper that strips the Resource Symbols

The `Worker` type is split into 2 types `Worker` (public type without Symbol Keys) and `WorkerResource` (internal type with Symbol keys).

WorkerResource is not exported from `worker.ts`

Possible Concerns:
- Not super robust enforcement that we don't leak types that need Symbol keys, particularly on arguments to other functions. If `WorkerResource` and other types like it can be kept internal (not exported) then this is fairly low risk

Alternatives considered:
- Declaring the alchemy Resource symbols in a separate only-one-version-will-ever-exist package

### DIFF
--- a/alchemy/src/cloudflare/worker.ts
+++ b/alchemy/src/cloudflare/worker.ts
@@ -1,7 +1,7 @@
 import path from "pathe";
 import type { Context } from "../context.ts";
 import type { BundleProps } from "../esbuild/bundle.ts";
-import { Resource, ResourceKind } from "../resource.ts";
+import { type ExportedResource, Resource, ResourceKind } from "../resource.ts";
 import type { type } from "../type.ts";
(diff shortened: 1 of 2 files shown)


sam-goodwin:
We could just remove them from the `Resource<..>` interface type and then update `apply` `destroy` etc. to cast to `InternalResource` that contains the props (or just expose functions like `getResouceKind`

eastlondoner:
Ok I’ll take a look at that approach

eastlondoner:
alternative approach here: https://github.com/sam-goodwin/alchemy/pull/1000/files

PR #1000, opened 2025-09-24
Remove Symbol keys from Resource interface

eastlondoner:
removing the symbol keys from the Resource interface as suggested by Sam

Splits `Resource` into `Resource` and `ResourceInternal`

Concept:
- Public methods should accept `Resource` instances so they don't care which alchemy they come from
- Public methods can return `ResourceInternal` instances that's ok

### DIFF
--- a/alchemy/src/alchemy.ts
+++ b/alchemy/src/alchemy.ts
@@ -10,7 +10,6 @@ import {
   ResourceKind,
   ResourceScope,
   ResourceSeq,
-  type PendingResourceInternal,
   type ResourceInternal,
 } from "./resource.ts";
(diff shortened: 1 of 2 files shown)


eastlondoner:
image

Separate alchemy node modules but everything builds no problem

eastlondoner:
I think this is OK.

I ran a single test on my cloudflare account to sanity check
image

sam-goodwin:
Took this a bit further https://github.com/sam-goodwin/alchemy/pull/1003

sam-goodwin:
Closing in favor of https://github.com/sam-goodwin/alchemy/pull/1003

PR #1008, opened 2025-09-25
chore: prefix unused variables to satisfy biome

utopyin:


alchemy/test/cloudflare/bucket.test.ts
```diff
@@ -563,7 +563,7 @@ async function getObject(
   });
 }
 
-async function headObject(
+async function _headObject(
```
sam-goodwin: If these aren't used, we can delete them.

alchemy/test/cloudflare/bucket.test.ts
```diff
@@ -439,7 +439,7 @@ describe("R2 Bucket Resource", async () => {
     }
   });
 
-  test("bucket operations head, get, put, and delete objects", async (scope) => {
+  test("bucket operations head, get, put, and delete objects", async () => {
```
sam-goodwin: Oops. Actually we need to call scope.finalize() down below in the finally. I forgot to uncomment it while testing.

utopyin: fixed [e903029](https://github.com/sam-goodwin/alchemy/pull/1008/commits/e903029122971752e6e02b3da0a96ceed3594250)

alchemy/test/cloudflare/bucket.test.ts
utopyin: fixed [bc75e2a](https://github.com/sam-goodwin/alchemy/pull/1008/commits/bc75e2a1807b884306aa97b012f0f2f4b897dbb8)

PR #1009, opened 2025-09-25
feat(cloudflare): support data catalog on buckets

Mkassabov:
## What this does
- adds support for data catalog to R2 buckets

### DIFF
--- a/alchemy/src/cloudflare/bucket.ts
+++ b/alchemy/src/cloudflare/bucket.ts
@@ -343,9 +343,21 @@ type _R2Bucket = Omit<BucketProps, "delete" | "dev"> & {
     remote: boolean;
   };
 
+  /**
+   * Data catalog for the bucket
+   */
   catalog?: {
+    /**
+     * ID of the data catalog
+     */
     id: string;
+    /**
+     * Name of the data catalog
+     */
     name: string;
+    /**
+     * Host of the data catalog
+     */
     host: string;
   };
 };


alchemy/test/cloudflare/bucket.test.ts
```diff
@@ -496,6 +496,40 @@ describe("R2 Bucket Resource", async () => {
...
+    try {
+      let bucket = await R2Bucket(bucketName, {
+        name: bucketName,
+        adopt: true,
+        dataCatalog: true,
+      });
+      expect(bucket.catalog).toBeDefined();
+      expect(bucket.catalog?.id).toBeDefined();
+      expect(bucket.catalog?.name).toBeDefined();
+      expect(bucket.catalog?.host).toBeDefined();
```
sam-goodwin, line 511: Anything to test by calling an API?

alchemy/src/cloudflare/bucket.ts
```diff
@@ -337,6 +342,12 @@ type _R2Bucket = Omit<BucketProps, "delete" | "dev"> & {
      */
     remote: boolean;
   };
+
+  catalog?: {
```
sam-goodwin, line 349: Add docs

alchemy/test/cloudflare/bucket.test.ts
Mkassabov: seems excessive. we would need to populate data in it and then attach a query engine, then query it. more can go wrong there than right imo.

we check that we get an id back which means it exists.

alchemy/src/cloudflare/bucket.ts
Mkassabov: done

PR #1010, opened 2025-09-25
chore(cloudflare): update miniflare

john-royal:


alchemy/src/cloudflare/miniflare/build-worker-options.ts
```diff
@@ -74,7 +74,7 @@ export const buildWorkerOptions = async (
       // workerd/io/worker.c++:2164: info: uncaught exception; source = Uncaught (in promise); stack = TypeError: Invalid URL string.
       {
         entrypoint: "default",
-        proxy: false,
+        proxy: true,
```
sam-goodwin, line 77: is this intentional?

PR #1017, opened 2025-09-27
main has been renamed to entrypoint - update docs

eastlondoner:
image
current docs are wrong


artimath:
i thought you renamed your default branch and i got really excited

*chuckles in F=ma* 

ref: https://github.com/artimath/agientic/tree/physics/spec

PR #1021, opened 2025-09-28
feat(cloudflare): Add Tail Consumers

JacobMGEvans:
Support for Tail Consumer configuratioin for Workers. Added integration/E2E related tests

### DIFF
--- a/alchemy/src/cloudflare/bindings.ts
+++ b/alchemy/src/cloudflare/bindings.ts
@@ -134,7 +134,6 @@ export type WorkerBindingSpec =
   | WorkerBindingSecretsStoreSecret
   | WorkerBindingService
   | WorkerBindingStaticContent
-  | WorkerBindingTailConsumer
   | WorkerBindingVectorize
   | WorkerBindingVersionMetadata
   | WorkerBindingWasmModule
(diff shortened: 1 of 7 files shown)


alchemy/src/cloudflare/worker.ts
```diff
@@ -375,6 +375,14 @@ export interface BaseWorkerProps<
...
     cpu_ms?: number;
   };
+
+  /**
+   * Tail consumers that will receive execution logs from this worker
+   */
+  tailConsumers?: Array<{
+    /** Name of the tail worker service */
+    service: string;
+  }>;
```
sam-goodwin: What services are supported? Are they meant to be other workers? If so, then it should be `(Worker | {service: string})[]`

JacobMGEvans: Yes, that is the polishing we are working on right now lol

alchemy/src/cloudflare/bindings.ts
```diff
@@ -420,18 +419,6 @@ export interface WorkerBindingService {
...
- */
-export interface WorkerBindingTailConsumer {
-  /** The name of the binding */
-  name: string;
-  /** Type identifier for Tail Consumer binding */
-  type: "tail_consumer";
-  /** Service name */
-  service: string;
-}
-
```
sam-goodwin, line 434: Why was this removed? This file models the api spec for bindings

JacobMGEvans: I could put it back, but that is an example of what I referring too for instance (my confusion), TailConsumer isn't a binding its a Worker configuration much like Cron Triggers. I was trying to figure out what was representation of Cloudflare API and what was Alchemy, this seemed like it was a saying TailConsumer was a "Binding"

PR #1024, opened 2025-09-29
feat(cloudflare): add httpMetadata prop to bucket put fn

focux:
This PR adds a httpMetadata property which is very useful for R2 file uploads and matches the Cloudflare bucket binding API.

### DIFF
--- a/alchemy/src/cloudflare/bucket.ts
+++ b/alchemy/src/cloudflare/bucket.ts
@@ -489,6 +489,9 @@ export async function R2Bucket(
     put: async (
       key: string,
       value: PutObjectObject,
+      options?: {
+        httpMetadata?: Record<string, string>;
+      },
     ): Promise<PutR2ObjectResponse> => {
       if (isLocal) {
         // @ts-expect-error - node built-ins vs cloudflare built-ins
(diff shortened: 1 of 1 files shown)


sam-goodwin (commented):
Can you add a test to check it works?

focux:
@sam-goodwin sure! will do now

focux:
Ok, added some tests

alchemy/src/cloudflare/bucket.ts
```diff
@@ -449,7 +453,17 @@ export async function R2Bucket(
...
       if (isLocal) {
-        return (await localBucket()).head(key);
+        const result = await (await localBucket()).head(key);
+        if (result) {
+          return {
+            key: result.key,
+            etag: result.etag,
+            uploaded: result.uploaded,
+            size: result.size,
+            contentType: result.httpMetadata?.contentType || null,
```
sam-goodwin: does this match Cloudflare's interface or should we be returning `httpMetadata` too?

focux: Well... it matches Cloudflare's now

sam-goodwin:
Looking good but failing linting checks. Always remember to run `bun check --write` from the root

focux:
Yeah, I tried that command before and I was getting a weird TS issue related to react router, had to run the lint manually `bun biome check --write`.

It should be good to go now

PR #1033, opened 2025-09-30
fix(cloudflare): use miniflare entry for proxy

john-royal:
Instead of fetching from the user worker directly — which bypasses the Miniflare entrypoint and causes all sorts of trouble — this registers a route for each user worker and uses the `MF-Route-Override` header to route requests to the correct worker.

This fixes the problem with websockets that required us to set `proxy: false` in order to reach the user worker, which was breaking service bindings.

### DIFF
--- a/examples/cloudflare-durable-object-websocket/alchemy.run.ts
+++ b/examples/cloudflare-durable-object-websocket/alchemy.run.ts
@@ -34,11 +34,11 @@ export const client = await Vite("client", {
 
 console.log("Client:", client.url);
 
-await app.finalize();
-
 if (process.env.ALCHEMY_E2E) {
   const { test } = await import("./test/e2e.ts");
   await test({
     url: server.url,
   });
 }
+
+await app.finalize();
--- a/examples/cloudflare-durable-object-websocket/test/e2e.ts
+++ b/examples/cloudflare-durable-object-websocket/test/e2e.ts
@@ -6,32 +6,30 @@ export async function test({ url }: { url: string | undefined }) {
   console.log("Running Durable Object WebSocket E2E test...");
 
   assert(url, "URL is not set");
+
   await pollUntilReady(new URL("/status", url));
 
   const ws = new WebSocket(new URL("/websocket", url));
-  try {
-    await once(ws, "open");
+  await once(ws, "open");
 
-    const data = crypto.randomUUID();
-    ws.send(data);
-    const message = await new Promise<string>((resolve) => {
-      ws.addEventListener("message", (event) => {
-        resolve(event.data.toString());
-      });
+  const data = crypto.randomUUID();
+  ws.send(data);
+  const message = await new Promise<string>((resolve) => {
+    ws.addEventListener("message", (event) => {
+      resolve(event.data.toString());
     });
-    console.log("Received message", message);
-    assert.equal(
-      message,
-      `Received message: ${data}`,
-      "Received message is not correct",
-    );
-  } finally {
-    console.log("Durable Object WebSocket E2E test passed");
-    ws.close(1000);
+  });
+  console.log("Received message", message);
+  assert.equal(
+    message,
+    `Received message: ${data}`,
+    "Received message is not correct",
+  );
 
-    // TODO(sam): what the actual fuck, why does WebSocket keep the process alive?
-    process.exit(0);
-  }
+  ws.close(1000);
+  await once(ws, "close");
+
+  console.log("Durable Object WebSocket E2E test passed");
 }
 
 async function pollUntilReady(url: URL) {


examples/cloudflare-durable-object-websocket/test/e2e.ts
john-royal, line 1: Reverted this file because the workaround is no longer necessary

alchemy/src/cloudflare/miniflare/miniflare-worker-proxy.ts
```diff
@@ -45,71 +45,50 @@ export async function createMiniflareWorkerProxy(options: {
...
-      !result.success
-    ) {
-      return new miniflare.Response(null, { status: 404 });
-    }
-
-    if (!result.success) {
-      throw result.error;
-    }
-
-    return result.response;
```
john-royal, line 83: No longer needed - handled internally by Miniflare

alchemy/src/cloudflare/miniflare/build-worker-options.ts
```diff
@@ -70,11 +72,9 @@ export const buildWorkerOptions = async (
     compatibilityFlags: input.compatibilityFlags,
     unsafeDirectSockets: [
       // This matches the Wrangler configuration by exposing the default handler (e.g. `export default { fetch }`).
-      // However, unlike Wrangler, we set `proxy: false` to avoid the following error when connecting via a websocket:
-      // workerd/io/worker.c++:2164: info: uncaught exception; source = Uncaught (in promise); stack = TypeError: Invalid URL string.
       {
         entrypoint: "default",
-        proxy: false,
```
john-royal, line 77: Routing through the Miniflare entry means setting `proxy: true` here is no longer an issue - DO websocket test passes

examples/cloudflare-durable-object-websocket/test/e2e.ts
```diff
@@ -6,32 +6,30 @@ export async function test({ url }: { url: string | undefined }) {
...
-      `Received message: ${data}`,
-      "Received message is not correct",
-    );
-  } finally {
-    console.log("Durable Object WebSocket E2E test passed");
-    ws.close(1000);
-
-    // TODO(sam): what the actual fuck, why does WebSocket keep the process alive?
-    process.exit(0);
-  }
```
john-royal, line 34: Reverted this change - workaround no longer necessary

examples/cloudflare-durable-object-websocket/alchemy.run.ts
```diff
@@ -34,11 +34,11 @@ export const client = await Vite("client", {
...
-await app.finalize();
-
 if (process.env.ALCHEMY_E2E) {
   const { test } = await import("./test/e2e.ts");
   await test({
     url: server.url,
   });
 }
+
+await app.finalize();
```
john-royal, line 44: Calling `app.finalize()` first meant the E2E test wouldn't actually run because the process would exit on finalize

alchemy/src/cloudflare/miniflare/build-worker-options.ts
sam-goodwin: Hooray!

alchemy/src/cloudflare/miniflare/build-worker-options.ts
```diff
@@ -85,6 +85,8 @@ export const buildWorkerOptions = async (
             : "unix:///var/run/docker.sock",
       },
     },
+    // This exposes the worker as a route that can be accessed by setting the MF-Route-Override header.
+    routes: [input.name],
```
sam-goodwin, line 89: clever

alchemy/src/cloudflare/miniflare/miniflare-controller.ts
```diff
@@ -104,6 +104,8 @@ export class MiniflareController {
         // This is required to allow websites and other separate processes
         // to detect Alchemy-managed Durable Objects via the Wrangler dev registry.
         unsafeDevRegistryDurableObjectProxy: true,
+        // This exposes other handlers like `scheduled` and `email` via HTTP.
+        unsafeTriggerHandlers: true,
```
sam-goodwin, line 108: Are we using that somewhere?

alchemy/src/cloudflare/miniflare/miniflare-worker-proxy.ts
```diff
@@ -45,71 +45,50 @@ export async function createMiniflareWorkerProxy(options: {
...
-      url: info.url,
-    };
+    const request = new miniflare.Request(info.url, {
+      method: info.method,
+      headers: info.headers,
+      body: info.body,
+      redirect: info.redirect,
+      duplex: info.duplex,
+    });
+    return await options.miniflare.dispatchFetch(request);
```
sam-goodwin, line 69: cc @eastlondoner

sam-goodwin (approved):

alchemy/src/cloudflare/miniflare/miniflare-controller.ts
john-royal: Yes, see [here](https://github.com/alchemy-run/alchemy/pull/1033/files#diff-4a32ab9ffdfde470f189bb2b2a6e57c7144566110302706d70b718a364d7f272R54-R59). It's how we expose the `/__scheduled` endpoint with a simple URL rewrite instead of custom handling. Also means you can test other handlers (e.g. for email I believe it's `/cdn-cgi/handler/email`).

PR #1038, opened 2025-10-01
docs: serve markdown to llms

john-royal:
- Every docs route has a `.md` equivalent generated at build time (e.g. `/concepts/cli` > `/concepts/cli.md`)
- Rewrite based on `accept` header (tested with Claude Code and opencode)
- Router worker handles rewrites and caches responses
- Uses Cloudflare `_headers` option to add `vary: accept` to relevant routes

PR #1045, opened 2025-10-02
docs: clean up unused resources

john-royal:
- Remove dead code from PostHog proxy
- Remove Static*File resources from `alchemy/fs` docs (resources were removed but docs are still there)


john-royal:
The base branch for this is #1038 so I'll wait for that to be approved before merging this one just to keep the diff minimal

Mkassabov (approved):

PR #1046, opened 2025-10-02
feat(cli): globally disable telemetry

john-royal:
Run `alchemy telemetry <enable/disable>` to set tracking globally on your machine.

When disabled, a file is created at `os.homedir()/.alchemy/telemetry-disabled`.

## Question

Would it be better to merge these into one file, `os.homedir()/.alchemy/telemetry.json`? This is definitely me being neurotic but I don't like how this looks:

image

### DIFF
--- a/alchemy/bin/alchemy.ts
+++ b/alchemy/bin/alchemy.ts
@@ -8,6 +8,7 @@ import { init } from "./commands/init.ts";
 import { login } from "./commands/login.ts";
 import { logout } from "./commands/logout.ts";
 import { run } from "./commands/run.ts";
+import { telemetry } from "./commands/telemetry.ts";
 import { getPackageVersion } from "./services/get-package-version.ts";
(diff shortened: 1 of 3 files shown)


sam-goodwin:
> When disabled, a file is created at os.homedir()/alchemy/telemetry-disabled.

Do you mean .alchemy? It should be a hidden dotfolder

john-royal:
> > When disabled, a file is created at os.homedir()/alchemy/telemetry-disabled.
> 
> Do you mean .alchemy? It should be a hidden dotfolder

Yes my bad

sam-goodwin:
Oh nevermind. I see it in the screenshot. Typo in pr

Mkassabov:
why not just set the env variable for the user to disable this?

also I don't know if we want this to be global vs profile based. I personally think profile based would be better

Mkassabov (approved):

PR #1051, opened 2025-10-02
(fix) Update the expectations in the Bun SPA test so it passes + some cosmetic improvments

eastlondoner:
The test was copypasta from the Vite resource but the BunSPA resource works a little differently.

Had to pass the api url separately.

To make that nicer I updated the resource to output it as a prop.

Added some comments in the app to the part where magic happens and changed to make it a bit more robust.

### DIFF
--- a/alchemy/src/cloudflare/bun-spa/bun-spa.ts
+++ b/alchemy/src/cloudflare/bun-spa/bun-spa.ts
@@ -14,16 +14,13 @@ import {
   Website,
   type WebsiteProps,
 } from "../website.ts";
-import type { Worker } from "../worker.ts";
 
 export interface BunSPAProps<B extends Bindings> extends WebsiteProps<B> {
   frontend: string;
   outDir?: string;
 }
 
-export type BunSPA<B extends Bindings> = B extends { ASSETS: any }
-  ? never
-  : Worker<B & { ASSETS: Assets }>;
+export type BunSPA<B extends Bindings> = Website<B> & { apiUrl: string };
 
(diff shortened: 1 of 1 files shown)


alchemy/src/cloudflare/bun-spa/bun-spa.ts
```diff
@@ -28,7 +28,7 @@ export type BunSPA<B extends Bindings> = B extends { ASSETS: any }
 export async function BunSPA<B extends Bindings>(
   id: string,
   props: BunSPAProps<B>,
-): Promise<BunSPA<B>> {
+): Promise<BunSPA<B> & { apiUrl: string }> {
```
sam-goodwin, line 28: Shouldn't we put this on the BunSPA<B> type?

eastlondoner: It’s only part of the output. It’s not part of the input. In prod it is the same as the url you input. In dev it’s different but in neither case should you set it on the input

eastlondoner: I can put it on BunSPA and exclude it from the inputs if that’s a better api

sam-goodwin: The pattern is


```
type {Resource}Props

type {Resource}

const {Resource}
```

Isn't BunSPA the output type? If not, it should be

eastlondoner: oh facepalm

PR #1052, opened 2025-10-02
chore(cloudflare): add observability wrangler config types

utopyin:
This adds the types for the observability config.

Mostly based on the [wrangler schema](https://unpkg.com/wrangler@4.40.3/config-schema.json) and the [wrangler config docs](https://developers.cloudflare.com/workers/wrangler/configuration/#observability).

Question: should `prepareWorkerMetadata` be updated too? Right now the base meta object is initialized this way:
```ts
const meta: WorkerMetadata = {
  // ...
  observability: {
    enabled: props.observability?.enabled !== false,
  },
  // ...
};
```

### DIFF
--- a/alchemy/src/cloudflare/worker.ts
+++ b/alchemy/src/cloudflare/worker.ts
@@ -166,17 +166,87 @@ export interface BaseWorkerProps<
   url?: boolean;
 
   /**
-   * Observability configuration for the worker
+   * Specify the observability behavior of the Worker.
+   * For reference, see [wrangler configuration](https://developers.cloudflare.com/workers/wrangler/configuration/#observability)
    *
-   * Controls whether worker logs are enabled
-   * @default { enabled: true }
+   * @default { enabled: true, head_sampling_rate: 1 }
    */
   observability?: {
     /**
-     * Whether to enable worker logs
+     * If observability is enabled for this Worker. Defaults to true for all new Workers.
      * @default true
      */
     enabled?: boolean;
+
+    /**
+     * A number between 0 and 1, where 0 indicates zero out of one hundred requests are logged, and 1 indicates every request is logged.
+     * If head_sampling_rate is unspecified, it is configured to a default value of 1 (100%).
+     * Read more about [head-based sampling](https://developers.cloudflare.com/workers/observability/logs/workers-logs/#head-based-sampling).
+     * @default 1
+     */
+    head_sampling_rate?: number;
+
+    /**
+     * Configuration for worker logs
+     */
+    logs?: {
+      /**
+       * Whether logs are enabled
+       * @default true
+       */
+      enabled?: boolean;
+
+      /**
+       * The sampling rate for logs
+       */
+      head_sampling_rate?: number;
+
+      /**
+       * Set to false to disable invocation logs
+       * @default true
+       */
+      invocation_logs?: boolean;
+
+      /**
+       * If logs should be persisted to the Cloudflare observability platform where they can be queried in the dashboard.
+       * @default true
+       */
+      persist?: boolean;
+
+      /**
+       * What destinations logs emitted from the Worker should be sent to.
+       * @default []
+       */
+      destinations?: string[];
+    };
+
+    /**
+     * Configuration for worker traces
+     */
+    traces?: {
+      /**
+       * Whether traces are enabled
+       * @default true
+       */
+      enabled?: boolean;
+
+      /**
+       * The sampling rate for traces
+       */
+      head_sampling_rate?: number;
+
+      /**
+       * If traces should be persisted to the Cloudflare observability platform where they can be queried in the dashboard.
+       * @default true
+       */
+      persist?: boolean;
+
+      /**
+       * What destinations traces emitted from the Worker should be sent to.
+       * @default []
+       */
+      destinations?: string[];
+    };
   };
 
   /**


alchemy/src/cloudflare/worker.ts
```diff
@@ -166,29 +166,88 @@ export interface BaseWorkerProps<
...
-   * @default false
-   */
-  logpush?: boolean;
+    /**
+     * A number between 0 and 1, where 0 indicates zero out of one hundred requests are logged, and 1 indicates every request is logged.
+     * If head_sampling_rate is unspecified, it is configured to a default value of 1 (100%).
+     * Read more about [head-based sampling](https://developers.cloudflare.com/workers/observability/logs/workers-logs/#head-based-sampling).
+     * @default 1
+     */
+    head_sampling_rate?: number;
```
sam-goodwin: We would usually use camelCase for consistency.

Do these properties get passed through as-is? I don't see any code using these props.

utopyin: yeah it seems like they are passed as-is

I'll fix the camelCase and map the props to their snake case version

utopyin: do we really want to manually map from camelCase to snake_case though? Like for example in limlits, the cpu_ms is using snake_case

```ts
limits?: {
    /**
     * The maximum CPU time in milliseconds that the worker can use.
     *
     * @see https://developers.cloudflare.com/workers/platform/limits/#cpu-time
     * @default 30_000 (30 seconds)
     */
    cpu_ms?: number;
  };
 ```

Comment (sam-goodwin) [reply alchemy/src/cloudflare/worker.ts]: Yeah I feel ya. I do like being 1:1 with API. Let's keep it snake

Comment (sam-goodwin) [reply alchemy/src/cloudflare/worker.ts]: Can we add a test? How would you test it?

Comment (utopyin) [reply alchemy/src/cloudflare/worker.ts]: I eventually figured out how to fetch from the observability API, but we must wait for about 2 minutes for the log to be queriable :(

Comment (sam-goodwin) [reply alchemy/src/cloudflare/worker.ts]: I think it's fine to just test the control plane for now. Let's assume that the binding works if cloudflare's api doesn't complain.

Comment (utopyin) [reply alchemy/src/cloudflare/worker.ts]: I added simple tests that check for the updated wrangler config.

Comment (utopyin) [reply alchemy/src/cloudflare/worker.ts]: Also found a lil copy pasta typo in the cpu limit test, I figured it might be worth patching it as well?

Comment (JacobMGEvans) [reply alchemy/src/cloudflare/worker.ts]: > do we really want to manually map from camelCase to snake_case though? Like for example in limlits, the cpu_ms is using snake_case

@sam-goodwin 
Maybe we can make a project level CamelCase TS utility, because I ran into a similar thing, I dont think I needed the raw Cloudflare API as an interface but having it colocated with the resources could be beneficial, just spitballing

Comment (Mkassabov) [reply alchemy/src/cloudflare/worker.ts]: I do prefer we do camelcase tbh, I feel like we don't have snake case in most of the api so it would feel strange to do it this way

File: alchemy/src/cloudflare/worker.ts [hunk 2]
```diff
@@ -166,29 +166,88 @@ export interface BaseWorkerProps<
...
-   * Controls whether worker logs are enabled
-   * @default { enabled: true }
+   * @see https://developers.cloudflare.com/workers/wrangler/configuration/#observability
+   * @default { enabled: true, head_sampling_rate: 1 }
    */
   observability?: {
     /**
-     * Whether to enable worker logs
+     * If observability is enabled for this Worker. Defaults to true for all new Workers.
      * @default true
```
JacobMGEvans: Seems redundant to mention it defaults to true in new workers if there is a `@default true` JSDoc field  
```suggestion
     * If observability is enabled for this Worker.
     * @default true
```

alchemy/test/cloudflare/worker.test.ts
```diff
@@ -2209,4 +2209,74 @@ describe("Worker Resource", () => {
...
+        `,
+        observability: baseObservability,
+      });
+
+      expect(worker.observability).toEqual(baseObservability);
+      expect(worker.observability?.logs).toBeUndefined();
+
+      // update the worker with observability with:
+      // - invocation_logs disabled
+      // - traces.head_sampling_rate set to 0.1
```
JacobMGEvans:
```suggestion
```

alchemy/test/cloudflare/worker.test.ts
```diff
@@ -2209,4 +2209,74 @@ describe("Worker Resource", () => {
...
+          export default {
+            async fetch(request, env, ctx) {
+              return new Response('Hello observability!', { status: 200 });
+            }
+          };
+        `,
+        observability: newObservability,
+      });
+
+      // verify the worker was updated successfully
```
JacobMGEvans:
```suggestion
```

alchemy/test/cloudflare/worker.test.ts
```diff
@@ -2209,4 +2209,74 @@ describe("Worker Resource", () => {
...
+    const workerName = `${BRANCH_PREFIX}-test-worker-observability`;
+
+    let worker: Worker | undefined;
+    try {
+      const baseObservability = {
+        enabled: true,
+        head_sampling_rate: 0.5,
+      };
+
+      // create a worker with no observability set (most things enabled by default)
```
JacobMGEvans:
```suggestion
```

alchemy/src/cloudflare/worker.ts
sam-goodwin: Fixed this by transforming

sam-goodwin (approved):

PR #1053, opened 2025-10-02
feat(prisma): prisma postgres resources

sorenbs:


alchemy/src/prisma/postgres/api.ts
```diff
@@ -0,0 +1,298 @@
...
+    this.url = props.url;
+    this.code = props.code;
+    this.responseBody = props.responseBody;
+  }
+}
+
+/**
+ * Minimal client for the Prisma Postgres Management API
+ */
+export class PrismaPostgresApi {
```
sam-goodwin, line 46: We've been trying to move away from hand-coding the API clients.

Does Prisma have a generated TypeScript SDK for its control plane?

@john-royal has been standardizing our API generation with hey-api. @john-royal, maybe you can take a look at their API spec and see how well it generates.

sam-goodwin:
This is very exciting @sorenbs. Great job.

Does Prisma Postgres support branching? I found this doc which indicates it does: https://www.prisma.io/blog/introducing-platform-environments

One feature to make sure we nail here would be CI/CD (see https://alchemy.run/guides/ci/). 

The ideal workflow is to have each PR deploy a branched copy of the infrastructure. It's not always possible depending on the service, but Planetscale and Neon each have branching capability exposed as Alchemy Resources that slot in nicely to Alchemy's CI/CD.

alchemy/src/prisma/postgres/api.ts
sorenbs: We don't have TypeScript SDK. hey-api looks really interesting.

sorenbs:
> This is very exciting @sorenbs. Great job.
> 
> Does Prisma Postgres support branching? I found this doc which indicates it does: https://www.prisma.io/blog/introducing-platform-environments
> 
> One feature to make sure we nail here would be CI/CD (see https://alchemy.run/guides/ci/).
> 
> The ideal workflow is to have each PR deploy a branched copy of the infrastructure. It's not always possible depending on the service, but Planetscale and Neon each have branching capability exposed as Alchemy Resources that slot in nicely to Alchemy's CI/CD.

We don't support Branching natively yet, but we have the ability to create a database from a snapshot. Currently a snapshot can only be created from the Console UI - I will speak to the team about getting it enabled in the API. I believe that should enable a complete CI/CD workflow as you describe.

sam-goodwin:
> I believe that should enable a complete CI/CD workflow as you describe.

That's similar to what we do for D1. 

Downside of this approach is cost, I'm assuming? Since it is a total copy?

alchemy-web/src/content/docs/guides/prisma-postgres.md
```diff
@@ -0,0 +1,67 @@
...
+  name: "prisma-app",
+  region: "us-east-1",
+  createDatabase: false,
+});
+
+const database = await Database("database", {
+  project,
+  name: "primary",
+  region: "us-east-1",
+});
```
Mkassabov, line 49: is it possible to make the `region` on database optional and just do `project.region`?

alchemy-web/src/content/docs/guides/prisma-postgres.md
```diff
@@ -0,0 +1,67 @@
...
+  Database,
+  DatabaseConnection,
+} from "alchemy/prisma/postgres";
+
+const app = await alchemy("prisma-postgres-example");
+
+const project = await Project("project", {
+  name: "prisma-app",
+  region: "us-east-1",
+  createDatabase: false,
```
Mkassabov, line 42: do we need `createDatabase: false` or can we set this as the default?

alchemy-web/src/content/docs/providers/prisma-postgres/database-connection.md
```diff
@@ -0,0 +1,41 @@
...
+## Rotate If Missing
+
+If the underlying connection is deleted, rerunning the resource automatically creates a replacement with a fresh secret:
+
+```ts
+await DatabaseConnection("application", {
+  database,
+  name: "application",
+});
+```
```
Mkassabov, line 30: I feel like this would work better as a `tip` on the create a connection section since you're explaining how the resource works not showing any unique configuration.

CC @JacobMGEvans

alchemy-web/src/content/docs/providers/prisma-postgres/database.md
```diff
@@ -0,0 +1,66 @@
...
+
+`Database` creates additional databases within a Prisma Postgres project. Databases inherit workspace access from their parent project and support restoring from backups created by Prisma.
+
+## Create a Database
+
+```ts
+import { Database } from "alchemy/prisma/postgres";
+
+const database = await Database("primary", {
+  project,
```
Mkassabov, line 14: where does project come from. its a bit ambiguous what project is here. Can we give an example with the simplest possible project?

alchemy-web/src/content/docs/providers/prisma-postgres/database-connection.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Create a Connection
+
+```ts
+import { DatabaseConnection } from "alchemy/prisma/postgres";
+
+const connection = await DatabaseConnection("application", {
+  database,
+  name: "application",
+});
```
Mkassabov, line 16: where does database come from. its a bit ambiguous what project is here. Can we give an example with the simplest possible database?

alchemy-web/src/content/docs/providers/prisma-postgres/database.md
```diff
@@ -0,0 +1,66 @@
...
+
+```ts
+import { Database } from "alchemy/prisma/postgres";
+
+const database = await Database("isolated", {
+  project,
+  name: "isolated",
+  region: "eu-central-1",
+  adopt: false,
+});
```
Mkassabov, line 32: we generally have adopt: false and you have to specify it manually or run with the `--adopt` flag. can we keep that behavior consistent with the rest of alchemy here.

alchemy-web/src/content/docs/providers/prisma-postgres/database.md
```diff
@@ -0,0 +1,66 @@
...
+  fromDatabase: {
+    database: sourceDatabase,
+    backupId: backups.mostRecent?.id,
+  },
+});
+```
+
+## Promote to Default Database
+
+Set `isDefault` to true to promote the database when the workspace allows it:
```
Mkassabov, line 57: this makes me hesitant since it could be `isDefault` on multiple databases but only one of them promotes. It dilutes alchemy as being the sense of truth.

could we do something like `canPromoteToDefault` or at least an output property to show if something is the default?

alchemy-web/src/content/docs/providers/prisma-postgres/project.md
```diff
@@ -0,0 +1,61 @@
...
+Disable adoption to ensure a fresh project is created:
+
+```ts
+import { Project } from "alchemy/prisma/postgres";
+
+const project = await Project("fresh", {
+  name: "fresh-project",
+  region: "eu-central-1",
+  createDatabase: false,
+  adopt: false,
```
Mkassabov, line 46: adopt should be false by default

Mkassabov (commented):
haven't looked at the code but read through the docs written here.
Its a good start, but some of the default behavior seems to differ from what we do for alchemy usually.

We'd prefer if we could align here but if there are good reasons for it we're open to discussion

alchemy-web/src/content/docs/guides/prisma-postgres.md
sorenbs: The way the region selection works is this:

- projects do not have a concept of region
- when you create a project and specify createDatabase: true, then you need to specify the region. You will then get a database in that region
- When you create a database in a project, you need to specify the region

To simplify things, what we could do in Alchemy is hide the ability to automatically create a database with the project. That would eliminate the `region` and `createDatabase` flags from Project creation.

alchemy-web/src/content/docs/guides/prisma-postgres.md
sorenbs: See comment above. I think it would be good to simplify this.

alchemy-web/src/content/docs/providers/prisma-postgres/database.md
sorenbs: Good catch - thank you!

alchemy-web/src/content/docs/providers/prisma-postgres/database.md
sorenbs: Yes. I was unsure what to do, so thanks for calling it out.

alchemy/src/prisma/postgres/api.ts
sorenbs: Would it be an option to use the hand-coded API client initially, and switch to a prisma provided SDK later?

alchemy-web/src/content/docs/guides/prisma-postgres.md
sam-goodwin: > To simplify things, what we could do in Alchemy is hide the ability to automatically create a database with the project. That would eliminate the region and createDatabase flags from Project creation.

This seems like the right approach. Centralize database config on the Database resource rather than conflate them.

alchemy-web/src/content/docs/providers/prisma-postgres/database.md
sam-goodwin: Inside a resource you need to check `props.adopt ?? this.scope.adopt`. 

`this.scope.adopt` is how we apply blanket adopt across an app or scope.

sorenbs:
> > I believe that should enable a complete CI/CD workflow as you describe.
> 
> That's similar to what we do for D1.
> 
> Downside of this approach is cost, I'm assuming? Since it is a total copy?

Yes, storage costs double. In you experience, is it a common use case to create a branch with the full production data? We haven't heard much request for this.

sam-goodwin:
> Yes, storage costs double. In you experience, is it a common use case to create a branch with the full production data? We haven't heard much request for this.

I don't see many people branch Prod, especially for a PR. But what we do care about is the cost per PR in terms of $$$ or quota consumption.

Prisma's free plan only allows 5 databases, which can be hit quite easily. So while a DB might be low $$$ cost when it contains less data, a PR still has a high, fixed "quota" cost.

image

sorenbs:
> I don't see many people branch Prod, especially for a PR.

👍 

> Prisma's free plan only allows 5 databases, which can be hit quite easily. So while a DB might be low $$$ cost when it contains less data, a PR still has a high, fixed "quota" cost.

We are currently fixing this by 10x-ing the db limits. So Free will give you 50 databases, and the $10 plan will include 100 dbs. As soon as a PR is closed, the database can be deleted and freed up for the next PR.

sam-goodwin:
> We are currently fixing this by 10x-ing the db limits. So Free will give you 50 databases, and the $10 plan will include 100 dbs. As soon as a PR is closed, the database can be deleted and freed up for the next PR.

Nice, sounds good. 100 databases is high enough to not worry about it.

alchemy-web/src/content/docs/providers/prisma-postgres/database.md
sorenbs: I noticed that Planetscale refers to orgs and databases by id instead of objects. Is that the preferred approach?

```
const devBranch = await Branch("feature-123", {
  name: "feature-123",
  organizationId: "my-org",
  databaseName: "my-database",
  parentBranch: "main",
  isProduction: false,
});
```

sam-goodwin: > I noticed that Planetscale refers to orgs and databases by id instead of objects. Is that the preferred approach?

No, but great question. 

The preferred approach is to drop the `Id` suffix and use a union type. For example: `organization: string | Organization`. 

The planetscale should be:
```ts
const devBranch = await Branch("feature-123", {
  name: "feature-123",
  organization: "my-org" | MyOrg,
  database: "my-database" | MyDatabase,
  parentBranch: "main",
  isProduction: false,
});
```

@Mkassabov we should make these changes in a separate PR.

Mkassabov: @sorenbs 
better example is our new clickhouse provider:
https://github.com/alchemy-run/alchemy/blob/80ef082c2835ebdd0329302058c4dc16db70f729/alchemy/src/clickhouse/service.ts#L390-L395

examples/prisma-postgres/alchemy.run.ts
```diff
@@ -0,0 +1,27 @@
...
+export const database = await Database("database", {
+  project,
+  name: "primary",
+  region: "us-east-1",
+});
+
+export const connection = await DatabaseConnection("connection", {
+  database,
+  name: "application",
+});
```
sam-goodwin, line 23: Should we also use hyperdrive ?

sam-goodwin: Hyperdrive and bind to a Worker would be killer demo

Mkassabov:
rewrote this in #1123 to be more aligned with other alchemy resources

Mkassabov:
closing this since we added prisma resources in #1123

PR #1063, opened 2025-10-02
fix(core): use scrypt for key derivation

john-royal:


Mkassabov (approved):
works tests passing

alchemy/src/serde.ts
```diff
@@ -183,7 +183,7 @@ export async function deserialize(
             "See: https://alchemy.run/concepts/secret/#encryption-password",
         );
       }
-      if (typeof value["@secret"] === "object") {
+      if (typeof value["@secret"] === "object" && "data" in value["@secret"]) {
```
sam-goodwin, line 186: What happens if data is not present? Do we just skip it?

john-royal: Quick answer: no, we don't skip it. The `data` property indicates an encrypted object; otherwise it's an encrypted string. This is the same as before.

```diff
-      if (typeof value["@secret"] === "object") {
+      if (typeof value["@secret"] === "object" && "data" in value["@secret"]) {
        return new Secret(
          JSON.parse(
            await decryptWithKey(value["@secret"].data, scope.password),
          ),
        );
      }
      return new Secret(await decryptWithKey(value["@secret"], scope.password));
```

Why the change? The encrypted value is now an [object with properties](https://github.com/alchemy-run/alchemy/pull/1063/files#diff-d321ef7ae3f0b1da62fe32aea4ed9f267260c1c052ecda14ec8e8c2eb4827e4bR11-R17), so `typeof value["@secret"] === "object"` is always true. Checking for the `data` property is how we differentiate the two cases.

The [serialization logic](https://github.com/alchemy-run/alchemy/pull/1063/files#diff-8344e7b3275004e0d66113116190bd08d8a6e6d4974b5406879e8b742fcebfadR76-R95) is completely unchanged.

sam-goodwin: I was looking for the code that is handling backwards compatibility (decrypting old secrets). Can you point that out to me.

john-royal: That's handled by the `decryptWithKey` function internally: https://github.com/alchemy-run/alchemy/pull/1063/files#diff-d321ef7ae3f0b1da62fe32aea4ed9f267260c1c052ecda14ec8e8c2eb4827e4bR23-R31

alchemy/src/encrypt.ts
```diff
@@ -1,11 +1,114 @@
...
+
+export function decryptWithKey(
+  value: string | Encrypted,
+  key: string,
+): Promise<string> {
+  if (typeof value === "string") {
+    return libsodiumDecrypt(value, key);
+  }
+  return scryptDecrypt(value, key);
+}
```
JacobMGEvans, line 31: Maybe add something like this, for inline documentation of changes in the system 😄 

```suggestion
/**
 * @legacy Decrypts a value preserving backward compatibility with
 * legacy implementation.
 */
export function decryptWithKey(
  value: string | Encrypted,
  key: string,
): Promise<string> {
  if (typeof value === "string") {
    return libsodiumDecrypt(value, key);
  }
  return scryptDecrypt(value, key);
}
```

PR #1065, opened 2025-10-03
feat(cloudflare): HealthCheck Resource

JacobMGEvans:


alchemy/src/cloudflare/health-check.ts
```diff
@@ -0,0 +1,577 @@
...
+   * @internal
+   */
+  healthCheckId?: string;
+}
+
+/**
+ * Output returned after Cloudflare Health Check creation/update
+ * IMPORTANT: The interface name MUST match the exported resource name
+ */
+export interface HealthCheck extends HealthCheckProps {
```
sam-goodwin: should be type

alchemy/src/cloudflare/health-check.ts
```diff
@@ -0,0 +1,577 @@
...
+ * });
+ */
+export const HealthCheck = Resource(
+  "cloudflare::HealthCheck",
+  async function (
+    this: Context<HealthCheck>,
+    id: string,
+    props: HealthCheckProps,
+  ): Promise<HealthCheck> {
+    const healthCheckId = props.healthCheckId || this.output?.healthCheckId;
```
sam-goodwin: This doesn't look quite right. Does this need to be unique within an account?

Think we should be doing:
```ts
const healthCheckId = props.healthCheckId ?? this.createPhysicalName(id);
```

Only look at `this.output?.healthCheckId` inside the `this.phase === "delete"` block

sam-goodwin: Is it the ID or the name?

alchemy/src/cloudflare/health-check.ts
```diff
@@ -0,0 +1,577 @@
...
+      }
+
+      try {
+        const deleteResponse = await api.delete(
+          `/zones/${props.zoneId}/healthchecks/${healthCheckId}`,
+        );
+
+        if (!deleteResponse.ok && deleteResponse.status !== 404) {
+          await handleApiError(deleteResponse, "delete", "health check", id);
+        }
```
sam-goodwin: Pull these out into their own exported functions, e.g. `deleteHealthCheck`

alchemy/src/cloudflare/health-check.ts
```diff
@@ -0,0 +1,577 @@
...
+  | "NSAM"
+  | "SSAM"
+  | "OC"
+  | "ME"
+  | "NAF"
+  | "SAF"
+  | "IN"
+  | "SEAS"
+  | "NEAS"
+  | "ALL_REGIONS";
```
Mkassabov: can we add `| (string & {})` to the end of this. user still gets autocomplete but they are allowed to enter in any value.

alchemy/src/cloudflare/health-check.ts
```diff
@@ -0,0 +1,577 @@
...
+   * @default 80
+   */
+  port?: number;
+}
+
+export interface HealthCheckProps extends CloudflareApiOptions {
+  /**
+   * Zone ID this health check belongs to
+   */
+  zoneId: string;
```
Mkassabov: can we have this be `zone: string | Zone;` since we have a zone resource

alchemy/src/cloudflare/health-check.ts
```diff
@@ -0,0 +1,577 @@
...
+  /**
+   * A case-insensitive sub-string to look for in the response body
+   * If this string is not found, the origin will be marked as unhealthy
+   */
+  expectedBody?: string;
+
+  /**
+   * The expected HTTP response codes (e.g. "200") or code ranges (e.g. "2xx" for all codes starting with 2)
+   */
+  expectedCodes?: string[];
```
Mkassabov, line 53: optional but no defaults in jsdoc?

alchemy/src/cloudflare/health-check.ts
```diff
@@ -0,0 +1,577 @@
...
+   */
+  path?: string;
+
+  /**
+   * Port number to connect to for the health check
+   * @default 80 for HTTP
+   * @default 443 for HTTPS
+   */
+  port?: number;
+}
```
Mkassabov, line 88: im confused when these defaults are used, what determines if this is http vs https?

alchemy/src/cloudflare/health-check.ts
JacobMGEvans: I dont think there are other Regions beyond these for this Resource, but happy to add that

alchemy/src/cloudflare/health-check.ts
JacobMGEvans: https://github.com/alchemy-run/alchemy/pull/1065/files/51384454f782536cf9590a16fbf00fa337e53ee0#diff-af39c7cf544f72400a10acf0649ab2c9a14696133810c5526176d271aa04bf7fR173-R178

JacobMGEvans: what happen to the permalink previews 😢

alchemy/src/cloudflare/health-check.ts
sam-goodwin: Yeah add that so we never block people when a new one is added

alchemy/src/cloudflare/health-check.ts
```diff
@@ -0,0 +1,577 @@
...
+   * @default false
+   */
+  followRedirects?: boolean;
+
+  /**
+   * The HTTP request headers to send in the health check
+   * It is recommended you set a Host header by default
+   * The User-Agent header cannot be overridden
+   */
+  header?: Record<string, string[]>;
```
sam-goodwin: Should we support secret values? `Record<string, (string | Secret<string>)[]>`

PR #1067, opened 2025-10-03
feat(cloudflare): add WorkerLoader binding type

gingerhendrix:
Add support for [Dynamic Worker Loaders](https://developers.cloudflare.com/workers/runtime-apis/bindings/worker-loader/)


examples/cloudflare-worker-loader/src/env.ts
```diff
@@ -0,0 +1,11 @@
+import type { worker } from "../alchemy.run.js";
```
Mkassabov: shouldn't this be `../alchemy.run.ts`

examples/cloudflare-worker-loader/alchemy.run.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const loader = WorkerLoader();
+
+export const worker = await Worker("worker", {
+  entrypoint: "./src/worker.ts",
+  bindings: {
+    LOADER: loader,
+  },
+  url: true,
+});
```
Mkassabov: can this just be
```
export const worker = await Worker("worker", {
  bindings: {
    LOADER: WorkerLoader(),
  },
  url: true,
});
```

makes the example simpler

Mkassabov (changes requested):
1. missing docs - just add an example under the [workers section](https://alchemy.run/providers/cloudflare/worker/)
2. missing tests


LMK if you don't feel comfortable making those changes and I can step in

examples/cloudflare-worker-loader/alchemy.run.ts
sam-goodwin: You also don't need url true.  That's default

examples/cloudflare-worker-loader/src/env.ts
gingerhendrix: I actually copied from the cloudflare-worker example - it turns out the examples use js/ts inconsistently (also `env.ts`/`env.d.ts`).

sam-goodwin: We should be using .ts. Any .js is a mistake and we should fix those.

gingerhendrix: Since I'd grepped already - https://github.com/alchemy-run/alchemy/pull/1068

alchemy/test/cloudflare/worker-loader.test.ts
```diff
@@ -0,0 +1,75 @@
...
+                          return new Response('Hello from dynamic worker!');
+                        }
+                      }
+                    \`,
+                  },
+                }),
+              );
+
+              const entrypoint = dynamicWorker.getEntrypoint();
+              return entrypoint.fetch(new URL(request.url));
```
sam-goodwin, line 50: What an awesome test. Super powerful feature.

sam-goodwin:
@john-royal what are we gonna do for local dev mode?

Does miniflare support it?

gingerhendrix:
> Does miniflare support it?

it seems to just work, but I haven't tested it beyond the simplest example.

john-royal:
> > Does miniflare support it?
> 
> it seems to just work, but I haven't tested it beyond the simplest example.

Funny enough, I tried implementing it the exact same way a few days ago but it didn't work. Perhaps it was a problem on my end.

alchemy/src/cloudflare/worker-loader.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * A binding for dynamic worker loaders.
+ *
+ */
+export interface WorkerLoader {
```
john-royal: Quick nitpick - this is an Alchemy convention to help with syntax highlighting:

```suggestion
export type WorkerLoader = {
```

alchemy/src/cloudflare/miniflare/build-worker-options.ts
```diff
@@ -308,6 +308,12 @@ export const buildWorkerOptions = async (
         };
         break;
       }
+      case "worker_loader": {
+        (options.workerLoaders ??= {})[key] = {
+          name: key,
```
john-royal: Interesting - last time I tried, the type of the binding was an empty object. Based on the schema from the [internal Miniflare implementation](https://github.com/cloudflare/workers-sdk/blob/e483b78e8c6d266e4fc97d79853d5fa6447cc807/packages/miniflare/src/plugins/worker-loader/index.ts), that appears to still be the case. Do you know if it works without setting that `name` property?

alchemy-web/src/content/docs/providers/cloudflare/worker.mdx
```diff
@@ -327,6 +327,57 @@ export default {
...
+```
+
+Then use the loader to create workers on-demand:
+
+```ts
+// ./src/worker.ts
+import { env } from "cloudflare:workers";
+
+export default {
+  async fetch(request) {
```
john-royal: Two nitpicks here:
- Prefer explicit passing of `env` to import from `cloudflare:workers`, especially for simple examples.
- Prefer explicit typing, including of environment variables, using an import from `alchemy.run.ts`. Otherwise, if you copy this code, you'll get some `any` types.

```suggestion
import type { worker } from "../alchemy.run.ts";

export default {
  async fetch(request: Request, env: typeof worker.Env) {
```

examples/cloudflare-worker-loader/.gitignore
john-royal, line 1: Is it safe to remove this file?

john-royal (changes requested):
Looks awesome! Just some questions and nitpicks.

examples/cloudflare-worker-loader/.gitignore
gingerhendrix: 👍 it comes from copying the worker example

alchemy/src/cloudflare/miniflare/build-worker-options.ts
gingerhendrix: yeah, this was overreach from claude - empty object works fine.

sam-goodwin (approved):

alchemy/test/cloudflare/worker-loader.test.ts
```diff
@@ -0,0 +1,75 @@
...
+          LOADER: WorkerLoader(),
+        },
+      });
+
+      expect(worker.id).toBeTruthy();
+      expect(worker.name).toEqual(workerName);
+      expect(worker.bindings?.LOADER).toBeDefined();
+      expect(worker.url).toBeTruthy();
+
+      // Test that the worker can use the WorkerLoader binding
```
JacobMGEvans: inferred from the assertions 
```suggestion
```

alchemy/test/cloudflare/worker-loader.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+const api = await createCloudflareApi();
+
+describe("WorkerLoader", () => {
+  test("create worker with WorkerLoader binding", async (scope) => {
+    const workerName = `${BRANCH_PREFIX}-test-worker-loader`;
+
+    let worker: Worker | undefined;
+    try {
+      // Create a worker with a WorkerLoader binding
```
JacobMGEvans: inferred from the test description
```suggestion
```

PR #1072, opened 2025-10-06
chore: update remote bindings docs

sam-goodwin:


john-royal (approved):

alchemy-web/src/content/docs/concepts/dev.mdx
JacobMGEvans, line 1: Should the phrasing of the "deployed resource" emphasize that its a temporary or ephemeral, not to confuse it with "real" as "production" or something similar?

JacobMGEvans (approved):
Non-blocking question.

alchemy-web/src/content/docs/concepts/dev.mdx
john-royal: Not sure. The remote resource isn't temporary or ephemeral; it sticks around until you run destroy.

PR #1075, opened 2025-10-06
docs(AGENTS): enhance resource implementation

JacobMGEvans:


AGENTS.md
```diff
@@ -294,6 +294,199 @@ yarn tsx ./alchemy.run --destroy
...
+// ❌ INCORRECT: Interface name doesn't match
+export interface HyperdriveOutput extends Resource<"cloudflare::Hyperdrive"> {
+  // ... properties  
+}
+export const Hyperdrive = Resource(/* ... */);
+```
+
+```ts
+// 1. Define Props interface for input parameters
+export interface MyResourceProps extends CloudflareApiOptions {
```
sam-goodwin: CloudflareApiOptions is only specific to cloudflare resurce

AGENTS.md
```diff
@@ -294,6 +294,199 @@ yarn tsx ./alchemy.run --destroy
...
+   */
+  name?: string;
+  
+  /**
+   * Configuration object
+   */
+  config: {
+    property: string;
+    secret?: string | Secret;
+  };
```
sam-goodwin: Don't love this. We should encourage flat properties

AGENTS.md
```diff
@@ -405,6 +710,94 @@ test("end-to-end workflow", async (scope) => {
...
+Alchemy provides secure secret handling with encryption:
+
+```ts
+// Create encrypted secrets
+const secret = alchemy.secret(process.env.API_KEY);
+
+// Use in resource props
+const resource = await MyResource("api", {
+  apiKey: secret,
+  database: alchemy.secret(process.env.DB_PASSWORD)
```
sam-goodwin: note: `alchemy.secret.env.DB_PASSWORD` gives better error messages, but LLMs are more familiar with `process.env.DB_PASSWORD`

AGENTS.md
JacobMGEvans: It was suppose to be a reference, it was what I was most familiar with but I will make it more generic lol

AGENTS.md
JacobMGEvans: If it expects a nested object then we create a new interface or type then use it there instead then.

AGENTS.md
```diff
@@ -294,6 +294,228 @@ yarn tsx ./alchemy.run --destroy
...
+
+Flat properties provide:
+- Better IDE autocomplete and type checking
+- Cleaner resource creation syntax
+- Easier validation and error handling
+- More intuitive API design
+
+```ts
+// ✅ CORRECT: Interface name matches exported resource name
+export interface Hyperdrive extends Resource<"cloudflare::Hyperdrive"> {
```
john-royal: Extending `Resource` is no longer needed, but `export type` is recommended for syntax highlighting:
```suggestion
export type Hyperdrive = {
```

PR #1081, opened 2025-10-07
fix(cloudflare): better error if an empty string is passed as a domain

Mkassabov:
normally this errors with a problem about scope needing a name, which is confusing AF:

see disc thread: https://discord.com/channels/1359694195782320389/1425025657691308074


Mkassabov:
> Makes sense. Maybe we should validate the whole domain as a further improvement?
> 
> Although that can be notoriously difficult to get right.
> 
> Maybe new URL()

I thought about it and decided against.
If its a bad domain they would get a CF error now; this is mostly to protect us from not even getting to the cloudflare error due to a scope issue

luisrudge:
Thanks for the improvement!

PR #1083, opened 2025-10-07
fix(core): validate Resource and Scope IDs are non-empty

sam-goodwin:


alchemy/src/scope.ts
```diff
@@ -265,21 +266,17 @@ export class Scope {
...
-      throw new Error(
-        `Scope name "${this.scopeName}" cannot contain double colons`,
-      );
-    }
-
     this.stage = stage ?? this.parent?.stage ?? DEFAULT_STAGE;
     this.profile = profile ?? this.parent?.profile;
     this.parent?.children.set(this.scopeName!, this);
     this.quiet = quiet ?? this.parent?.quiet ?? false;
     if (this.parent && !this.scopeName) {
```
sam-goodwin: remove `!this.scopeName` from this if

sam-goodwin: "if we're not the root scope, then do these validations"

PR #1084, opened 2025-10-08
feat: resend provider

john-royal:
- [x] Audience
- [x] ApiKey
- [ ] Domain
  - [x] create/delete implementation
  - [ ] update implementation - failing with API error
  - [ ] put DNS records and update?

### DIFF
--- a/alchemy/src/resend/domain.ts
+++ b/alchemy/src/resend/domain.ts
@@ -102,23 +102,6 @@ export const Domain = Resource(
   ): Promise<Domain> {
     const resend = createResend(props);
 
-    if (this.phase === "delete") {
-      if (this.output.id) {
-        const { error } = await resend.deleteDomainsByDomainId({
-          path: {
-            domain_id: this.output.id,
-          },
-          throwOnError: false,
-        });
-        if (error && error.status !== 404) {
-          throw new Error(`Failed to delete domain "${id}"`, {
-            cause: error,
-          });
-        }
-      }
-      return this.destroy();
-    }
-
     switch (this.phase) {
       case "create": {
         const { data } = await resend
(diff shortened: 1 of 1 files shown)


alchemy/src/resend/audience.ts
```diff
@@ -0,0 +1,100 @@
...
+import { Resource } from "../resource.ts";
+import { createResend, type ResendProps } from "./api.ts";
+
+export interface AudienceProps extends ResendProps {
+  /**
+   * The name of the audience.
+   * @default ${app}-${stage}-${id}
+   */
+  name?: string;
+}
```
john-royal, line 11: Audiences can't be adopted (name is non-unique), but we do have[ a way to reference an existing one by ID](https://github.com/alchemy-run/alchemy/pull/1084/files#diff-58578214ee149f178184248daf29411d4fa342e050a3eebe0deb50587e173309R35-R53).

Should we let you adopt by ID by passing in like an `existingId` prop? Not sure if there's a point since the only property available is the `name`.

alchemy/src/resend/audience.ts
```diff
@@ -0,0 +1,100 @@
...
+        id: props.id,
+      },
+    });
+    return {
+      id: data.id!,
+      name: data.name!,
+      createdAt: new Date(data.created_at!),
+    };
+  };
+}
```
john-royal, line 53: This feels cleaner to me than a separate `AudienceRef` but happy to change to that to follow conventions.

alchemy/src/resend/domain.ts
```diff
@@ -0,0 +1,220 @@
...
+                (domain) => domain.name === props.name,
+              );
+              if (domain) {
+                return await resend.getDomainsByDomainId({
+                  path: {
+                    domain_id: domain.id!,
+                  },
+                });
+              }
+            }
```
john-royal, line 132: This is necessary because the list response doesn't include DNS records. However, I ran into an issue where the first API call succeeded but the second one failed? Not sure what to do about that.

scripts/clients/resend.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export const patch = async () => {
+  const file = Bun.file("alchemy/src/resend/api/sdk.gen.ts");
+  let text = await file.text();
+  text = text.replace(
+    "import",
+    'import type {ResendError} from "../api.ts"\nimport',
+  );
+  await file.write(text.replaceAll(/unknown,/g, "ResendError,"));
+};
```
john-royal, line 36: Their OpenAPI spec doesn't have any errors, so this is a hack to sub in our own error type.

PR #1089, opened 2025-10-08
feat(cli): add command to make a cloudflare god token

Mkassabov:


sam-goodwin:
Missing docs. Would prefer to make it a first class feature than an "off to the side, god mode token feature"

PR #1095, opened 2025-10-09
Fix-build-args

eastlondoner:
Quoting docker args is bad, the quotes get passed through to the Dockerfile breaking the number one use case for args. This removes the quotes and adds args to the examples test

This will fail

```
ARG BASE_DOCKERFILE
ARG BASE_DOCKERFILE_TAG
FROM ${BASE_DOCKERFILE}:${BASE_DOCKERFILE_TAG}
```

with error

```
 - InvalidDefaultArgInFrom: Default value for ARG ${BASE_DOCKERFILE}:${BASE_DOCKERFILE_TAG} results in empty or invalid base image name (line 4)
Dockerfile:4
--------------------
   2 |     ARG BASE_DOCKERFILE
   3 |     ARG BASE_DOCKERFILE_TAG
   4 | >>> FROM ${BASE_DOCKERFILE}:${BASE_DOCKERFILE_TAG}
   5 |     
   6 |     USER 1000:1000
--------------------
ERROR: failed to solve: failed to parse stage name "\"andrewjefferson/mineflare-base\":\"cd5796be9936\"": invalid reference format
```

image

### DIFF
--- a/alchemy/src/docker/image.ts
+++ b/alchemy/src/docker/image.ts
@@ -190,7 +190,7 @@ export const Image = Resource(
 
       // Add build arguments
       for (const [key, value] of Object.entries(buildOptions)) {
-        buildArgs.push("--build-arg", `${key}="${value}"`);
+        buildArgs.push("--build-arg", `${key}=${value}`);
       }
 
       buildArgs.push("-f", dockerfile);
--- a/examples/cloudflare-container/Dockerfile
+++ b/examples/cloudflare-container/Dockerfile
@@ -1,6 +1,9 @@
 # syntax=docker/dockerfile:1
 
-FROM golang:1.24-alpine AS build
+# deliberately use an unset arg to ensure arg functionality works
+ARG IMAGE_VERSION
+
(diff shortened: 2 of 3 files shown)


alchemy/src/cloudflare/bun-spa/bun-spa.ts
```diff
@@ -15,7 +15,12 @@ import {
 } from "../website.ts";
 
 export interface BunSPAProps<B extends Bindings> extends WebsiteProps<B> {
-  frontend: string;
+  /**
+   * The path to the frontend entrypoints that bun should bundle for deployment & serve in dev mode.
+   * These are usually html files. Glob patterns are supported.
+   * Typically set to index.html
+   */
+  frontend: string | string[];
```
sam-goodwin, line 23: Are you on latest main? Thought we  merged that change already

eastlondoner: Oh shit I can’t use git

PR #1098, opened 2025-10-09
fix(cloudflare): improve websocket proxy handling

john-royal:
## Update

- I updated `browser-rendering.test.ts` to test both locally and in production. Was able to reproduce the problem with websockets and fix.
- We weren't including all relevant headers when proxying the websocket. This has been resolved.
- I copied over some websocket handling code from Miniflare. This includes what was found regarding exit codes.

## Background

image


alchemy/src/cloudflare/miniflare/miniflare-worker-proxy.ts
```diff
@@ -84,13 +84,11 @@ export async function createMiniflareWorkerProxy(options: {
     options.transformRequest?.(info);
     const name = options.getWorkerName(info);
 
-    const server = new WebSocket(url, protocols, {
+    return new WebSocket(url, protocols, {
       headers: {
         "MF-Route-Override": name,
       },
     });
-    await once(server, "open");
```
john-royal, line 92: This is now handled by `coupleWebSocket` for consistency with the remote binding proxy.

alchemy/test/cloudflare/browser-rendering.test.ts
john-royal, line 1: Updated this to also run locally. Test is now passing after websocket/proxy update.

alchemy/src/cloudflare/miniflare/remote-binding-proxy.ts
```diff
@@ -61,11 +61,19 @@ export async function createRemoteProxyWorker(input: {
     websocket: async (req) => {
       const input = new URL(req.url);
       const proxied = new URL(input.pathname + input.search, `https://${host}`);
+      const headers: Record<string, string> = {
+        "cf-workers-preview-token": token,
+        host,
+      };
+      // We don't want to include all headers because some can mess with websocket connections.
+      // However, it's important to include `mf-` prefixed headers to configure bindings.
```
ericclemmons:
```suggestion
      // However, it's important to include `mf-` prefixed headers to configure bindings for miniflare to work.
```

(I had to lookup what `mf-*` was for :D)

alchemy/src/util/http.ts
```diff
@@ -11,19 +12,22 @@ export class HTTPServer {
     websocket?: (request: Request) => Promise<WebSocket>;
     fetch: (request: Request) => Promise<Response>;
   }) {
+    const websocket = options.websocket;
+    if (websocket) {
+      const wss = new WebSocketServer({ noServer: true });
+      this.webSocketServer = wss;
+      this.httpServer.on(
+        "upgrade",
+        createUpgradeHandler({
```
ericclemmons, line 21: I'm diggin this consolidation.

I'm assuming this is what you were referring to on today's call about "some places we handle it already"?

alchemy/src/util/http.ts
```diff
@@ -48,32 +52,58 @@ export class HTTPServer {
...
+
+export function createUpgradeHandler(props: {
+  wss: WebSocketServer;
+  createServerWebSocket: (req: http.IncomingMessage) => Promise<WebSocket>;
+}) {
+  async function waitForWebSocketOpen(server: WebSocket) {
+    if (server.readyState === WebSocket.CONNECTING) {
+      // Wait for client to be open before accepting worker pair (which would
+      // release buffered messages). Note this will throw if an "error" event is
+      // dispatched (https://github.com/cloudflare/miniflare/issues/229).
```
ericclemmons, line 68: Great resource linking here 👍

alchemy/src/util/http.ts
```diff
@@ -48,32 +52,58 @@ export class HTTPServer {
...
-    await new Promise<void>((resolve, reject) => {
-      this.httpServer.close((err) => (err ? reject(err) : resolve()));
+  }
+
+  function forwardWebSocketEvents(from: WebSocket, to: WebSocket) {
+    const ready = waitForWebSocketOpen(to);
+    from.on("message", (event, binary) => {
+      // Wait for target to be open before sending message;
+      // Silently discard messages received after close:
+      // https://www.rfc-editor.org/rfc/rfc6455#section-1.4
```
ericclemmons: Dude, TIL!

I thought these were supposed to be queued up and sent once the connection was established.

alchemy/src/util/http.ts
```diff
@@ -48,32 +52,58 @@ export class HTTPServer {
...
+      void ready.then(() => {
+        to.send(event, { binary });
+      });
+    });
+    from.on("close", (code, reason) => {
+      // Handle close events similar to Miniflare:
+      // https://github.com/cloudflare/workers-sdk/blob/88f081f4e2bd299e715d18bcfe181971f534ff76/packages/miniflare/src/http/websocket.ts#L276-L282
```
ericclemmons, line 94: Again, love the research

alchemy/src/util/http.ts
```diff
@@ -48,32 +52,58 @@ export class HTTPServer {
...
+
+      if (code === 1005 /* No Status Received */) {
```
ericclemmons: TIL: https://datatracker.ietf.org/doc/html/rfc6455#section-7.1.5

alchemy/src/util/http.ts
```diff
@@ -48,32 +52,58 @@ export class HTTPServer {
...
-    server.on("message", (event, binary) => client.send(event, { binary }));
-    server.on("close", (code, reason) => client.close(code, reason));
-    wss.emit("connection", client, req);
-  });
+  return async (req: http.IncomingMessage, socket: Duplex, head: Buffer) => {
+    const server = await props.createServerWebSocket(req);
+    props.wss.handleUpgrade(req, socket, head, async (client) => {
+      forwardWebSocketEvents(server, client);
+      forwardWebSocketEvents(client, server);
+      props.wss.emit("connection", client, req);
```
ericclemmons, line 120: Does this event mean "connected and ready to receive messages?"

In my patch I tried to wait for the `server`'s `open` event before emitting `connection`.

Or did I just misunderstand how websockets work?

examples/cloudflare-worker-simple/e2e.test.ts
```diff
@@ -0,0 +1,147 @@
+import { hc } from "hono/client";
+import assert from "node:assert";
+import fs from "node:fs/promises";
+import { after, before, describe, it } from "node:test";
+import type app from "./src/worker1.ts";
+
+describe("cloudflare-worker-simple", () => {
```
ericclemmons, line 7: Damn @john-royal you went hard on this one 👀

alchemy/src/util/http.ts
john-royal: They are, but once it closes, apparently you're supposed to just silently ignore it which is interesting.

The `ready.then(() => ...)` is my way of queueing without actually buffering them in memory. Though admittedly ordering might be an issue there.

alchemy/src/util/http.ts
john-royal: Exactly!

alchemy/src/util/http.ts
john-royal: Interesting thing: I thought that too. I wanted to wait for the server websocket to open before fully accepting the connection so that way we don't have to buffer messages in-memory.

But I found that if we blocked on that before calling `wss.handleUpgrade`, either the `ws` module would throw some strange internal error or messages would get lost. So instead, we accept the connection but assume it might not be ready to receive messages and handle that accordingly. Learned the hard way why buffering is so common!

alchemy/src/util/http.ts
john-royal: The real interesting thing is that forcing this consolidation revealed edge cases that weren't present if you allowed the implementation to be different between the two. So the current implementation is (hopefully) robust to more use cases.

sam-goodwin (approved):

PR #1100, opened 2025-10-09
fix: update links to point to alchemy org pages

hacknug:


hacknug:
Forgot to mention I didn't touch these files in case something could break and to avoid blocking the merge.

- https://github.com/alchemy-run/alchemy/blob/main/CHANGELOG.md
- https://github.com/alchemy-run/alchemy/blob/main/scripts/bump.ts
- https://github.com/alchemy-run/alchemy/blob/main/alchemy/test/github/repository-environment.test.ts

hacknug:
This is now ready. Didn't touch things pointing to `sam-goodwin/test-alchemy-resources` since that seems to still be under Sam's account 👍

PR #1106, opened 2025-10-09
feat(cloudflare): DevTunnel pseudo-resource allows specific hostnames to be attached to workers in dev

eastlondoner:
A single tunnel can handle many workers - I used the same dev-machine-side single proxy worker and the tunnel directs all traffic to that proxy. The proxy looks at the request hosts and based on those routes to an assigned worker.

You create a DevTunnel instance like this - it creates a single tunnel under the hood but only in local scope and "lazily" inside miniflare-controller

```ts
const devTunnel = await DevTunnel("mineflare-agent-dev-tunnel", {
  name: `${app.name}-agent-dev`,
  adopt: true,
  hostnames: ["mineflare-agent-dev.gptkids.app", "mineflare-dynmap-dev.gptkids.app"],
});
```

You have to specify the hostnames you are going to use up front so that we can create the tunnel before all the hostnames have been assigned workers. The tunnel definition doesn't care about the workers, only the port of the proxy worker.

Once you have a dev tunnel you can vend out "routes" from it to your workers. This is *typesafe* the hostnames are consumed as a tuple of string literals and you can only use a hostname that's in the original list of you'll get type errors.

```ts
const agentWorker = await Worker("mineflare-agent", {
  name: `${app.name}-agent`,
  entrypoint: "src/agent.ts",
  adopt: true,
  compatibility: "node",
  bindings: {
    MCP_OBJECT: agentDO,
    MINEFLARE_WORKER: worker // bind to the main worker
  },
  dev: {
    tunnel: devTunnel.addRoute("mineflare-agent-dev.gptkids.app"),
  }
});

export const dynmapWorker = await Worker("dynmap-worker", {
  name: `${app.name}-dynmap-public`,
  entrypoint: "src/dynmap-worker.ts",
  adopt: true,
  bindings: {
    DYNMAP_BUCKET: dynmapBucket,
    BUCKET_DOMAIN: dynmapBucket.domain ?? "",
    MINECRAFT_WORKER_URL: "https://*", // should be worker url but we don't know it yet
  },
  dev: {
    tunnel: devTunnel.addRoute("mineflare-dynmap-dev.gptkids.app"),
  }
});
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,32 @@
+## v0.72.0
+
+### &nbsp;&nbsp;&nbsp;🚀 Features
+
+- **planetscale**: Orgref and object based resource params &nbsp;-&nbsp; by **Michael K** in https://github.com/alchemy-run/alchemy/issues/1076 [<samp>(62087)</samp>](https://github.com/alchemy-run/alchemy/commit/62087823)
+
+##### &nbsp;&nbsp;&nbsp;&nbsp;[View changes on GitHub](https://github.com/alchemy-run/alchemy/compare/v0.71.1...v0.72.0)
+
+---
+
+## v0.71.1
+
+### &nbsp;&nbsp;&nbsp;🚀 Features
+
+- **cloudflare**: HealthCheck Resource &nbsp;-&nbsp; by **Jacob M-G Evans** in https://github.com/alchemy-run/alchemy/issues/1065 [<samp>(28d1f)</samp>](https://github.com/alchemy-run/alchemy/commit/28d1fb0d)
+- **test**: Export the test vite util &nbsp;-&nbsp; by **Sam Goodwin** [<samp>(6acff)</samp>](https://github.com/alchemy-run/alchemy/commit/6acff8ea)
+
+### &nbsp;&nbsp;&nbsp;🐞 Bug Fixes
+
+- **cloudflare**:
+  - Use remote instead of experimental_remote in wrangler.json &nbsp;-&nbsp; by **Sam Goodwin** in https://github.com/alchemy-run/alchemy/issues/1094 [<samp>(8a2e3)</samp>](https://github.com/alchemy-run/alchemy/commit/8a2e3463)
+  - Deterministic force to support monorepo &nbsp;-&nbsp; by **Sam Goodwin** in https://github.com/alchemy-run/alchemy/issues/1105 [<samp>(ccf59)</samp>](https://github.com/alchemy-run/alchemy/commit/ccf59ae9)
+- **docker**:
+  - Don’t quote docker args &nbsp;-&nbsp; by **Andrew Jefferson** in https://github.com/alchemy-run/alchemy/issues/1095 [<samp>(d1cf3)</samp>](https://github.com/alchemy-run/alchemy/commit/d1cf3661)
+
+##### &nbsp;&nbsp;&nbsp;&nbsp;[View changes on GitHub](https://github.com/alchemy-run/alchemy/compare/v0.71.0...v0.71.1)
+
+---
+
 ## v0.71.0
 
 ### &nbsp;&nbsp;&nbsp;🚀 Features
--- a/alchemy-web/src/content/docs/guides/planetscale-drizzle.mdx
+++ b/alchemy-web/src/content/docs/guides/planetscale-drizzle.mdx
@@ -47,12 +47,12 @@ This guide shows how to create and manage [PlanetScale](https://planetscale.com/
    Create a PlanetScale account and get your API credentials. Add them to your `.env` file:
 
    ```bash
-   PLANETSCALE_ORG_ID=your_organization_id
+   PLANETSCALE_ORGANIZATION=your_organization_name
    PLANETSCALE_API_TOKEN=your_api_token
    ```
 
    :::tip
-   You can find your organization ID and create API tokens in your [PlanetScale dashboard](https://app.planetscale.com/).
+   You can find your organization name and create API tokens in your [PlanetScale dashboard](https://app.planetscale.com/).
    :::
 
 3. **Define schema**
@@ -115,7 +115,6 @@ This guide shows how to create and manage [PlanetScale](https://planetscale.com/
    const database = await Database("Database", {
      adopt: true,
      name: "sample-database",
-     organizationId: process.env.PLANETSCALE_ORG_ID!,
      region: {
        slug: "us-east",
      },
@@ -132,8 +131,7 @@ This guide shows how to create and manage [PlanetScale](https://planetscale.com/
    const branch = await Branch("Branch", {
      adopt: true,
      name: `${app.name}-${app.stage}-branch`,
-     organizationId: process.env.PLANETSCALE_ORG_ID!,
-     databaseName: database.name,
+     database,
      parentBranch: database.defaultBranch,
      isProduction: false,
      safeMigrations: !app.local,
@@ -142,9 +140,8 @@ This guide shows how to create and manage [PlanetScale](https://planetscale.com/
    // Create credentials for the branch
    const password = await Password("Password", {
      name: `${app.name}-${app.stage}-password`,
-     organizationId: process.env.PLANETSCALE_ORG_ID!,
-     database: database,
-     branch: branch,
+     database,
+     branch,
      role: "admin",
    });
 
--- a/alchemy-web/src/content/docs/guides/planetscale-postgres.mdx
+++ b/alchemy-web/src/content/docs/guides/planetscale-postgres.mdx
@@ -47,14 +47,14 @@ This guide shows how to create and manage [PlanetScale PostgreSQL](https://plane
    Create a PlanetScale account and get your API credentials. Add them to your `.env` file:
 
    ```properties
-   PLANETSCALE_ORG_ID=your_organization_id
+   PLANETSCALE_ORGANIZATION=your_organization_name
    PLANETSCALE_API_TOKEN=your_api_token
    CLOUDFLARE_ACCOUNT_ID=your_cloudflare_account_id
    CLOUDFLARE_API_TOKEN=your_cloudflare_api_token
    ```
 
    :::tip
-   You can find your organization ID and create API tokens in your [PlanetScale dashboard](https://app.planetscale.com/).
+   You can find your organization name and create API tokens in your [PlanetScale dashboard](https://app.planetscale.com/).
    :::
 
 3. **Define schema**
@@ -158,7 +158,6 @@ This guide shows how to create and manage [PlanetScale PostgreSQL](https://plane
    // Create the PostgreSQL database
    const database = await Database("Database", {
      name: "sample-database",
-     organizationId: alchemy.env.PLANETSCALE_ORG_ID,
      clusterSize: "PS_10",
      kind: "postgresql",
    });
--- a/alchemy-web/src/content/docs/providers/planetscale/branch.md
+++ b/alchemy-web/src/content/docs/providers/planetscale/branch.md
@@ -14,8 +14,8 @@ import { Branch } from "alchemy/planetscale";
 
 const devBranch = await Branch("feature-123", {
   name: "feature-123",
-  organizationId: "my-org",
-  databaseName: "my-database",
+  organization: "my-org",
+  database: "my-database",
   parentBranch: "main",
   isProduction: false,
 });
@@ -30,8 +30,8 @@ import { Branch } from "alchemy/planetscale";
 
 const prodBranch = await Branch("production", {
   name: "production",
-  organizationId: "my-org",
-  databaseName: "my-database",
+  organization: "my-org",
+  database: "my-database",
   parentBranch: "main",
   isProduction: true,
   clusterSize: "PS_20",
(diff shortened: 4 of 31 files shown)


eastlondoner:
It works but I am gonna check I haven't broken anything and write an example in the morning

alchemy/src/cloudflare/miniflare/miniflare-worker-proxy.ts
```diff
@@ -46,9 +46,8 @@ export async function createMiniflareWorkerProxy(options: {
     req: http.IncomingMessage,
   ): Promise<miniflare.Response> => {
     const info = parseIncomingMessage(req);
-    options.transformRequest?.(info);
-
     const name = options.getWorkerName(info);
+    options.transformRequest?.(info);
```
Mkassabov, line 50: why did this get moved? I'd prefer to keep the diff cleaner if this change isn't necessary

alchemy/src/cloudflare/miniflare/tunnel.ts
```diff
@@ -17,19 +19,72 @@ import {
...
 
-export async function createTunnel(
+export async function createNamedTunnel(
+  miniflare: miniflare.Miniflare,
+  tunnel: DevTunnelRoute,
+): Promise<Tunnel> {
+  // maps hostname to worker name
+  const workers = new Map<string, string>(); // used to avoid exposing workers that are not running with tunneling enabled
+  const proxy = await createMiniflareWorkerProxy({
+    port: await findOpenPort(9977), // alchemy auth uses 9976, so one above that
```
Mkassabov: this comment seems unnecessary to be in the codebase.

alchemy/src/cloudflare/miniflare/tunnel.ts
```diff
@@ -17,19 +19,72 @@ import {
...
+      request.headers.set("host", proxy.url.host);
+    },
+    getWorkerName: (request) => {
+      const hostname = request.url.hostname;
+      if (!hostname) {
+        throw new MiniflareWorkerProxyError(
+          "Hostname is missing from request URL. This indicates a bug in Alchemy.",
+          530,
+        );
+      }
```
Mkassabov, line 59: can we elaborate on what this actually means? If we know it can be a bug why not fix it?

(I am aware it exists in the current `createTunnel` implementation)

alchemy/src/cloudflare/miniflare/tunnel.ts
```diff
@@ -17,19 +19,72 @@ import {
...
+      return name;
+    },
+  });
+  tunnel.tunnel.setPort(proxy.url.port);
+  const realTunnel = await TunnelResource(tunnel.tunnel.id, tunnel.tunnel);
+  await Scope.current.spawn("tunnel", {
+    processName: `cloudflared-${tunnel.tunnel.name}`,
+    cmd: `cloudflared tunnel run --token ${realTunnel.token.unencrypted}`,
+    quiet: !process.env.DEBUG,
+  });
```
Mkassabov, line 76: what happens if `cloudflared` isn't installed

alchemy/src/cloudflare/miniflare/tunnel.ts
```diff
@@ -41,6 +96,7 @@ export async function createTunnel(
       if (request.url.origin === remote.origin) {
         request.url.protocol = proxy.url.protocol;
         request.url.host = proxy.url.host;
+        request.url.hostname = proxy.url.hostname;
```
Mkassabov, line 99: why is this change for the trycloudflare version needed?

alchemy/src/cloudflare/dev-tunnel.ts
```diff
@@ -0,0 +1,116 @@
...
+      service: "http_status:404",
+    },
+  ];
+
+  function addRoute(props: DevTunnel<Hostnames>, hostname: Hostnames[number]) {
+    if (ingress.some((rule) => rule.hostname === hostname)) {
+      throw new Error(`Route ${hostname} already exists`);
+    }
+    ingress.unshift({
+      service: port ? `http://localhost:${port}` : "LOCAL_WORKER_PLACEHOLDER",
```
Mkassabov: what is `"LOCAL_WORKER_PLACEHOLDER"`?

alchemy/src/cloudflare/dev-tunnel.ts
```diff
@@ -0,0 +1,116 @@
...
+      devTunnelProps,
+    )) as unknown as DevTunnel<Hostnames>;
+    // TBH I don't think this is needed but just in case
+    tunnel.hostnames = props.hostnames as Hostnames;
+    tunnel.name = props.name;
+    tunnel.id = id;
+    tunnel.configureWorker = configureWorker;
+    tunnel.addRoute = (hostname: string) => addRoute(tunnel, hostname);
+    return tunnel as DevTunnel<Hostnames>;
+  }
```
Mkassabov, line 124: why does this destroy the tunnel?

alchemy/src/cloudflare/miniflare/miniflare-worker-proxy.ts
eastlondoner: this is necessary, it doesn't make a difference to the existing proxy routing which order they go in. But for the named tunnel routing it does make a difference 

The transformRequest method re-writes the url - with named tunnels routing for multiple workers need to use the original incoming url to know which worker it's supposed to be in `getWorkerName` so `getWorkerName` has to run before the request is transformed which overwrites the original incoming url.

alchemy/src/cloudflare/miniflare/tunnel.ts
eastlondoner: this would mean the incoming request somehow has an invalid url with empty or no hostname, this basically exists to placate typescript

alchemy/src/cloudflare/miniflare/tunnel.ts
eastlondoner: same as existing quick tunnels, you get an error thrown here

alchemy/src/cloudflare/miniflare/tunnel.ts
eastlondoner: this isn't needed I put it in as a nit/cleanup. if you're overwriting the host property on url you should also overwrite the hostname property. Many people/AI don't realise that there are two properties and they're slightly different.

alchemy/src/cloudflare/dev-tunnel.ts
eastlondoner: it's a placeholder!

because addRoute can be called before or after we know what local port the worker we create for tunnels in miniflare is listening on. If it's called before we know which port we set LOCAL_WORKER_PLACEHOLDER. When the port is set we overwrite it with `http://localhost:${port}` on line 88

alchemy/src/cloudflare/dev-tunnel.ts
eastlondoner: the tunnel is an actual resource that exists in Cloudflare Access (with ingress rules attached) so if someone runs delete we clean up the resource, otherwise it'd leak and people would still have these tunnels lying around in their Cloudflare account and dashboard

Mkassabov:
running the `cloudflare-dev-tunnel-named` I get this
```
🚀 Dev Tunnel Example Running!
──────────────────────────────────────────────────
API Worker:  https://api-dev.alchemy-test.us/
Web Worker:  https://api-dev.alchemy-test.us/
──────────────────────────────────────────────────
```

looking at the `alchemy.run.ts` file I expect the web worker to be running on `web-dev.alchemy-test.us` so it seems there is an issue somewhere.

other than that issue the example does work

Mkassabov:
> looking at the `alchemy.run.ts` file I expect the web worker to be running on `web-dev.alchemy-test.us` so it seems there is an issue somewhere.

can confirm this is fixed

alchemy/src/cloudflare/miniflare/miniflare-controller.ts
```diff
@@ -46,15 +46,25 @@ export class MiniflareController {
     this.options.set(input.name, first.value);
     const miniflare = await this.update();
     let url: URL;
+    const port = input.port ?? (await findOpenPort());
```
sam-goodwin, line 49: Not changed by this PR but I hope we can find better solution to `findOpenPort` because it's not thread-safe.

alchemy-web/src/content/docs/concepts/dev.mdx
```diff
@@ -158,6 +162,47 @@ You must download and install `cloudflared` - see the [Cloudflare Tunnels docume
...
+  hostnames: ["api-dev.example.com", "web-dev.example.com"] as const,
+});
+
+// API worker on api-dev.example.com
+const apiWorker = await Worker("api", {
+  entrypoint: "./src/api-worker.ts",
+  dev: {
+    tunnel: devTunnel.addRoute("api-dev.example.com"),
+  },
+});
```
sam-goodwin, line 184: Interesting, we're introducing our first imperative API to alchemy. Why do the host names need to be specified upfront and then consumed with `addRoute`?

alchemy-web/src/content/docs/providers/cloudflare/tunnel.md
```diff
@@ -5,6 +5,10 @@ description: Connect private services securely to the internet without exposing
 
 A [Cloudflare Tunnel](https://developers.cloudflare.com/cloudflare-one/connections/connect-networks/) provides a secure connection between your origin server and Cloudflare's global network without exposing your server's IP address.
 
+:::tip[Local Development]
+For routing public traffic to your local development environment, use [`DevTunnel`](/providers/cloudflare/dev-tunnel/) instead. DevTunnel integrates with [`alchemy dev`](/concepts/dev/) and automatically manages `cloudflared` for you. See the [Local Development guide](/concepts/dev/#tunnel) for more information.
+:::
```
sam-goodwin, line 10: Is this right? I am not sure we should be recommending DevTunnel as the way to do this over quick tunnels because DevTunnel requires DNS.

alchemy/src/cloudflare/dev-tunnel.ts
```diff
@@ -0,0 +1,127 @@
...
+          rule.service = `http://localhost:${port}`;
+        }
+      }
+    },
+    configureWorker,
+    configSrc: "cloudflare",
+    ingress,
+    addRoute(hostname: Hostnames[number]) {
+      return addRoute(devTunnelProps, hostname);
+    },
```
sam-goodwin, line 108: This won't work. Remember that all inputs and outputs of a Resource need to be serializable and closures are not serializable.

Instead, keep the resource private and wrap in a function.

```ts
async function DevTunnel(id: string, props: ..) {
  return {
    ...(await _DevTunnel(id, props)),
    addRoute(hostname) { .. }
  }
}
```

alchemy/src/cloudflare/dev-tunnel.ts
```diff
@@ -0,0 +1,127 @@
...
+  };
+
+  if (scope.phase === "destroy") {
+    // this ensures the tunnel resource is deleted, no further information is needed
+    const tunnel = (await Tunnel(
+      id,
+      devTunnelProps,
+    )) as unknown as DevTunnel<Hostnames>;
```
sam-goodwin, line 116: Why create the resource in the destroy phase? Just omit it?

alchemy/src/cloudflare/dev-tunnel.ts
sam-goodwin: Does DevTunnel need to be a resource? Can it just be a function?

alchemy/src/cloudflare/dev-tunnel.ts
sam-goodwin: I am confused. When is the Tunnel resource created? We only create it during `"destroy"`?

eastlondoner:
voice chat with Sam concluded that we should try a different approach - allow users to declare a `workerName` and a `domain` on the tunnel. Those could be used ephemerally with the quick tunnel in mini flare-controller. That way the 


```
const worker = await Worker('my-worker', {
  name: "worker"
  dev: {
    tunnel: {
      workerName: "worker-dev-relay"
      route: "dev.my-domain.com/a/path" // route is optional
    }
  }
})
```

This is much better

eastlondoner:
I am very sorry, I have been quite stupid with the dev tunnel stuff.

I realised I can do it just fine with the existing tunnel resource and without touching the dev tunnel stuff.
All I have to do is

```
const agentWorker = await Worker('agent-worker', {
  ...ordinary worker spec
})

const dynmapWorker = await Worker('dynmap-worker', {
  ...ordinary worker spec
})

const devTunnel = await Tunnel("mineflare-agent-dev-tunnel", {
  name: `${app.name}-agent-dev`,
  adopt: true,
  ingress: [
    {
      service: agentWorker.url!.replace(/\/$/, ""), // replace becase Cloudflare tunnels like http://localhost:3000 but not http://localhost:3000/
      hostname: "mineflare-agent-dev.mydomain.app",
    },
    {
      service: dynmapWorker.url!.replace(/\/$/, ""),
      hostname: "mineflare-dynmap-dev.mydomain.app",
    },
    {
      service: "http_status:404",
    }
  ]
});
```
mentally I had pre-determined that
- The cloudflare tunnel should be set up before the workers
- The tunnel had to work like the quick tunnels by being embedded inside the miniflare controller so it was only used in dev
Alchemy already had all the right primitives to do what I wanted really quite neatly, I just had the wrong mental models

PR #1108, opened 2025-10-10
fix(cloudflare): pass all ContainerProps to ContainerApplication

eastlondoner:
For example maxInstances was not being passed through so changing it had no effect

### DIFF
--- a/alchemy/src/cloudflare/redwood/redwood.ts
+++ b/alchemy/src/cloudflare/redwood/redwood.ts
@@ -50,9 +50,9 @@ export async function Redwood<B extends Bindings>(
       },
     },
     noBundle: props?.noBundle ?? true,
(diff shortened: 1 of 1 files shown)


JacobMGEvans:
For fixes/bugs could you please write a test verifying fix and preventing regressions 😃

eastlondoner:
@JacobMGEvans could you do me a favour and point at an existing example containing of the kind of test  that’s appropriate and I can take it from there

JacobMGEvans:
> @JacobMGEvans could you do me a favour and point at an existing example containing of the kind of test  that’s appropriate and I can take it from there

@eastlondoner would love too... Unfortunately Im still relatively new to the codebase and still learning parts. My guess would be to check against an object that reflects all the prop types and negative check would be for the previous behavior you fixed

PR #1113, opened 2025-10-11
fix(cloudflare): reference @cloudflare/workers-types

sam-goodwin:


eastlondoner:
HALLELUJA

I confirm it's fixed my type problems

PR #1118, opened 2025-10-14
feat(planetscale): allow planetscale resources to be not deleted

### LINKED ISSUE #1117: implement `delete` prop for all planetscale resources
currently there is no `delete` option of any of the Planetscale related resources
on cloudflare resources there is a delete prop, which when set to false, destory doesn't remove the db via api, only deletes it from the state
we want same thing for Database/Branch/Password/Role resources in planetscale, to make sure we dont accidently nuke our prod db

@claude implement this

BlankParticle:
fixes #1117


alchemy/src/planetscale/role.ts
```diff
@@ -44,6 +44,21 @@ export interface RoleProps extends PlanetScaleProps {
...
    */
   inheritedRoles: InheritedRole[] | Role;
+
+  /**
+   * Whether to adopt an existing role if it exists.
+   * If false and a role with similar permissions exists, an error will be thrown.
+   * If true and a role with similar permissions exists, it will be updated with the provided properties.
+   * @default false
+   */
+  adopt?: boolean;
```
Mkassabov: added in props but not implemented? I think its fine to just remove this, IDK if adoption here is that important.

especially not the way the comment outlines
>  If false and a role with similar permissions exists, an error will be thrown.

that would be crazy behavior

alchemy/test/planetscale/branch.test.ts
```diff
@@ -236,7 +236,7 @@ describe.skipIf(!process.env.PLANETSCALE_TEST).concurrent.each(kinds)(
         });
         expect(response.status).toEqual(404);
       }
-    }, 300_000);
+    }, 1_000_000);
```
Mkassabov, line 239: @BlankParticle 

not super relevant to this pr, we are just slowly trying to dial in better timeouts for planetscale since it can be quite slow; nothing you did caused this change. Just easier to change it now as I'm running tests.

alchemy/src/planetscale/role.ts
BlankParticle: oh I started implementing that, but found out you cant get the password back, will remove

alchemy-web/src/content/docs/providers/planetscale/branch.md
```diff
@@ -21,6 +21,23 @@ const devBranch = await Branch("feature-123", {
 });
 ```
 
+## Deletion
+
+By default, when a branch is deleted, the branch will be removed from the state but not deleted via API. This is to prevent accidental loss of data. This setting can be changed by setting the `delete` property to `false`.
```
BlankParticle: it should say setting delete to true, false is default

alchemy-web/src/content/docs/providers/planetscale/index.md
```diff
@@ -14,6 +14,21 @@ PlanetScale is a serverless database platform that supports both MySQL and Postg
 - [Password](/providers/planetscale/password) - Create and manage database passwords with specific roles and permissions (MySQL only)
 - [Role](/providers/planetscale/role) - Create and manage database roles with inherited permissions (PostgreSQL only)
 
+## Authentication
+
+PlanetScale authentication is handled via environment variables.
+
+### Setting the token id and secret
```
JacobMGEvans: Do we want to make I.D. all caps in the docs? 
```suggestion
### Setting the token ID and secret
```

alchemy-web/src/content/docs/providers/planetscale/index.md
```diff
@@ -14,6 +14,21 @@ PlanetScale is a serverless database platform that supports both MySQL and Postg
...
+The token id and secret can be set by setting the `PLANETSCALE_SERVICE_TOKEN_ID` and `PLANETSCALE_SERVICE_TOKEN` environment variables. Alchemy will use those to generate the correct header for the planetscale api.
```
JacobMGEvans:
```suggestion
The token ID and secret can be set by setting the `PLANETSCALE_SERVICE_TOKEN_ID` and `PLANETSCALE_SERVICE_TOKEN` environment variables. Alchemy will use those to generate the correct header for the planetscale api.
```

alchemy-web/src/content/docs/providers/planetscale/index.md
```diff
@@ -14,6 +14,21 @@ PlanetScale is a serverless database platform that supports both MySQL and Postg
...
+
+### Directly setting the api token
+
+The api token can be set directly by setting the `PLANETSCALE_API_TOKEN` environment variable. This must follow the format outlined by planetscale here: [Authentication](https://planetscale.com/docs/concepts/authentication).
```
JacobMGEvans: Same with other acronyms are we going keep them uppercase? 
```suggestion
The API token can be set directly by setting the `PLANETSCALE_API_TOKEN` environment variable. This must follow the format outlined by planetscale here: [Authentication](https://planetscale.com/docs/concepts/authentication).
```

alchemy-web/src/content/docs/providers/planetscale/organization.md
```diff
@@ -0,0 +1,22 @@
+---
+title: OrganizationRef
+description: Learn how to manage PlanetScale Organizations using Alchemy.
+---
+
+For ease of use we expose `OrganizationRef` to allow you to reference an organization by name.
+
+## Minimal Example
+
+An organization can be referenced by name or by id, then passed to other planetscale resources.
```
JacobMGEvans:
```suggestion
An organization can be referenced by name or by ID, then passed to other planetscale resources.
```

alchemy-web/src/content/docs/providers/planetscale/index.md
```diff
@@ -14,6 +14,21 @@ PlanetScale is a serverless database platform that supports both MySQL and Postg
...
 - [Role](/providers/planetscale/role) - Create and manage database roles with inherited permissions (PostgreSQL only)
 
+## Authentication
+
+PlanetScale authentication is handled via environment variables.
+
+### Setting the token id and secret
+The token id and secret can be set by setting the `PLANETSCALE_SERVICE_TOKEN_ID` and `PLANETSCALE_SERVICE_TOKEN` environment variables. Alchemy will use those to generate the correct header for the planetscale api.
+
+### Directly setting the api token
```
JacobMGEvans:
```suggestion
### Directly setting the API token
```

PR #1121, opened 2025-10-14
fix(core): error on mismatch between options.phase and CLI args

### LINKED ISSUE #977: Config Management Proposal
John, Jacob, and I were talking about some stuff RE credential management.

This started with Jacob having both env variables and cli arguments that conflicted, and alchemy doesn't handle that edge case.

After a chat we decided on the following:
- allow params from anywhere (code params, scope, cli agruments, env variables, etc)
- check all sources for params, if there is more than 1 defined and the definitions don't match throw an error
  - only exception is credentials on resource (in order to support orchestrating resources across multiple accounts on the same provider, e.g. workers in 2 different cloudflare accounts)

This would apply both to credentials (api tokens, account ids, api keys, etc) AND general alchemy params (alchemy password, phase, stage, etc)

JacobMGEvans:
When the user intent is to run in development mode, implied by CLI dev or dev flag, then stages set to "up" should error rather than ignoring CLI usage and create the resources 

Fixes #977


JacobMGEvans:
Needs tests.

alchemy/src/alchemy.ts
```diff
@@ -162,6 +162,22 @@ If this is a mistake, you can disable this check by setting the ALCHEMY_CI_STATE
 `);
   }
 
+  if (mergedOptions.local && mergedOptions.stage === "up") {
```
john-royal: I think you mixed up `stage` and `phase`. Phase is `up`/`destroy`/`read`

JacobMGEvans: Thanks for catching this!

PR #1126, opened 2025-10-15
chore(SEO) Improve SEO with JSON + LD

JacobMGEvans:
Rich Results: Enables breadcrumbs in search results, knowledge panels, rich cards
Semantic Understanding: Helps Google/Bing understand content hierarchy and relationships
Documentation-Specific: TechArticle schema specifically designed for technical docs
Competitive Advantage: Most documentation sites lack proper structured data
Future-Proof: AI search engines (ChatGPT, Perplexity) increasingly rely on structured data


hacknug:
@JacobMGEvans I made some changes:
- added `astro-seo-schema` and `schema-dts` to simplify extremely little the way we set this up and ensure typesafety
- replaced hardcoded `https://alchemy.run` with `Astro.url.origin` so we can test things outside of production and we don't need to update in case of changing the site's url
- fixed the link generated for our pages og image (it sometimes had one two many dashes and was stripping all slashes)
- added an extra `ListItem` for the current page without `item` property as recommended in [yoast's docs site](https://developer.yoast.com/features/schema/pieces/breadcrumb/#:~:text=@type%20:%20BreadcrumbList%20.,of%20the%20page%20in%20question.)

Let me know if you'd like me to make any changes to this. Not sure we can test it live since the preview was deployed before we fixed the proxy/router issue yesterday.

PR #1127, opened 2025-10-15
fix(cloudflare): auto-detect server entry for tanstack start

john-royal:
TanStack Start allows an optional server entrypoint located at `src/server.ts` (see [docs](https://tanstack.com/start/latest/docs/framework/react/guide/server-entry-point)). This must be set in the `wrangler.jsonc`, otherwise it won't work.


JacobMGEvans (approved):
~Please write tests~
Tests are hard for this.

PR #1133, opened 2025-10-16
chore: pathe everywhere

Mkassabov:
- moves pathe to the catalog
- moves alchemy-web to pathe
- moves scripts to pathe
- adds a biome rule to ban imports of `node:path`
- replaces all `node:path` instances with `pathe`

- also added prettier as a dev dep because something was using it (that should probably be refactored, but whatever)


JacobMGEvans (approved):

Mkassabov:
between oxfmt and andy's changes this is redundant. the remaining node:paths are mostly docs and a few old scripts we probably shouldn't touch

PR #1134, opened 2025-10-16
chore(docs): generate llms.txt files with website contents

hacknug:
This PR generates a few new txt files to help llms consume the contents of our website, complementing the individual markdown files we serve for each document.

It uses the `starlight-llms-txt` plugin with a little config to make the order better match our sidebar (although with some differences). You can check the available options here: https://delucis.github.io/starlight-llms-txt/configuration/

I excluded the `telemetry` dir but maybe we should just demote it to the end since it is the only place where we mention how to disable telemetry?

### DIFF
--- a/alchemy-web/astro.config.mjs
+++ b/alchemy-web/astro.config.mjs
@@ -6,6 +6,7 @@ import icon from "astro-icon";
 import { defineConfig } from "astro/config";
 import starlightBlog from "starlight-blog";
 import starlightLinksValidator from "starlight-links-validator";
+import starlightLlmsTxt from "starlight-llms-txt";
 import theme from "starlight-theme-nova";
 
 // @ts-expect-error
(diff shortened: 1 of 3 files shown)


JacobMGEvans (approved):

alchemy-web/astro.config.mjs
```diff
@@ -117,6 +118,18 @@ export default defineConfig({
...
+          promote: [
+            "index",
+            "what-is-alchemy",
+            "getting-started",
+            "!*/*", // pages at the root dir
+            "concepts/**",
+            "guides/**",
+          ],
+          demote: ["providers/**", "blog/**"],
+          exclude: ["telemetry/**"],
```
sam-goodwin, line 131: Have you checked the size of the `llms.txt`? When i tested it, it was a million tokens because of `providers/aws/control`.

We should exclude the aws/control docs and check if the LLMs.txt is usable/not too big.

Is there a way to provide hierarchical llms.txt?

Mkassabov:
yeah this is not viable like this. full returns over a million tokens, can't fit in any model context, let alone inside the useful part of it.

PR #1138, opened 2025-10-17
fix(cloudflare): Get the tunnel token when it's not returned by the first API call

eastlondoner:
This fix from my dev tunnel work never actually made it to main and it is necessary to use tunnels reliably as documented

### DIFF
--- a/alchemy/src/cloudflare/tunnel.ts
+++ b/alchemy/src/cloudflare/tunnel.ts
@@ -566,7 +566,6 @@ export const Tunnel = Resource(
           (error.message.includes("already have a tunnel with this name") ||
             error.message.includes("already exists"))
         ) {
-          console.log(error);
           logger.log(`Tunnel '${name}' already exists, adopting it`);
 
           // Find the existing tunnel by name


alchemy/src/cloudflare/tunnel.ts
```diff
@@ -596,6 +598,12 @@ export const Tunnel = Resource(
       }
     }
 
+    // Ensure tunnel token is available (unless we're replacing)
+    // Sometimes Cloudflare doesn't return the token in the initial response
```
sam-goodwin, line 602: That is very strange lol

alchemy/src/cloudflare/tunnel.ts
```diff
@@ -918,3 +926,30 @@ export async function findTunnelByName(
...
+  if (!response.ok) {
+    logger.error(
+      `Failed to fetch token for tunnel ${tunnelId}: ${response.status} ${response.statusText}`,
+    );
+    await handleApiError(response, "get token", "tunnel", tunnelId);
+  }
+
+  const data = (await response.json()) as CloudflareApiResponse<string>;
+  if (!data.result) {
+    logger.warn(`Could not retrieve token for tunnel ${tunnelId}`);
```
sam-goodwin, line 950: Is it not an error?

JacobMGEvans: Not if you are immediately trying to get it again, right?

sam-goodwin: The type looks wrong. We are returning `data.result` when `!data.result`. So it isn't guaranteed to be a `string`. The `CloudflareApiResponse<string>` casts it. Is it possible for it to return `undefined`?

alchemy/src/cloudflare/bun-spa/bun-spa.ts
```diff
@@ -86,7 +86,7 @@ export async function BunSPA<B extends Bindings>(
     },
     build: spreadBuildProps(
       props,
-      `bun build '${frontendPaths.join("' '")}' --target=browser --minify --define:process.env.NODE_ENV='"${nodeEnv}"' --env='BUN_PUBLIC_* --outdir ${outDir}`,
+      `bun build '${frontendPaths.join("' '")}' --target=browser --minify --define:process.env.NODE_ENV='"${nodeEnv}"' --env='BUN_PUBLIC_*' --outdir ${outDir}`,
```
sam-goodwin: I just fixed this :) https://github.com/alchemy-run/alchemy/commit/b03ce2423d410021055de9e81153137bab9d93b1

alchemy/src/cloudflare/tunnel.ts
eastlondoner: I think this is claude being defensive but it's not any worse than the current behaviour.

Throwing an error feels over the top since people might not actually care about the token (depending on how they're using the tunnel).

Can we merge this? Or if not how would you like this changed?

alchemy/src/cloudflare/bun-spa/bun-spa.ts
eastlondoner: I don't know how this sneaked through the last PR

alchemy/src/cloudflare/tunnel.ts
sam-goodwin: Let's merge

PR #1139, opened 2025-10-17
fix(cloudflare): insert kv records to miniflare

john-royal:


alchemy/src/cloudflare/kv-namespace.ts
```diff
@@ -340,31 +349,18 @@ export async function insertKVRecords(
...
-        if (entry.expirationTtl) {
-          item.expiration_ttl = entry.expirationTtl;
-        }
-
-        if (entry.metadata) {
-          item.metadata = entry.metadata;
-        }
-
-        return item;
-      });
```
john-royal, line 367: Apologies for the Bonus Refactor™

JacobMGEvans (approved):

PR #1143, opened 2025-10-18
fix(cloudflare): use workspace root for persist path

john-royal:


alchemy/src/cloudflare/miniflare/paths.ts
```diff
@@ -1,6 +1,8 @@
 import { AsyncLocalStorage } from "node:async_hooks";
-import { existsSync, readlinkSync, statSync } from "node:fs";
+// biome-ignore lint/style/noRestrictedImports: I/O is acceptable here
```
sam-goodwin, line 2: you don't need biome ignores anymore

PR #1154, opened 2025-10-19
chore(agents): Improve Contributing and Agents docs

JacobMGEvans:


AGENTS.md
```diff
@@ -516,6 +401,300 @@ export const MyResource = Resource(
 );
 ```
 
+### Advanced Resource Patterns
+
+#### Input Normalization with Wrapper Functions
+
+When resources accept flexible input types (e.g., `string | Secret`, `string | Resource`), use a public wrapper function to normalize inputs before passing to the internal Resource:
+
+```ts
```
Mkassabov, line 412: idk if we always want to do this, seems a little extreme to do it if its only 1 input type

AGENTS.md
```diff
@@ -516,6 +401,300 @@ export const MyResource = Resource(
...
+  ?? this.scope.createPhysicalName(id);  // Default: {app}-{stage}-{id}
+```
+
+#### Resource Replacement for Immutable Properties
+
+When an immutable property changes, signal replacement via `this.replace()`:
+
+```ts
+if (this.phase === "update" && this.output.name !== name) {
+  logger.log(`Immutable property 'name' changed, replacing resource`);
```
Mkassabov: we don't want to do this logger.log call. we already have logs that say a resource is being replaced so this log would just cause clutter

AGENTS.md
```diff
@@ -516,6 +401,300 @@ export const MyResource = Resource(
...
+  if (props.delete !== false && this.output?.id) {
+    try {
+      await api.delete(`/resources/${this.output.id}`);
+    } catch (error) {
+      if (error.status !== 404) throw error; // OK if already deleted
+    }
+  }
+  return this.destroy();
+}
+```
```
Mkassabov, line 546: this is only something we do for data resources, it doesn't make much sense for compute

AGENTS.md
```diff
@@ -516,6 +401,300 @@ export const MyResource = Resource(
...
+
+  /**
+   * Whether to adopt an existing resource
+   * @default false
+   */
+  adopt?: boolean;
+
+  // Internal properties for lifecycle management
+  /** @internal */
+  resourceId?: string;
```
Mkassabov: internal doesn't make much sense being here, if we expose it to the user it isn't internal

AGENTS.md
```diff
@@ -516,6 +401,300 @@ export const MyResource = Resource(
...
+   */
+  secret: Secret;
+
+  /**
+   * Resource type identifier for binding
+   * @internal
+   */
+  type: "my-resource";
+};
+```
```
Mkassabov, line 700: is this not duplcated of `Output Type Pattern with Omit` section?

AGENTS.md
JacobMGEvans: I can try to make it more clear when its NOT necessary.

AGENTS.md
JacobMGEvans: Gotcha!

AGENTS.md
JacobMGEvans: So only for things like DB's we wanna make it optional to fully delete them or not, but for compute we just nuke them without the optional field. I think I understand why

AGENTS.md
JacobMGEvans: Yeah, I will fix the one in Hyperdrive too

PR #1155, opened 2025-10-19
fix(util): add base case to findWorkspaceRoot and findWorkspaceRootSync

nbbaier:


JacobMGEvans:
Hey appreciate the effort and work! @john-royal was able to get a fix in with PR #1156 

Thanks again!!

nbbaier:
Oh damn was too slow! But glad this got fixed!

PR #1156, opened 2025-10-19
fix: improve workspace root detection

### LINKED ISSUE #1150: `alchemy dev` and `alchemy destroy` hang when running on `alchemy@0.75.1`
I noticed an odd difference between `alchemy@0.75.0` and `alchemy@0.75.1`. With `0.75.0`, `alchemy dev/deploy/destroy` all behave as expected. With `0.75.1`, `alchemy deploy` works as expected, but `alchemy dev/destroy` do not. These commands hang - I let them both run for over a minute with no results.

Furthermore, while preparing this issue, I discovered that this behavior **only occurs when the project is not a git repo**. When `.git` exists, everything works as expected in `0.75.1`. When `.git` **does not** exist,  `alchemy dev/destroy` hang as described above.

Repo w/ minimal repro: https://github.com/nbbaier/repro-alchemy-hangs

To reproduce, clone the repo locally and delete `.git`. Then run `bun alchemy dev` to see that the process hangs.

--- 
NOTE: See [this post](https://discord.com/channels/1359694195782320389/1359694196830765059/1429327800577425450) in the discord for my original question, which includes screenshots.

john-royal:
Closes #1150.

Changes:
- Detect lockfiles (e.g. `bun.lock`, `package-lock.json`)
- Detect `turbo.json`, but only if it's the root one ([docs](https://turborepo.com/docs/reference/configuration#extends))

Implementation:
- `predicates` is a mapping of file names to a predicate — either `true` if we only care if the file exists, or a function that returns a boolean if we want to check the file contents

To-do:
- [x] Does this fix #1150?

## Summary by CodeRabbit

* **New Features**
  * Added asynchronous workspace root detection capability to the workspace utilities.


john-royal:
Review me please @coderabbitai

PR #1157, opened 2025-10-19
chore(type): removed internal from hyperdriveId

JacobMGEvans:


Mkassabov (approved):

alchemy/src/cloudflare/hyperdrive.ts
```diff
@@ -167,7 +167,6 @@ export interface HyperdriveProps extends CloudflareApiOptions {
   /**
    * UUID of the hyperdrive (only used for update/delete operations)
    * This is provided by Cloudflare and is different from the resource ID
-   * @internal
    */
   hyperdriveId?: string;
```
Mkassabov, line 171: rename this to just `id`

JacobMGEvans: No.

PR #1159, opened 2025-10-19
fix(cloudflare): avoid fetching settings when inferring zone id

john-royal:
The `getZoneByDomain` function currently also fetches the zone settings after fetching the zone. This can cause errors when using an OAuth token for authentication — we can read zones but not zone settings for some reason.

It turns out that we don't actually need the zone settings to be returned by `getZoneByDomain` — the function is only referenced by `inferZoneIdFromPattern`:

image

## Summary by CodeRabbit

* **Breaking Changes**
  * Updated Cloudflare zone lookup API - now returns `undefined` instead of `null` for non-existent zones.

* **Refactor**
  * Enhanced zone data structure with improved type definitions and clearer return type handling.


john-royal:
Review me please @coderabbitai

sam-goodwin (approved):

JacobMGEvans (approved):
Oh this is AWESOME, I was thinking of making all the Resources that take in a Zone to derive Zone ID, to also be able to infer it from domain name, but I was concerned about too many or too large of API requests for inference

PR #1170, opened 2025-10-23
fix(cloudflare): prevent esbuild errors from crashing reload

Mkassabov:
~~image
A little but unhappy that it logs twice but IDK where the original one logs from and it doesn't say the resource it came from.~~

changed it to this but idk if I like it either
image


alchemy/src/cloudflare/worker-bundle.ts
```diff
@@ -1,4 +1,4 @@
-import esbuild from "esbuild";
+import esbuild, { formatMessages } from "esbuild";
```
john-royal: Is this import used?

alchemy/src/cloudflare/worker-bundle.ts
```diff
@@ -308,6 +308,12 @@ export namespace WorkerBundleSource {
...
       for await (const result of hotReload.iterator) {
         count++;
+        if (result.errors.length > 0) {
+          logger.error(
+            `Error bundling worker \`${this.props.id}\`, see detailed logs above.`,
+          );
+          continue;
+        }
         const { entrypoint, root, modules } = await this.formatBuildOutput(
           result.metafile!,
```
john-royal: Nit: Might be worth changing this to `result.metafile` (removing the `!`) and changing the check from `result.errors.length > 0` to `result.errors.length > 0 || !result.metafile`? Seems like the fatal error is coming specifically from trying to read a null metafile after a build error.

john-royal (approved):
LGTM! Just some nitpicking

alchemy/src/cloudflare/worker-bundle.ts
Mkassabov: nah it was before I removed the error

alchemy/src/cloudflare/worker-bundle.ts
Mkassabov: good catch

PR #1171, opened 2025-10-23
feat(cloudflare): support cloudflare access for remote bindings

john-royal:
Remote bindings were failing for a user because their domain was protected by Cloudflare Access. This adds support by using `cloudflared` to fetch the access token if necessary.


alchemy/src/cloudflare/miniflare/remote-binding-proxy.ts
```diff
@@ -171,3 +194,61 @@ async function createWorkersPreviewSession(api: CloudflareApi) {
...
+});
+
+/**
+ * Returns true if the domain uses Cloudflare Access.
+ */
+async function domainUsesAccess(hostname: string) {
+  try {
+    const response = await fetch(`https://${hostname}`, {
+      redirect: "manual",
+      signal: AbortSignal.timeout(1000),
```
john-royal, line 235: I don't fully understand why the timeout is necessary, but [Wrangler does it](https://github.com/cloudflare/workers-sdk/blob/a5eb5134f73d3983655d325a4de71c6370c57faa/packages/wrangler/src/user/access.ts#L23-L31), so I assume it's there for a reason?

alchemy/src/cloudflare/miniflare/remote-binding-proxy.ts
```diff
@@ -162,7 +180,12 @@ async function createWorkersPreviewSession(api: CloudflareApi) {
     "create workers preview session",
     api.get(`/accounts/${api.accountId}/workers/subdomain/edge-preview`),
```
john-royal, line 180: You might be wondering why we didn't include this [zone check](https://github.com/cloudflare/workers-sdk/blob/a5eb5134f73d3983655d325a4de71c6370c57faa/packages/wrangler/src/dev/create-worker-preview.ts#L162-L164):

```ts
ctx.zone
		? `/zones/${ctx.zone}/workers/edge-preview`
		: `/accounts/${accountId}/workers/subdomain/edge-preview`
```

My inclination is to support it in another PR iff it's necessary because it does add some extra complication.

alchemy/src/cloudflare/miniflare/remote-binding-proxy.ts
```diff
@@ -146,8 +159,13 @@ async function createWorkersPreviewToken(
 
 async function prewarm(url: string, previewToken: string) {
   try {
+    const accessToken = await getAccessToken(new URL(url).hostname);
     await fetch(url, {
-      headers: { "cf-workers-preview-token": previewToken },
+      method: "POST",
```
john-royal, line 163: This happens to be [a POST request in Wrangler](https://github.com/cloudflare/workers-sdk/blob/a5eb5134f73d3983655d325a4de71c6370c57faa/packages/wrangler/src/dev/create-worker-preview.ts#L338-L348). Not sure why it wasn't here.

alchemy/src/cloudflare/miniflare/remote-binding-proxy.ts
```diff
@@ -162,7 +180,12 @@ async function createWorkersPreviewSession(api: CloudflareApi) {
     "create workers preview session",
     api.get(`/accounts/${api.accountId}/workers/subdomain/edge-preview`),
   );
-  const res = await fetch(exchange_url);
+  const accessToken = await getAccessToken(new URL(exchange_url).hostname);
```
john-royal, line 182: Why is `getAccessToken` called three times?

1. `exchange_url` - some randomly generated `workers.dev` URL (e.g. `<random-string>.chipotle.workers.dev`), otherwise you get an access error on the domain
2. `prewarm` - seems like the same domain as the `exchange_url`, but can't tell if that's guaranteed; Wrangler just calls `getAccessToken` separately but caches the result, which we do too by wrapping the function in `memoize`
3. The actual worker URL (e.g. `my-worker.chipotle.workers.dev`)

alchemy/src/cloudflare/miniflare/remote-binding-proxy.ts
```diff
@@ -171,3 +194,61 @@ async function createWorkersPreviewSession(api: CloudflareApi) {
...
+ * Otherwise, returns undefined.
+ * @see https://github.com/cloudflare/workers-sdk/blob/a5eb5134f73d3983655d325a4de71c6370c57faa/packages/wrangler/src/user/access.ts#L10
+ */
+const getAccessToken = memoize(async (hostname: string) => {
+  if (!(await domainUsesAccess(hostname))) {
+    return undefined;
+  }
+  const child = spawn("cloudflared", ["access", "login", hostname], {
+    stdio: "pipe",
+  });
```
Mkassabov: is there a reason we dont use scope.spawn

john-royal: The main benefit of `scope.spawn` is that you can spawn a child process and resume it if the Alchemy process restarts — useful for long-running dev servers, but I figured it might be overkill for a one-off command like this one.

Maybe it would be useful if you happen to call this command more than once in a multi-worker setup? Not sure though.

Mkassabov: hmmm I was more thinking that we have better logging for scope.spawn

alchemy/src/cloudflare/miniflare/remote-binding-proxy.ts
```diff
@@ -171,3 +194,61 @@ async function createWorkersPreviewSession(api: CloudflareApi) {
...
+        .then((chunks) => chunks.map((chunk) => chunk.toString()).join("")),
+      once(child, "close"), // throws an error if the command fails
+    ]);
+  } catch (error) {
+    throw new Error(
+      [
+        `The \`cloudflared\` CLI is not installed, but is required to access the domain "${hostname}".`,
+        `Please install it from https://developers.cloudflare.com/cloudflare-one/connections/connect-apps/install-and-setup/installation and run \`cloudflared access login ${hostname}\`.`,
+      ].join("\n"),
+      { cause: error },
```
sam-goodwin: Why are you assumig it's not installed?

alchemy/src/cloudflare/miniflare/remote-binding-proxy.ts
sam-goodwin: Discusssed with @john-royal and he proposed adding `this.exec` that is a simple standard way to execute a short-lived command that returns stdout/stderr/exitcode and logs appropriately, but does not have the complexity of long-running resumable processes like spawn does

alchemy/src/cloudflare/miniflare/remote-binding-proxy.ts
sam-goodwin: Are you asking or explaining?

john-royal: Explaining since I know it looks strange

alchemy/src/cloudflare/miniflare/remote-binding-proxy.ts
sam-goodwin: Maybe the worker can hang?

This is a pretty awful thing to have to do. Are we hitting the user's data plane just to do an access check?

john-royal: Kind of - 2/3 times this function is called, it's hitting a Cloudflare control plane URL. The third time, it could be hitting the user's actual worker to do the access check.

Would it help if we made it a HEAD request?

PR #1172, opened 2025-10-23
fix(cloudflare): environment undefined when using react router with rolldown-vite

john-royal:


alchemy/src/cloudflare/react-router/plugin.ts
```diff
@@ -1,26 +1,37 @@
...
-          name: "dummy",
-          compatibility_date: "2025-09-23",
-        },
-        null,
-        2,
-      ),
-    );
+const DEFAULT_SSR_MAIN_PATH = "../../workers/app.ts";
+
+const alchemy = async (config: PluginConfig = {}) => {
```
sam-goodwin: changing to async? Will this confuse people who aren't used to that in vite.config.ts

alchemy/src/cloudflare/react-router/plugin.ts
```diff
@@ -1,32 +1,21 @@
...
-    );
+const alchemy = (config: PluginConfig = {}) => {
+  // plugin is not required for react-router typegen, so return a no-op to avoid errors
+  if (
+    process.argv.some((arg) => arg.includes("react-router")) &&
+    process.argv.includes("typegen")
+  ) {
+    return {
+      name: "alchemy-noop",
+    };
```
BlankParticle:
```suggestion
const alchemy = (config: PluginConfig = {}): PluginOption => {
  // plugin is not required for react-router typegen, so return a no-op to avoid errors
  if (
    process.argv.some((arg) => arg.includes("react-router")) &&
    process.argv.includes("typegen")
  )
    return null;
```
you can always return `null`, its part of `FalsyPlugin` type of `PluginOption`

sam-goodwin (approved):

PR #1173, opened 2025-10-24
feat(cloudflare): make container bindings work and use official wrangler json spec

BlankParticle:
I started to fix containers not working with vite dev as the generated wrangler json didnt have the binding

found out that the wrangler json spec was written by hand, so instead of trying to figure out what the containers spec look like, I just used the types from wrangler


alchemy/src/cloudflare/wrangler.json.ts
```diff
@@ -162,20 +164,26 @@ export async function WranglerJson(
...
+    await processBindings(
       spec,
       worker.bindings,
       worker.eventSources,
       worker.name,
       cwd,
       props.secrets ?? false,
       Scope.current.local && !props.worker.dev?.remote,
+      async () =>
+        worker.accountId ?? (await createCloudflareApi(worker)).accountId,
```
john-royal, line 182: @BlankParticle I made this slight change so we're more sure that we have the correct accountId for the worker. Let me know if this is okay with you.

alchemy/src/cloudflare/wrangler.json.ts
```diff
@@ -1,12 +1,14 @@
 import fs from "node:fs/promises";
 import path from "pathe";
+import type { Unstable_Config as WranglerJsonConfig } from "wrangler";
```
john-royal, line 3: Good choice! Copying the config object ourselves was annoying.

alchemy/src/cloudflare/wrangler.json.ts
```diff
@@ -1,23 +1,21 @@
...
 import type { EventSource } from "./event-source.ts";
 import { isQueueEventSource } from "./event-source.ts";
 import { isQueue } from "./queue.ts";
 import { unencryptSecrets } from "./util/filter-env-bindings.ts";
 import { isWorker, type Worker, type WorkerProps } from "./worker.ts";
 
-type WranglerJsonRateLimit = Omit<WorkerBindingRateLimit, "type"> & {
-  type: "rate_limit";
-};
-
```
john-royal, line 20: Noticed that this was no longer necessary either since there's `spec.ratelimits` so went ahead and changed that

alchemy/src/cloudflare/wrangler.json.ts
BlankParticle: yea thats fine, I forgot that you can specify it on the worker too

alchemy/src/cloudflare/wrangler.json.ts
BlankParticle: there are still unsafe bindings which don't have any types, but probably no need to support them as those wont be running with vite plugin anytime soon i guess

PR #1176, opened 2025-10-27
add env file to templates

frolic:
I was following the [getting started guide](https://alchemy.run/getting-started/) and running `pnpm dev` for the first time showed an error like
```
Error: Missing token for CloudflareStateStore. Please set ALCHEMY_STATE_TOKEN in the environment or set the `stateToken` option in the CloudflareStateStore constructor. See https://alchemy.run/guides/cloudflare-state-store/
```
even though I had added this to my `.env` file (included in the templates and already .gitignore'd)

Turns out this was because the templates, by default, don't seem to load this env file? I wasn't sure if it was best to add this as a CLI flag to each template or to set `.env` as the default for the CLI flag itself, so went with the former approach for now.


john-royal:
Appreciate the contribution. I wonder if it might be better to just pick up the `.env` file by default if the option isn't set. Will discuss with team today.

john-royal:
This is fixed in #1180

lokhmakov:
@john-royal Started today with template using `bunx alchemy@latest create --template typescript` and the same error

PR #1181, opened 2025-10-28
feat(cloudflare): r2 bucket custom domain

### LINKED ISSUE #1107: R2 Bucket Custom Domain
Add support for attaching a custom domain to a Cloudflare r2 bucket 

https://developers.cloudflare.com/r2/buckets/public-buckets/#custom-domains

Desired experience 
```ts
await R2Bucket("bucket", {
  domain: "sub.domain.com"
});
```

@claude take a stab at this please.

john-royal:
Closes #1107


alchemy-web/src/content/docs/providers/cloudflare/bucket.md
```diff
@@ -49,9 +49,9 @@ const euBucket = await R2Bucket("eu-bucket", {
 });
 ```
 
-## With Public Access
+## With `r2.dev` Domain
 
-Create a development bucket with public access enabled:
+Create a development bucket with the `r2.dev` domain enabled:
```
sam-goodwin, line 54: Is r2.dev the same as Public Access?

alchemy/src/cloudflare/bucket-custom-domain.ts
```diff
@@ -0,0 +1,160 @@
...
+import { Resource } from "../resource";
+import { diff } from "../util/diff";
+import { createCloudflareApi, type CloudflareApiOptions } from "./api";
+import { CloudflareApiError } from "./api-error";
+import { extractCloudflareResult } from "./api-response";
+import { withJurisdiction, type R2BucketJurisdiction } from "./bucket";
+import { inferZoneIdFromPattern } from "./route";
+import type { Zone } from "./zone";
+
+export interface R2BucketCustomDomainProps extends CloudflareApiOptions {
```
sam-goodwin: `delete?: boolean`

alchemy-web/src/content/docs/providers/cloudflare/bucket.md
john-royal: Yes, "public access" is just what we call it for some reason (Claude?). Cloudflare doesn't call it that.

The API route we call is `/accounts/{account_id}/r2/buckets/{bucket_name}/domains/managed`, and it's listed in the docs as "update R2 dev domain of bucket" ([link](https://developers.cloudflare.com/api/resources/r2/subresources/buckets/subresources/domains/subresources/managed/methods/update/)).

Should we rename the parameter?

Mkassabov (approved):

alchemy-web/src/content/docs/providers/cloudflare/bucket.md
sam-goodwin: Oh this is for workers.dev domain? I thought it was for custom domain?

john-royal: The option for enabling the `r2.dev` domain, in our R2Bucket resource, is called `allowPublicAccess` (probably because Claude). I changed the documentation header for it so users don't confuse it with the `customDomain` prop since the title "public access" is a bit misleading.

Should we also change the prop name?

alchemy-web/src/content/docs/providers/cloudflare/bucket.md
```diff
@@ -65,6 +65,19 @@ console.log(publicBucket.domain); // [random-id].r2.dev
...
+## With Custom Domain
+
+Create a bucket with a custom domain:
+
+```ts
+import { R2Bucket } from "alchemy/cloudflare";
+
+const customDomainBucket = await R2Bucket("custom-domain-bucket", {
+  name: "custom-domain-bucket",
+  customDomain: "custom-domain.com",
```
sam-goodwin: I'd prefer `domain: string`

sam-goodwin:
```ts
domains: string | string[];
devDomain: boolean; (input)
devUrl: string (output)

/** @deprecated use managedDomain (or devDomain) instead */
allowPublicAccess: boolean;
```

john-royal: Implemented this with two exceptions: 
- turns out we were returning `<id>.r2.dev` for the dev domain, so it's a URL not a domain - so I made the output property `devDomain`
- `domains` input is string | string[] | R2BucketCustomDomainOptions | R2BucketCustomDomainOptions[] so you can set the options if you want to

alchemy/test/cloudflare/bucket.test.ts
```diff
@@ -364,10 +364,7 @@ describe("R2 Bucket Resource", async () => {
       });
 
       await new Promise((r) => setTimeout(r, 1000));
-      const rules = await getBucketLifecycleRules(api, bucketName, {
-        ...bucket,
-        delete: true,
-      });
```
john-royal, line 370: This was causing a type error because of the difference between the props and output type, and was unnecessary because the only thing this third parameter is used for is the jurisdiction for requests

alchemy/src/cloudflare/bucket.ts
```diff
@@ -590,10 +624,14 @@ const _R2Bucket = Resource(
       props.name ?? (this.output?.name || this.scope.createPhysicalName(id));
 
     if (this.phase === "update" && this.output?.name !== bucketName) {
-      this.replace();
+      this.replace(
+        // force replacement if custom domain is enabled; otherwise, replace will fail because the bucket has child resources
+        !!this.output?.devDomain,
+      );
```
sam-goodwin, line 630: Hmm shit, this isn't going to work when a Bucket is bound to a Worker I believe. Do you get an error complaining that the "Bucket is bound to a Worker"?

john-royal: Ohhh that's a good catch... I haven't tried that but let me double check

john-royal: I just tested manually and it worked fine! I'm surprised.

Should I add a test for this specific case just to be sure?

PR #1186, opened 2025-10-29
fix(telemetry): telemetry will timeout if unable to send in 1 second

Mkassabov:


alchemy/src/util/telemetry.ts
```diff
@@ -375,6 +376,13 @@ async function fetchNoResponse(
...
 
   await new Promise<void>((resolve, reject) => {
+    setTimeout(() => {
+      if (!SUPPRESS_TELEMETRY_ERRORS) {
+        logger.warn("Telemetry request timed out");
+      }
+      reject(new Error("Request timed out"));
+    }, options?.timeout ?? 1000);
+
     const req = client.request(requestOptions);
```
sam-goodwin, line 380: Should we use an AbortSignal.timeout?

sam-goodwin (approved):

Issue #1109, opened 2025-10-10
Cloudflare Containers: support prebuilt/static images (skip rebuilds)


adunne09:
**Summary**
- Current Cloudflare Container resource always builds and pushes an image on deploy.
- Add support to use a prebuilt/static image reference (e.g., from Cloudflare Registry or a remote registry) without rebuilding.
- Keep current behavior as default; introduce a clear, validated path to skip build/push.

**Motivation**
- Faster CI/CD by reusing immutable, prebuilt images.
- Support external registries and promotion workflows (dev → staging → prod).
- Deterministic deploys with signed/attested images.
- Aligns with Docker provider patterns already present in the repo.

**Current Behavior**
- `alchemy/src/cloudflare/container.ts` always derives from `docker/Image` and triggers `docker build` and push to Cloudflare registry.
- `alchemy/src/docker/image.ts` is the only path; no direct acceptance of a prebuilt image reference.
- `worker` deployment consumes the image produced by the build step.

Key files:
- `alchemy/src/cloudflare/container.ts` — Container resource delegates to `Image` and pushes to CF registry.
- `alchemy/src/cloudflare/worker.ts` — Uses the image from the Container resource during `ContainerApplication`.
- `alchemy/test/cloudflare/container.test.ts` — Tests build/update/adoption via build flow.
- `examples/cloudflare-container/*` — Demonstrates only the build flow.
- `alchemy/src/docker/remote-image.ts` — Remote prebuilt image resource exists but is unused by Cloudflare container.

**Proposed Changes**
- Extend `ContainerProps` to accept a prebuilt image and allow skipping build/push.
- Validate mutual exclusivity between build and prebuilt-image paths.
- Preserve current defaults and ergonomics.

Proposed props (flat, backwards-compatible):
- `image?: string | RemoteImage` — Prebuilt image reference (e.g., `<image name>:<tag>` per current wrangler pattern).
- `push?: boolean` — When `image` is provided and not in Cloudflare registry, optionally retag and push to Cloudflare. Default `false`.
- Existing `build`-related props remain supported and unchanged.

Behavior:
- If `image` is provided:
  - Skip `docker build`.
  - If `image` is a `string` without a registry host: assume `registry.cloudflare.com/<acct>/<image>:<tag>`.
  - If `image` points to a non-Cloudflare registry:
    - Cloudflare currently requires images to be in the Cloudflare registry. Enforce `push: true` to retag and push into CF; otherwise, throw a clear validation error.
[...]

PR #1187, opened 2025-10-29
feat(docker): memoize image builds

john-royal:
Should help with #1109

### DIFF
--- a/alchemy/src/docker/image.ts
+++ b/alchemy/src/docker/image.ts
@@ -1,13 +1,15 @@
-import crypto from "node:crypto";
+import { exec } from "node:child_process";
 import fs from "node:fs/promises";
 import os from "node:os";
+import { promisify } from "node:util";
 import path from "pathe";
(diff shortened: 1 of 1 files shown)


alchemy/src/cloudflare/container.ts
```diff
@@ -195,10 +195,6 @@ export async function Container<T>(
 ): Promise<Container<T>> {
   const scope = Scope.current;
   const name = props.name ?? scope.createPhysicalName(id);
-  const tag =
-    props.tag === undefined || props.tag === "latest"
-      ? `latest-${Date.now()}`
-      : props.tag;
```
john-royal, line 201: Why was this needed?

Was Cloudflare throwing an error or did it just not update when necessary?

It doesn't throw an error but should I double check that the image updates as expected?

john-royal: Update: this is now handled automatically within the `Image` resource.

PR #1189, opened 2025-10-30
fix(cloudflare): while adopting non alchemy worker check if old binding is a container

BlankParticle:


alchemy/src/cloudflare/worker-metadata.ts
```diff
@@ -312,7 +312,10 @@ export async function prepareWorkerMetadata(
         // let's now apply a herusitic based on binding name (assume binding name is consistent)
         // TODO(sam): this has a chance of being wrong, is that OK? Users should be encouraged to upgrade alchemy version and re-deploy
         const object = props.bindings?.[oldBinding.name];
-        if (object && isDurableObjectNamespace(object)) {
+        if (
+          object &&
+          (isDurableObjectNamespace(object) || isContainer(object))
+        ) {
```
sam-goodwin, line 318: What was the error that prompted this fix?

BlankParticle: i had an non alchemy worker which I wanted to adopt, but it kept trying to delete my Sandbox container which is the last fallback branch here
will provide more details on discord

PR #1196, opened 2025-11-05
chore(telemetry): add retrying to telemetry errors to telemetry data

Mkassabov:
requires https://github.com/alchemy-run/telemetry/pull/4


alchemy/src/util/telemetry.ts
```diff
@@ -299,25 +300,43 @@ export async function createAndSendEvent(
...
+          telemetryErrors.map((e) => serializeError(e)),
+        ),
+      };
+      await fetchNoResponse(TELEMETRY_API_URL, {
+        method: "POST",
+        headers: {
+          "Content-Type": "application/json",
+        },
+        body: JSON.stringify(eventData),
+      });
```
sam-goodwin, line 327: should you return after this so it doesn't loop?

Mkassabov: fixed

PR #1197, opened 2025-11-05
feat(cloudflare): inherit binding for workers

Mkassabov:


alchemy/src/cloudflare/inherit.ts
```diff
@@ -0,0 +1,31 @@
...
+ * await Worker("my-worker", {
+ *   name: "my-worker",
+ *   entrypoint: "./src/worker.ts",
+ *   bindings: {
+ *     KEY: Inherit("Key")
+ *   }
+ * });
+ * ```
+ */
+export function Inherit(options?: { oldName?: string; versionId?: string }) {
```
john-royal, line 19: Is there a reason we didn't return type `Inherit`? Wondering if this might mess up syntax highlighting

alchemy/src/cloudflare/miniflare/build-worker-options.ts
```diff
@@ -205,6 +205,10 @@ export const buildWorkerOptions = async (
         }
         break;
       }
+      case "inherit": {
+        logger.warn("unable to inherit bindings in development mode");
```
john-royal, line 209: Makes sense but wonder if this should be an error instead of a warning. I don't have a strong opinion though

john-royal (approved):

PR #1198, opened 2025-11-05
feat(docker):  add build.cacheTo prop to docker image and build.options as escape hatch for additional docker build options

asterikx:
Fully support caching of docker images/layers to external cache destination to speed up docker image builds, e.g. in github actions: https://docs.docker.com/reference/cli/docker/buildx/build/#cache-to

### DIFF
--- a/alchemy/src/docker/image.ts
+++ b/alchemy/src/docker/image.ts
@@ -54,6 +54,14 @@ export interface DockerBuildOptions {
    *
    */
   cacheTo?: string[];
+
+  /**
+   * Additional options to pass to the Docker build command. This serves as an escape hatch for any additional options that are not supported by the other properties.
+   *
+   * @see https://docs.docker.com/reference/cli/docker/buildx/build/#options
+   *
+   */
+  options?: string[];
 }
 
 export interface ImageRegistry {
@@ -181,7 +189,7 @@ export const Image = Resource(
       const buildOptions: Record<string, string> = props.build?.args || {};
 
       // Add platform if specified
-      let buildArgs = ["build", "-t", imageRef];
+      const buildArgs = ["build", "-t", imageRef];
 
       if (props.build?.platform) {
         buildArgs.push("--platform", props.build.platform);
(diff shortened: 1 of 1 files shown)


alchemy/src/docker/image.ts
```diff
@@ -188,6 +199,13 @@ export const Image = Resource(
         }
       }
 
+      // Add cache destinations if specified
+      if (props.build?.cacheTo && props.build.cacheTo.length > 0) {
+        for (const cacheTarget of props.build.cacheTo) {
+          buildArgs.push("--cache-to", cacheTarget);
+        }
+      }
```
sam-goodwin, line 215: I noticed we have a `buildArgs: Record<string, string>` in `ImageProps` but maybe `buildArgs: string[]` would be better

sam-goodwin: It would mean we'd never have to add first class properties like this because you'd never be blocked

asterikx: I would agree with you @sam-goodwin! Docker calls them `options` in their docs: https://docs.docker.com/reference/cli/docker/buildx/build/#options. “--build-arg” is an option, just like “--cache-from” or “--platform” etc.
Wdyt about the naming? "buildArgs", "args", or "options" as Docker calls them? And what is your opinion on backward compatibility/the breaking change this would bring?

Creating a docker image could look like:
```ts
import * as docker from "alchemy/docker";

const myImage = await docker.Image("app-image", {
  name: "my-app",
  // tag: "v1.0", <-- now part of options
  build: {
    context: "./app",
    options: ["--build-arg key1=val1", "--build-arg key2=val2", "--platform linux/amd64,linux/arm64", "-t v1.0"]
  }
});
```

This is super flexible, but also not type-safe at all. If type-safety is a priority, one could mimic the Docker GitHub Action's customization params in TS: https://github.com/docker/build-push-action?tab=readme-ov-file#customizing.
Happy to get feedback on this. 🙂

sam-goodwin: build.options sounds good.

We can always have type safe alternatives that achieve the same thing but an escape hatch is more important 

Blocking is catastrophic

alchemy/src/docker/image.ts
```diff
@@ -40,9 +40,28 @@ export interface DockerBuildOptions {
...
+   */
+  cacheTo?: string[];
+
+  /**
+   * Additional options to pass to the Docker build command. This serves as an escape hatch for any additional options that are not supported by the other properties.
+   *
+   * @see https://docs.docker.com/reference/cli/docker/buildx/build/#options
+   *
+   */
+  options?: string[];
```
sam-goodwin, line 64: PR title should be update to reflect the scope of this change

PR #1207, opened 2025-11-11
feat(cloudflare): vpc service

john-royal:
Adds support for Cloudflare VPC service bindings.


alchemy/src/cloudflare/miniflare/build-worker-options.ts
```diff
@@ -30,29 +30,26 @@ export interface MiniflareWorkerInput {
...
+          | "d1"
+          | "images"
+          | "kv_namespace"
+          | "queue"
+          | "r2_bucket";
       }
-    > & {
-      raw: true;
-    })
-  | WorkerBindingService;
```
sam-goodwin, line 56: Why did this change

alchemy/src/cloudflare/vpc-service.ts
```diff
@@ -0,0 +1,320 @@
...
+
+export type VpcService = Omit<VpcServiceProps, "name" | "adopt"> & {
+  name: string;
+  serviceId: string;
+  createdAt: number;
+  updatedAt: number;
+  type: "vpc_service";
+};
+
+export const VpcService = Resource(
```
sam-goodwin, line 235: Docs

sam-goodwin:
merge conflicts to resolve

alchemy/src/cloudflare/miniflare/build-worker-options.ts
john-royal: Type needed to be updated to add VPC service, which does not require raw flag. Messed around with it while trying to figure out if it was needed. Realized that the `RemoteOptional`/`RemoteOnly` type was not actually used except for here so simplified.

```ts
type RemoteBinding =
  // Supported remote bindings that are NOT a fetcher require the `raw` flag.
  | (Extract<
      WorkerBindingSpec,
      {
        type:
          | "ai"
          | "browser"
          | "dispatch_namespace"
          | "mtls_certificate"
          | "vectorize"
          | "d1"
          | "images"
          | "kv_namespace"
          | "queue"
          | "r2_bucket";
      }
    > & { raw: true })
  // Fetcher type bindings do not require the `raw` flag and will throw an error if it is present.
  | Extract<WorkerBindingSpec, { type: "service" | "vpc_service" }>;
```

PR #1217, opened 2025-11-17
Fix CI workflow not deleting prod resources

simplyzetax:
This update fixes the example workflow in the documentation to not delete resources from the `prod` stage when running the cleanup job after merging a pull request.


sam-goodwin:
Are you sure this is a problem with our docs? The documented workflow only runs when a PR is closed. If that deletes prod, i'd assume your `alchemy.run.ts` is not parameterized properly?
```yml
if: ${{ github.event_name == 'pull_request' && github.event.action == 'closed'}}
```

rem4ik4ever:
> Are you sure this is a problem with our docs? The documented workflow only runs when a PR is closed. If that deletes prod, i'd assume your `alchemy.run.ts` is not parameterized properly?
> 
> ```yaml
> if: ${{ github.event_name == 'pull_request' && github.event.action == 'closed'}}
> ```

I did run into the same issue with documentation CI example for `deploy.yml`. It cleans up and destroys production environment when PR is closed.

sam-goodwin:
Oh i misread the PR, this looks good. Thanks @simplyzetax

sam-goodwin:
I wonder if we should also add a check in the CLI that requires `--yes` if you try to run `alchemy destroy --stage prod`. That might be safer than relying on logic in a github CI workflow definition

PR #1224, opened 2025-11-20
feat(neon): NeonProject supports `adopt`

ericclemmons:
## Summary

~`adopt: "adj-noun-123"` will adopt the existing Neon project.~
~I know this is a departure from `adopt?: boolean`, but it made sense at the time.~
~The alternative would be something like:~

```ts
await NeonProject("db", {
  adopt: true,
  name: "the-exact-name-used-in-neon-not-the-id"
});
```

~And then I'd use https://api-docs.neon.tech/reference/listprojects to get it.~
~Let me know!~

### My Project

```ts
const neon = await NeonProject("db", {
  // https://console.neon.tech/app/projects/aged-term-60453743
  adopt: "aged-term-60453743",
});

let [connectionUri] = neon.connection_uris;

if (app.local) {
  const branch = await NeonBranch("db-branch", {
    endpoints: neon.endpoints,
    initSource: "schema-only",
    project: neon,
  });

  connectionUri = branch.connectionUris[0];
}

console.info({ connectionUri });
```

Screenshot 2025-11-20 at 2 21 24 PM
Screenshot 2025-11-20 at 2 24 57 PM

### Alchemy Monorepo

```shell
/Users/eric/Projects/ericclemmons/alchemy $
bun vitest run ./alchemy/test/neon/project.test.ts


 RUN  v3.2.4 /Users/eric/Projects/ericclemmons/alchemy

 ✓ alchemy/test/neon/project.test.ts (1 test) 8072ms
   ✓ NeonProject Resource > create, update, and delete neon project  8070ms

 Test Files  1 passed (1)
      Tests  1 passed (1)
   Start at  14:20:24
   Duration  9.80s (transform 868ms, setup 27ms, collect 1.36s, tests 8.07s, environment 0ms, prepare 70ms)
```

### DIFF
--- a/alchemy/src/neon/project.ts
+++ b/alchemy/src/neon/project.ts
@@ -1,6 +1,6 @@
+import { alchemy } from "../alchemy.ts";
 import type { Context } from "../context.ts";
 import { Resource } from "../resource.ts";
-import { Secret } from "../secret.ts";
 import { createNeonApi, type NeonApiOptions } from "./api.ts";
 import type { NeonClient } from "./api/sdk.gen.ts";
 import type * as neon from "./api/types.gen.ts";
@@ -436,13 +436,13 @@ async function fetchProject(
     });
     const url = new URL(uri);
     return {
-      connection_uri: new Secret(uri),
+      connection_uri: alchemy.secret(uri),
       connection_parameters: {
         database: database.name,
         host: url.host,
         port: 5432,
         user: url.username,
-        password: new Secret(url.password),
+        password: alchemy.secret(url.password),
       },
     };
   }


alchemy/src/neon/project.ts
```diff
@@ -32,6 +33,11 @@ export type NeonPgVersion = 14 | 15 | 16 | 17 | 18;
  * Properties for creating or updating a Neon project
  */
 export interface NeonProjectProps extends NeonApiOptions {
+  /**
+   * Given a Project ID (e.g. `adjective-noun-123`), adopt the project if it exists.
+   */
+  adopt?: string;
```
ericclemmons: Nevermind – we need `name` I think to prevent it from being generated and destroying the same DB because the name differs across stages (e.g. `${app}-${stage}-${id}`)

ericclemmons: Resolved via 28d6743db51eaaecf877a187e6e25045711ecb02

alchemy/src/neon/project.ts
```diff
@@ -193,6 +213,101 @@ export const NeonProject = Resource(
...
+          const {
+            data: { endpoints },
+          } = await api.listProjectEndpoints({
+            path: { project_id: project.id },
+          });
+          const {
+            data: { roles },
+          } = await api.listProjectBranchRoles({
+            path: { project_id: project.id, branch_id: branch.id },
+          });
```
sam-goodwin: nit: Should we parallelize these?

alchemy/src/neon/project.ts
```diff
@@ -32,6 +33,18 @@ export type NeonPgVersion = 14 | 15 | 16 | 17 | 18;
...
 export interface NeonProjectProps extends NeonApiOptions {
+  /**
+   * When `true`, will adopt an existing project by `name`
+   */
+  adopt?: true;
+
+  /**
+   * Whether to delete the database when the resource is destroyed.
+   * When false, the database will only be removed from the state but not deleted via API.
+   * @default true, unless the resource was adopted
```
Mkassabov, line 45: This is nice, but I think its better to be consistent with other alchemy resource and have delete always default to true.

Mkassabov (commented):
can you also add adopt to the neon project docs?

alchemy/src/neon/project.ts
john-royal: I agree. I wonder if we should add a `ProjectRef` resource if you want a read-only reference that doesn't get deleted?

alchemy/src/neon/project.ts
ericclemmons: totes

alchemy/src/neon/project.ts
ericclemmons: consistency ftw

@john-royal That would be interesting!  I wonder if it could be a top-level 

```ts
const db = await ReadOnly(NeonProject({ ... })
```

But that's probably outside of this scope.  I think being able to have "global" or multi-stage definitions makes more sense to how things are used today.  (Though "alchemy-aware, but not managed is a great value prop!")

sam-goodwin: I think this PR is ready to merge if we change the default behavior to not enable delete: false when adopted so that it's consistent with alchemy?

ericclemmons:
I'll be making these updates soon: But `NeonBranch` *also* needs `adopt?: boolean`, otherwise it's possible for a shared branch (e.g. `development`) that individuals branch off of to hit a `409` error.

alchemy/src/neon/project.ts
```diff
@@ -193,6 +214,17 @@ export const NeonProject = Resource(
 
     switch (this.phase) {
       case "create": {
+        if (props.adopt) {
+          try {
+            return await fetchProject(api, { ...props, name });
+          } catch (error) {
+            if (!(error instanceof NeonProjectNotFound)) {
```
john-royal, line 221: Added a custom error type here so behavior can be more consistent with `adopt` in other Alchemy resources, i.e. we adopt if we can find it or create otherwise.

alchemy/src/neon/project.ts
```diff
@@ -298,3 +331,125 @@ export const NeonProject = Resource(
...
+  ): Promise<NeonConnectionUri> {
+    const {
+      data: { uri },
+    } = await api.getConnectionUri({
+      path: { project_id: project.id },
+      query: {
+        branch_id: database.branch_id,
+        database_name: database.name,
+        role_name: database.owner_name,
+        pooled: false, // match default behavior from create endpoint (should this be true?)
```
john-royal, line 434: Open to feedback on this one - did this to make tests pass but perhaps we should change the default behavior to always use the pooled URL?

alchemy/src/neon/project.ts
john-royal: Made this change

alchemy/src/neon/project.ts
```diff
@@ -298,3 +331,125 @@ export const NeonProject = Resource(
...
+        throw new NeonProjectNotFound(props.name);
+      }
+      if (projects.length > 1) {
+        throw new Error(
+          `Multiple projects found with name "${props.name}". Name must be unique when referencing existing projects.`,
+        );
+      }
+      return projects[0];
+    });
+  const endpointsPromise = api
```
john-royal, line 370: Declared this and `rolesPromise` separately from awaiting them in `Promise.all` at the end as a slight optimization since they're not part of the branch > databases > connection URI waterfall.

john-royal:
Hey @ericclemmons, I made some slight changes - would you like to give this a once over before we merge? Specific changes are:

- For consistency with other Alchemy resources, `adopt: true` does not automatically set `delete: false`.
- Also for consistency with Alchemy, `adopt: true` will try to create the new resource if an existing one is not found.
- If you'd like a read-only reference, you can use `NeonProjectRef` - same logic used for adoption.
- Updated docs and tests.

alchemy/src/neon/project.ts
sam-goodwin: Why did it help the tests pass

alchemy/src/neon/project.ts
```diff
@@ -415,3 +331,125 @@ export const NeonProject = Resource(
...
+      query: {
+        branch_id: database.branch_id,
+        database_name: database.name,
+        role_name: database.owner_name,
+        pooled: false, // match default behavior from create endpoint (should this be true?)
+      },
+    });
+    const url = new URL(uri);
+    return {
+      connection_uri: new Secret(uri),
```
sam-goodwin: should this be `alchemy.secret(uri)`?

sam-goodwin (approved):

alchemy/src/neon/project.ts
sam-goodwin: The comment needs to be updated

PR #1226, opened 2025-11-24
fix(cloudflare): delete workflows on worker delete

john-royal:


alchemy/src/cloudflare/workflow.ts
```diff
@@ -116,3 +116,12 @@ export async function upsertWorkflow(
...
 }
+
+export async function deleteWorkflow(api: CloudflareApi, name: string) {
+  const response = await api.delete(
+    `/accounts/${api.accountId}/workflows/${name}`,
+  );
+  if (!response.ok && response.status !== 404) {
+    await handleApiError(response, "delete", "workflow", name);
+  }
+}
```
sam-goodwin, line 127: Have you tested this? I don't see any updated tests

PR #1227, opened 2025-11-27
feat(docker): support `adopt` for `Container` and `Volume`

unknown:
Implements `adopt` flag for `Container` and `Volume`. I've chosen to start adopted containers if `start` is `true` because I would find this behavior useful.


alchemy/src/docker/volume.ts
```diff
@@ -133,33 +139,65 @@ export const Volume = Resource(
...
-      };
     }
+
+    // Check if volume already exists
+    const volumeExists = await api.volumeExists(volumeName);
+
+    if (volumeExists) {
+      if (this.phase === "update") {
+        // Remove existing volume for update
+        await api.removeVolume(volumeName);
```
sam-goodwin: This is destructive, why should it be removed during the update phase?

sam-goodwin: Tests are passing so this looks mostly good to merge, but I am a bit unsure about this destructive action. Seems counter to the point of Volumes

unknown: I was mostly following the logic of `Container`, but you're right and I didn't clock that it's destructive to the data stored in volumes. In that case, not sure what we should do in the update case: throw a explicit error saying that it's not supported or just fail at `createVolume` (like what would've happened before)?

sam-goodwin: Can any of these properties be updated without destroying the volume?

```ts
export interface VolumeProps {
  /**
   * Volume name
   *
   * @default ${app}-${stage}-${id}
   */
  name?: string;

  /**
   * Volume driver to use
   * @default "local"
   */
  driver?: string;

  /**
   * Driver-specific options
   */
  driverOpts?: Record<string, string>;

  /**
   * Custom metadata labels for the volume
   */
  labels?: VolumeLabel[] | Record<string, string>;
}
```

If a property is updated that can't be changed, then you should call `this.replace()` which will destroy and re-create the resource. Hopefully we don't need to do that since destroying the user's data is bad.

unknown: Unfortunately, it seems like only cluster volumes (Docker-Swarm specific) are able to be modified directly ([see docs](https://docs.docker.com/reference/cli/docker/volume/update/)). I'll use `this.replace()` for now, but there could be some room for improvement in the future around automatically backing up volume data before it's destroyed and automatically restoring it in the new volume

PR #1229, opened 2025-11-29
Modifies CamelToSnake to be able to return primitive types if nested

### LINKED ISSUE #1228: Error adding Logs or Traces observability destination in Cloudflare Worker
Hi! First off - huge fan of what you guys are doing. Keep it up! :D

There is an issue when trying to setup OpenTelemetry destinations in Cloudflare workers. Consider the worker below:

`export const worker = await Worker("Api", {
  entrypoint: "src/worker.ts",
  //url: true,
  domains: [`api.example.com`],
  compatibilityDate: "2025-11-29",
  compatibility: "node",
  observability: {
    logs: {
      enabled: true,
      headSamplingRate: 1,
      invocationLogs: true,
      persist: false,
      destinations: ["betterstack-logs"],
    }
  }
})`

Right now this will fail with this error:
CloudflareApiError: Failed to upload worker script "[scriptname here]" (400 Bad Request):
- [10021] json: cannot unmarshal object into Go struct field LogSettings.observability.logs.destinations of type string (https://developers.cloudflare.com/workers/observability/errors/#validation-errors-10021)

Indeed if we log the finalMetadata object in [worker.ts](https://github.com/alchemy-run/alchemy/blob/main/alchemy/src/cloudflare/worker.ts) we can see this:

```
observability: {
    logs: {
      enabled: true,
      head_sampling_rate: 1,
      invocation_logs: true,
      persist: false,
      destinations: [
        [Object ...]
      ],
    },
    enabled: true,
  },
```

So an object is currently being sent instead of a string. I've identified the issue and it seems to be located in [camel-to-snake.ts](https://github.com/alchemy-run/alchemy/blob/main/alchemy/src/util/camel-to-snake.ts).  The culprit is this method:

```
export function camelToSnakeObjectDeep<T extends object | undefined>(
  obj: T,
): T extends undefined ? undefined : CamelToSnake<T> {
  return (
    obj
      ? Object.fromEntries(
          Object.entries(obj).map(([key, value]) => [
            key
              .replace(/([A-Z]+)([A-Z][a-z])/g, "$1_$2") // Handle consecutive capitals: "FOOBar" -> "FOO_Bar"
              .replace(/([a-z])([A-Z])/g, "$1_$2") // Handle normal camelCase: "fooBar" -> "foo_Bar"
              .toLowerCase(),
            Array.isArray(value)
              ? value.map(camelToSnakeObjectDeep)
              : typeof value === "object" && value !== null
                ? camelToSnakeObjectDeep(value)
                : value,
          ]),
        )
      : undefined
  ) as T extends undefined ? undefined : CamelToSnake<T>;
```

The problem is that this method will never return a basic type like a string or int if it is nested in an array - only objects. The solution is this

```
export function camelToSnakeObjectDeep<T>(
  obj: T,
): T extends undefined ? undefined : T extends object ? CamelToSnake<T> : T {
  if (obj === undefined || obj === null) {
    return obj as any;
  }
  
  // If it's not an object (primitive types), return as-is
  if (typeof obj !== "object") {
    return obj as any;
  }
  
  // Handle arrays
  if (Array.isArray(obj)) {
    return obj.map(camelToSnakeObjectDeep) as any;
  }
  
  // Handle objects
  return Object.fromEntries(
    Object.entries(obj).map(([key, value]) => [
      key
        .replace(/([A-Z]+)([A-Z][a-z])/g, "$1_$2") // Handle consecutive capitals: "FOOBar" -> "FOO_Bar"
        .replace(/([a-z])([A-Z])/g, "$1_$2") // Handle normal camelCase: "fooBar" -> "foo_Bar"
        .toLowerCase(),
      Array.isArray(value)
        ? value.map(camelToSnakeObjectDeep)
        : typeof value === "object" && value !== null
          ? camelToSnakeObjectDeep(value)
          : value,
    ]),
  ) as any;
}
```

Fixed in pull request #1229

bjorntechCarl:
Closes #1228

### DIFF
--- a/alchemy/test/util/camel-to-snake.test.ts
+++ b/alchemy/test/util/camel-to-snake.test.ts
@@ -120,6 +120,11 @@ describe("camelToSnakeObjectDeep", () => {
       expect(result).toEqual(["a", "b", "c"]);
     });
 
+    it("handles object with arrays of strings", () => {
+      const result = camelToSnakeObjectDeep({ someArray: ["a", "b", "c"] });
+      expect(result).toEqual({ some_array: ["a", "b", "c"] });
+    });
+
     it("handles array of objects", () => {
       const result = camelToSnakeObjectDeep([
         { firstName: "John" },


sam-goodwin:
Thanks for the initiative here. It might help to add some test cases to more explicitly document the behavior we want and prevent regressions.

sam-goodwin:
I added some tests

PR #1235, opened 2025-12-03
fix(cloudflare): pass safe user headers to miniflare websocket

john-royal:


julienroubieu:
This fixes my authentication issues, thanks!

PR #1236, opened 2025-12-04
feat(cloudflare): Worker.DevDomain and Worker.DevUrl

sam-goodwin:
`Worker.DevUrl` and `Worker.DevDomain` bindings can now be used to auto-inject the `workers.dev` domain
```ts
export const website = await Vite("website", {
  entrypoint: "src/index.ts",,
  bindings: {
    VITE_PUBLIC_DOMAIN: Worker.DevDomain,
    VITE_PUBLIC_URL: Worker.DevUrl,
  },
  dev: {
    command: "vite dev --port 5006",
    // for local dev, you must tell us the domain because we can't know what port you will use
    domain: "localhost:5006",
  },
});
```


alchemy/src/cloudflare/website.ts
```diff
@@ -207,29 +195,50 @@ export async function Website<B extends Bindings>(
...
+        value.type === "cloudflare::Worker::DevDomain" ||
+        value.type === "cloudflare::Worker::DevUrl"
+      ) {
+        if (Scope.current.local) {
+          const domain = typeof dev === "object" ? dev.domain : undefined;
+          if (!domain) {
+            // we can't know which port the local web server (e.g. vite dev) will use
+            // vite dev # default to 5173 or configured in vite.config.ts
+            // vite dev --port XYZ
+            // next dev # who fucking knows
```
Mkassabov, line 253: lmfoa its port 3000 for next

sam-goodwin: > michael fucking knows

alchemy/src/cloudflare/miniflare/build-worker-options.ts
```diff
@@ -92,6 +82,14 @@ export const buildWorkerOptions = async (
...
       continue;
     }
+    if (binding.type === "cloudflare::Worker::DevDomain") {
+      (options.bindings ??= {})[key] = `localhost:${port}`;
+      continue;
+    }
+    if (binding.type === "cloudflare::Worker::DevUrl") {
+      (options.bindings ??= {})[key] = `http://localhost:${port}`;
+      continue;
+    }
```
Mkassabov, line 104: why are we hard coding these here instead of using `computeWorkerDevDomain`?

Mkassabov (approved):
looks right about right, left a question but its not blocking. do we have a formatter issue, since a lot of this diff seems to be formatting?

fair warnining: I didn't run tests

alchemy/src/cloudflare/miniflare/build-worker-options.ts
sam-goodwin: This is for miniflare only and doesn't apply to `Website` (only raw `Worker`).

sam-goodwin:
> do we have a formatter issue, since a lot of this diff seems to be formatting?

I just pushed a commit with `bun format`

alchemy/src/cloudflare/website.ts
Mkassabov: I wish I didn't

alchemy/src/cloudflare/miniflare/build-worker-options.ts
Mkassabov: good enough for me

PR #1240, opened 2025-12-05
chore(cloudflare): add missing cache fields to hyperdrive

john-royal:


sam-goodwin:
This is a type-only change, do they just get passed through?

PR #1243, opened 2025-12-06
Cloudflare: Resources for warp device profiles and CIDR Routes for tunnels 

eastlondoner:
This is all vibe coded. I ran the tests and I am using these resources and they're working for me.

### DIFF
--- a/alchemy/src/cloudflare/warp-default-profile.ts
+++ b/alchemy/src/cloudflare/warp-default-profile.ts
@@ -1,7 +1,6 @@
 import type { Context } from "../context.ts";
 import { Resource, ResourceKind } from "../resource.ts";
 import { handleApiError } from "./api-error.ts";
-import { extractCloudflareResult } from "./api-response.ts";
 import {
   createCloudflareApi,
   type CloudflareApi,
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -27,3 +27,7 @@ smoke
 .wrangler
 
 *.tgz
+# SpecStory
+.specstory/
+# SpecStory explanation file
+.specstory/.what-is-this.md
```
eastlondoner:
```suggestion
```

alchemy/src/cloudflare/warp-default-profile.ts
```diff
@@ -0,0 +1,322 @@
...
+  }
+  if (props.disableAutoFallback !== undefined) {
+    deviceSettings.disable_auto_fallback = props.disableAutoFallback;
+  }
+  if (props.allowModeSwitch !== undefined) {
+    deviceSettings.allow_mode_switch = props.allowModeSwitch;
+  }
+  if (props.switchLocked !== undefined) {
+    deviceSettings.switch_locked = props.switchLocked;
+  }
```
sam-goodwin: I don't know if this is necessary since `undefined` is omitted when serialized as JSON?

eastlondoner: good point - tidied up

PR #1253, opened 2025-12-11
fix(prisma-postgres): use connection.database.name in connection string

xiaoyu2er:
## Summary

Fixed the Prisma Postgres connection string construction to use `connection.database.name` instead of `connection.database`.

## Changes

- **connection.ts**: Fixed connection string to correctly access the database name via `connection.database.name`
- **connection.test.ts**: Added additional test assertions for `prismaConnectionString`, `host`, and `user` properties

## Code Example

```typescript
// Before (incorrect)
connectionString: secret(
  \`postgres://\${connection.user}:\${connection.pass}@\${connection.host}/\${connection.database}\`,
),

// After (correct)
connectionString: secret(
  \`postgres://\${connection.user}:\${connection.pass}@\${connection.host}/\${connection.database.name}\`,
),
```

## Testing

- [x] Test passes with valid Prisma service token

### DIFF
--- a/alchemy/src/prisma-postgres/connection.ts
+++ b/alchemy/src/prisma-postgres/connection.ts
@@ -143,7 +143,7 @@ export const Connection = Resource(
       createdAt: connection.createdAt,
       prismaConnectionString: secret(connection.connectionString),
       connectionString: secret(
-        `postgres://${connection.user}:${connection.pass}@${connection.host}/${connection.database}`,
+        `postgres://${connection.user}:${connection.pass}@${connection.host}/${connection.database.name}`,
       ),
       database: database,
       host: connection.host,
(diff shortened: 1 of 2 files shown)


alchemy/src/prisma-postgres/connection.ts
```diff
@@ -143,7 +143,7 @@ export const Connection = Resource(
       createdAt: connection.createdAt,
       prismaConnectionString: secret(connection.connectionString),
       connectionString: secret(
-        `postgres://${connection.user}:${connection.pass}@${connection.host}/${connection.database}`,
+        `postgres://${connection.user}:${connection.pass}@${connection.host}/${connection.database.name}`,
```
xiaoyu2er, line 146: `connection.database` is an object not a string.

PR #1254, opened 2025-12-11
feat(cloudflare): d1 import

john-royal:
Specify `importDir` on `D1Database` to specify SQL files to import. Uses the D1 import API, which it turns out we already had implemented for database cloning.

Notes:
- Imports are run as migrations on the local database. To fix the statement-splitting problem, I split by semicolons at the end of new lines - not 100% bulletproof but should be more than good enough given what an import file will typically look like.
- I updated the `listMigrationsFiles` function:
  - rename to `listSqlFiles` since it's also used for imports
  - mark the `sql` property as non-enumerable (keeps it out of state, to prevent state store issues caused by massive files)
  - include a `hash` property (so we can still accurately diff props based on file contents)


sam-goodwin:
>  I split by semicolons at the end of new lines - not 100% bulletproof but should be more than good enough given what an import file will typically look like.

Won't that not work for a drizzle migration script which adds a comment?

nickbalestra:
Also tested on my end with `https://pkg.pr.new/alchemy-run/alchemy@1254` and that fixed my deployment being broken getting  `CloudflareApiError: Error 400 creating D1 database 'couch-platform-pr-58-db': 7400: Invalid property: name => Required` 

Thanks John!!!

rem4ik4ever:
> Also tested on my end with `https://pkg.pr.new/alchemy-run/alchemy@1254` and that fixed my deployment being broken getting `CloudflareApiError: Error 400 creating D1 database 'couch-platform-pr-58-db': 7400: Invalid property: name => Required`
> 
> Thanks John!!!

My deployments started failing with the same error. Does this PR fixes this issue?

PR #1256, opened 2025-12-11
fix(cloudflare): miniflare compatibility date warning

john-royal:
Sometimes in Alchemy, you'll see an error like this when running dev mode:

```
The latest compatibility date supported by the installed Cloudflare Workers Runtime is <date>, but you've requested <date>. Falling back to <date>...
```

This is because we base our default compatibility date on the latest workerd version at the time of release, rather than the one actually bundled with Alchemy. Here's a quick fix.


scripts/bump.ts
```diff
@@ -152,10 +152,6 @@ await $`bun install`;
 
 console.log(`Updated version to ${newVersion} in package.json`);
 
-// Generate build date for the release
-console.log("Generating Workers compatibility date...");
-await $`cd alchemy && bun ./scripts/generate-compatibility-date.ts`;
```
sam-goodwin, line 157: Delete the script too

alchemy/src/cloudflare/compatibility-date.ts
```diff
@@ -0,0 +1,6 @@
+import { supportedCompatibilityDate } from "miniflare";
```
sam-goodwin, line 1: Or it's that easy? Lol

scripts/bump.ts
john-royal: It's gone (see https://github.com/alchemy-run/alchemy/pull/1256/changes#diff-a08c8bea875bc9db609437b891bfcd848653dea788de16b679a7840e15f57e1a)

PR #1259, opened 2025-12-12
fix(cloudflare): handle next.js 16 absolute paths in wasm imports

### LINKED ISSUE #1241: Next.js 16 build fails due to absolute path in wasm import
Can we support next 16？

john-royal:
Our WASM plugin wasn't handling absolute paths in imports. Didn't know this was a necessity, but Next.js 16 deployments fail unless we handle this specific edge case.

Closes #1241.


vrabota:
@sam-goodwin can we merge it?

vrabota (approved):

sam-goodwin (changes requested):
Can we add some tests?

PR #1261, opened 2025-12-12
feat(planetscale): default role resource, add missing database props, add adopt to tests

john-royal:
I'll split this up into separate PRs so y'all can actually review it, don't worry...


john-royal:
Replaced by #1326

PR #1268, opened 2025-12-14
fix(cloudflare): website resources properly lowercase worker names

Mkassabov:
- tests not included because the way we generate app name for tests makes this a pain to test


alchemy/src/cloudflare/website.ts
```diff
@@ -156,7 +156,7 @@ export async function Website<B extends Bindings>(
   props: WebsiteProps<B>,
 ) {
   const {
-    name = Scope.current.createPhysicalName(id),
+    name = Scope.current.createPhysicalName(id).toLowerCase(),
```
sam-goodwin, line 159: should we do it in `worker.ts` too? any other resources?

sam-goodwin: oh nvm, i see it is in worker.ts already

sam-goodwin (approved):

alchemy/src/cloudflare/website.ts
Mkassabov: This came from a guy who commented on the merged pr for workers saying it was still happening for sveltekit. So it's already in workers.

Might need to be added to other resources, but idk what has this naming limitation and I don't want to bulk change things that don't need it

PR #1273, opened 2025-12-15
feat: add support for 'adopt' prop in neon branches

utopyin:
I had a project where I wanted to migrate to the newest Neon resources and couldn't because the resource had no way of adopting an existing branch. This will help others like me to migrate to Alchemy resources all the way!

### DIFF
--- a/alchemy/src/neon/branch.ts
+++ b/alchemy/src/neon/branch.ts
@@ -301,7 +301,6 @@ async function fetchBranch(
   const branchList: neon.Branch[] = [];
 
   let branchesResponse = await api.listProjectBranches({
-    throwOnError: false,
     path: { project_id: props.projectId },
     query: {
       search: props.name,
@@ -311,7 +310,6 @@ async function fetchBranch(
 
   while (branchesResponse.data?.pagination?.next) {
     branchesResponse = await api.listProjectBranches({
-      throwOnError: false,
       path: { project_id: props.projectId },
       query: {
         search: props.name,


alchemy/src/neon/branch.ts
```diff
@@ -182,6 +188,16 @@ export const NeonBranch = Resource(
         return this.destroy();
       }
       case "create": {
+        if (props.adopt) {
```
utopyin: should this also check for `this.scope.adopt`?

sam-goodwin: Yes

sam-goodwin: this.scope.adopt || props.adopt

alchemy/src/neon/branch.ts
```diff
@@ -277,3 +293,100 @@ export const NeonBranch = Resource(
...
+    query: {
+      search: props.name,
+    },
+  });
+
+  const branch = branchList.data?.branches.find((b) => b.name === props.name);
+
+  if (!branch) {
+    throw new NeonBranchNotFound(props.name);
+  }
```
sam-goodwin: Branch names are not unique AFAIK, so we should fail to adopt if there are two branches with the same name

alchemy/src/neon/branch.ts
```diff
@@ -277,3 +293,100 @@ export const NeonBranch = Resource(
...
+  api: NeonClient,
+  props: NeonBranchProps & { name: string; projectId: string },
+): Promise<NeonBranch> {
+  const branchList = await api.listProjectBranches({
+    throwOnError: false,
+    path: { project_id: props.projectId },
+    query: {
+      search: props.name,
+    },
+  });
```
sam-goodwin: Should we paginate in case there are many?

alchemy/src/neon/branch.ts
utopyin: from: https://neon.com/docs/manage/branches#branch-naming-requirements
 Branch names must be unique within a project. You cannot have two branches with the same name in the same project.

I still added checks for good measure

alchemy/src/neon/branch.ts
utopyin: I was thinking because we pass a search query we should be fine, but I think it's reasonable to check if a pagination cursor is returned in the response 👍

PR #1276, opened 2025-12-18
feat(cloudflare): add "delete: false" option to hyperdrive

john-royal:
Should this default to false?

I had it default to `delete: true` for backwards compatibility. Not sure what the right solution is here though. Either option is a potential footgun.


sam-goodwin:
Why should it default to false? What are the foot guns

PR #1278, opened 2025-12-18
fix(cloudflare): d1 database create error due to malformed request

john-royal:


alchemy/src/cloudflare/d1-database.ts
```diff
@@ -24,7 +24,7 @@ type PrimaryLocationHint =
   | "weur"
   | "eeur"
   | "apac"
-  | "auto"
+  | "oc"
```
sam-goodwin, line 27: Why was auto removed

john-royal: It's not actually an option [according to the API spec](https://developers.cloudflare.com/api/resources/d1/subresources/database/methods/create/):

image

It determines it automatically *if* the parameter is omitted. But I don't think there's an actual `auto` option unless the API has changed.

PR #1279, opened 2025-12-19
fix(cloudflare): d1 local migrations fail with "column already exists"

john-royal:


alchemy/src/cloudflare/d1-database.ts
```diff
@@ -300,7 +300,7 @@ const _D1Database = Resource(
 
     const local = this.scope.local && !props.dev?.remote;
     const dev = {
-      id: this.output?.dev?.id ?? this.output?.id ?? id,
+      id: this.output?.dev?.id ?? this.output?.id ?? databaseName,
```
john-royal, line 303: First cause of errors: the `dev.id` is used by Miniflare to identify databases internally. In a monorepo Miniflare stores all data in the root, under .alchemy/miniflare. If you have more than one D1 database with the same Alchemy ID, then that refers to the same entity in Miniflare. Names have to be unique so this seems safer.

alchemy/src/cloudflare/d1-local-migrations.ts
```diff
@@ -25,20 +25,34 @@ export const applyLocalD1Migrations = async (
...
-      .run();
+        )
+        .run();
+    } else if (!tableInfo.results.some((col) => col.name === "type")) {
+      await session
+        .prepare(
+          `ALTER TABLE ${options.migrationsTable} ADD COLUMN type TEXT NOT NULL DEFAULT 'migration';`,
+        )
+        .run();
+    }
```
john-royal, line 55: Second error cause: once the column is added, this shouldn't run again, otherwise you get a column already exists error. Checking the table info first fixes that.

PR #1288, opened 2026-01-03
fix: pin libsodium-wrappers to fix import error

hstove:
Hey, thanks for making Alchemy!

I just updated a project to use the latest version of Alchemy. This installed `libsodium-wrappers@0.7.16`. When running Alchemy, this caused the error:

```
error: Cannot find module './libsodium.mjs' from '/Users/hank/personal/slot-watch/node_modules/.bun/libsodium-wrappers@0.7.16/node_modules/libsodium-wrappers/dist/modules-esm/libsodium-wrappers.mjs'
```

This is a known error that (I believe) already has a fix, but it's not published to NPM yet. I figured I'd submit a small PR to help others who might hit this issue. You can also easily workaround this issue by overriding in your `package.json`:

```json
  "overrides": {
    "libsodium-wrappers": "0.7.15"
  }
```

Relevant issues:

- https://github.com/jedisct1/libsodium.js/issues/357
- https://github.com/jedisct1/libsodium.js/pull/356

### DIFF
--- a/alchemy/package.json
+++ b/alchemy/package.json
@@ -194,7 +194,7 @@
     "find-process": "^2.0.0",
     "glob": "^10.0.0",
     "jszip": "^3.0.0",
-    "libsodium-wrappers": "0.7.15",
+    "libsodium-wrappers": "^0.8.0",
     "miniflare": "^4.20250906.0",
     "neverthrow": "^8.2.0",
     "open": "^10.1.2",
--- a/bun.lock
+++ b/bun.lock
@@ -52,7 +52,7 @@
         "find-process": "^2.0.0",
         "glob": "^10.0.0",
         "jszip": "^3.0.0",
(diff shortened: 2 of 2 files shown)


sam-goodwin:
Hmm, is there a way we can follow latest version of libsodium?

sam-goodwin:
Ah I see, we have to wait on https://github.com/jedisct1/libsodium.js/issues/357

hstove:
`libsodium-wrappers@0.8.0` is released as of today - I haven't been able to test it yet but it should let alchemy do `^0.8.0`

hstove:
@sam-goodwin I've updated to use `^0.8.0`, and I've confirmed it works in my project with a linked build

Mkassabov:
looks like we are already on 0.8

PR #1289, opened 2026-01-03
chore: update @hey-api/openapi-ts to 0.90.0

mrlubos:
Updated to the latest and greatest.

- I added `pathe` to root `devDependencies` because `scripts/clients/generate.ts` was complaining without it.
- Using a single `createClient()` call instead of one for each client
- Running the script produces a lot of changes due to spec updates, I excluded those changes from this pull request so it's easier to merge


Mkassabov:
What do we get with this update? Or are the new changes not super relevant to alchemy and this is just a version bump?

mrlubos:
@Mkassabov Mainly a version bump that allows you to take advantage of new features and bug fixes if needed. It's hard for me to tell how this affects the generated clients because the input specs are live and seemingly haven't been updated in a while. Maybe the first step would be to generate against the specs using the current version, and upgrading only after that to see the exact diff

Mkassabov:
I don't think we get much value out of merging this. we aren't really planning on adding anything to our sdk generation for async alchemy as of now. might revisit this if that changes.

we appreciate the contribution efforts though!

PR #1292, opened 2026-01-04
fix(cloudflare): prevent Docker pull for prebuilt Container images

adunne09:
## Summary

Fixes 401 errors when passing a prebuilt Cloudflare registry image to the `Container` resource. Previously, the Container resource would attempt to pull the image via Docker even for images already in the CF registry, causing authentication failures.

### `alchemy/src/cloudflare/container.ts`
- Added `resolveImageName()` function that normalizes image references following wrangler's logic:
  - Short names like `myapp:v1` → `registry.cloudflare.com/{accountId}/myapp:v1`
  - CF registry without accountId → adds accountId
  - External registries → pass through unchanged
- Short-circuits Container for prebuilt images - constructs Image object directly without Docker operations
- Added validation that throws if both `build` and `image` are specified
- For dev mode with CF registry images, throws helpful error directing users to use `dev: { remote: true }`

### `alchemy/test/cloudflare/container.test.ts`
- Unit tests for `resolveImageName()` covering short names, CF registry with/without accountId, external registries
- Integration test verifying prebuilt CF registry images skip Docker pull
- Test for mutual exclusivity of build/image props

## Behavior Matrix (matches wrangler)

| Input | Behavior |
|-------|----------|
| Dockerfile path | Build locally, push to CF registry |
| Registry URI | Pass directly to Cloudflare API (no pull/push) |

## Example Usage

```ts
// Deploy now works without attempting to pull the image first
const container = await Container('my-container', {
  className: 'MyContainer',
  image: `registry.cloudflare.com/my-image:v1`,
});
```


alchemy/src/cloudflare/container.ts
```diff
@@ -229,6 +279,76 @@ export async function Container<T>(
...
+      }
+
+      // For external registry images in dev mode, pull and re-tag for Miniflare
+      const image = await Image(id, {
+        image: props.image,
+        tag,
+      });
+
+      // Re-tag the image for Miniflare's cloudflare-dev namespace
+      const dockerApi = new (await import("../docker/api.ts")).DockerApi();
```
sam-goodwin: Don't use dynamic imports, just import at the top of the file

alchemy/test/cloudflare/container.test.ts
```diff
@@ -346,5 +348,96 @@ describe.sequential("Container Resource", () => {
...
+
+    test("throws error when both build and image are specified", async (scope) => {
+      await expect(
+        Container(`${BRANCH_PREFIX}-invalid-container`, {
+          className: "TestContainer",
+          name: "invalid-container",
+          image: "nginx:alpine",
+          build: {
+            context: path.join(import.meta.dirname, "container"),
+          },
```
sam-goodwin: We should also make this a type error with a union type for the props

alchemy/src/cloudflare/container.ts
```diff
@@ -201,10 +201,60 @@ export type Container<T = any> = {
...
+ * - Short names like "myapp:v1" → "registry.cloudflare.com/{accountId}/myapp:v1"
+ * - CF registry without accountId like "registry.cloudflare.com/myapp:v1" → adds accountId
+ * - External registries like "docker.io/nginx:1.25" → pass through unchanged
+ */
+export function resolveImageName(accountId: string, image: string): string {
+  const cfRegistry = getCloudflareContainerRegistry();
+
+  // Check if image has a registry prefix (contains a dot in the first segment)
+  const firstSegment = image.split("/")[0];
+  const hasRegistryPrefix = firstSegment.includes(".");
```
Mkassabov: Is this a safe check? What if a user includes a `.` in their image name?

alchemy/src/cloudflare/container.ts
```diff
@@ -229,6 +279,76 @@ export async function Container<T>(
...
+  // This matches wrangler's behavior where registry URIs are passed directly
+  // to the Cloudflare API without any local Docker operations
+  if (props.image && !props.build) {
+    const rawImageRef =
+      typeof props.image === "string" ? props.image : props.image.imageRef;
+
+    if (isDev) {
+      // For local dev with prebuilt images, we need to pull and re-tag
+      // UNLESS it's a Cloudflare registry image (which would 401)
+      if (isCloudflareRegistryLink(rawImageRef)) {
```
Mkassabov, line 370: I wonder if there is a better dev behavior we can do here

alchemy/test/cloudflare/container.test.ts
```diff
@@ -346,5 +348,96 @@ describe.sequential("Container Resource", () => {
...
+      const containerName = `${BRANCH_PREFIX}-prebuilt-cf-image`;
+      const cfRegistry = getCloudflareContainerRegistry();
+
+      try {
+        // Use a CF registry image reference - should NOT attempt to pull
+        // This would cause a 401 if pulled, but should pass since we skip the pull
+        const container = await Container(containerName, {
+          className: "TestContainer",
+          name: containerName,
+          image: `${cfRegistry}/${api.accountId}/test-image:v1`,
```
sam-goodwin: Is this in everyone's account?

Mkassabov (approved):
A little concerned with how big the diff is for what seems like a small change but it looks good. Assuming tests pass this is good to merge.

Curious about that test image

alchemy/test/cloudflare/container.test.ts
Mkassabov: Did we ever get an answer to this? I see it's still in the diff @adunne09

adunne09: as far as I can tell- yes. its used throughout the codebase

https://github.com/search?q=repo%3Aalchemy-run%2Falchemy%20api.accountId&type=code

adunne09:
> A little concerned with how big the diff is for what seems like a small change but it looks good. Assuming tests pass this is good to merge.
> 
> Curious about that test image

yeah there's a lot of comments, i can clean those up if desired.

to address your feedback https://github.com/alchemy-run/alchemy/pull/1292/changes/BASE..4bcba262b94b15cbab75d6fe514c51c4875c7510#r2659859673 - I extended the changes past updating the container deployment logic to also mirror wrangler dev behavior so that overall parity is achieved (or at least closer).

alchemy/test/cloudflare/container.test.ts
Mkassabov: We are asking about the `test-image:v1` being on everyone's account, not the account id existing on API.

adunne09: yes, that is handled here: https://github.com/alchemy-run/alchemy/pull/1292/changes#diff-149600075d4f46904c1a039c10e8ef99baf37d63b61c2088d685dfe2e42cc1daR476

sam-goodwin:
Don't know why it took so long to merge this. Merging and releasing

PR #1302, opened 2026-01-17
fix(cloudflare): reflect actual queue consumer defaults in jsdoc

sbekrin:


alchemy/src/cloudflare/queue-consumer.ts
```diff
@@ -20,7 +20,7 @@ export interface QueueConsumerSettings {
 
   /**
    * Maximum number of concurrent consumer worker invocations
-   * @default 2
+   * @default undefined
```
sam-goodwin, line 23: is the default "no maximum"?

sbekrin: @sam-goodwin cf [states](https://developers.cloudflare.com/queues/configuration/configure-queues/#:~:text=be%20created%20automatically.-,max_concurrency,-number) that leaving it out will use maximum limit, which is [currently 250](https://developers.cloudflare.com/queues/platform/limits/#:~:text=Concurrent%20consumer%20invocations), but the phrasing suggests it might change hence i used `undefined`

PR #1305, opened 2026-01-20
Fix infinite dev loop for Astro request handler error

### LINKED ISSUE #1293: [Astro] Error in .astro page causes reloading loop
Steps to reproduce:

- clone https://github.com/macklinu/alchemy-astro-monorepo
- run `bun i`
- run `bun dev`
- visit `http://localhost:4321/oops`

Expected: page loads _one time_ with Astro dev error overlay

Actual: page loads _in an infinite loop_ with Astro dev error overlay

This makes local development with alchemy + astro more challenging, as I need to restart my dev server every time an error occurs in a .astro page/route handler. It could also be related to the fact that I'm using bun monorepo + turbo + alchemy + astro in some way? Any ideas on what the issue might be? Thanks!

https://github.com/user-attachments/assets/cedb7189-fecc-416e-8e4d-ee06919863ce

Versions:
- alchemy 0.83.0
- bun 1.3.5
- turbo 2.7.2

macklinu:
Resolves #1293

👋 I explored the `.alchemy` folder and noticed the `./alchemy/pids/*.pids.json` file was being rewritten over and over indefinitely if my Astro page request handler threw an error while developing with `astro dev`. While this _does_ fix my issue (assisted by Claude), I don't know what potential ramifications this might have.

Please take a look and let me know if there is a better approach to this - maybe a debouncing or some other conditional call to `persist()` at this juncture would make more sense (requesting the owners of this project with more codebase knowledge to chime in).


john-royal:
Replacing with #1308 which solves the root cause

PR #1308, opened 2026-01-21
fix(cloudflare): ignore .alchemy directory in astro integration

### LINKED ISSUE #1293: [Astro] Error in .astro page causes reloading loop

john-royal:
Closes #1293.

Uses the same `server.watch.ignored` option as our Vite plugin. Astro lets you modify Vite settings directly but wondering if there might be something in the middle that would cause this not to work. Need to test.

May be a better solution than #1305 because it addresses the root cause: the dev server shouldn't be watching `.alchemy`. @macklinu would you mind trying out the preview package and letting me know if this fixes it?


alchemy/src/cloudflare/astro/plugin.ts
```diff
@@ -7,13 +7,27 @@ const isAstroCheck =
...
+  integration.hooks["astro:config:setup"] = async (options) => {
+    options.updateConfig({
+      vite: {
+        server: {
+          watch: {
+            ignored: ["**/.alchemy/**"],
+          },
+        },
+      },
+    });
```
sam-goodwin, line 27: Does this merge in with existing settings?

john-royal: [Yes](https://docs.astro.build/en/reference/integrations-reference/#updateconfig-option)

PR #1311, opened 2026-01-26
chore: update cloudflare deps, oxfmt, and rwsdk

john-royal:
Formatting in my editor wasn't working for me. Updating the `oxfmt` version fixed it.

Updating the rwsdk version fixed the smoke test.

The file changes are all just formatting, save for one thing to fix a rwsdk type error.


examples/cloudflare-redwood/package.json
```diff
@@ -26,7 +26,7 @@
     "alchemy": "workspace:*",
     "dotenv": "^16.5.0",
     "drizzle-orm": "^0.44.2",
-    "rwsdk": "^0.1.36"
+    "rwsdk": "^1.0.0-beta.51"
```
sam-goodwin, line 29: will this work?

john-royal: Yes it fixes the redwood test always failing - this is the version you get during a regular npm install

PR #1313, opened 2026-02-02
fix(cloudflare): use hashed paths for wasm imports to fix next.js deploy

john-royal:
Looks like Next.js messed with their WASM imports again! As a result, WASM files were being written to `.alchemy/node_modules/.bun/next@16.1.6+fdf525fa27021d59/node_modules/next/dist/compiled/@vercel/og/resvg.wasm` which was causing a Cloudflare internal error on upload.

This matches Wrangler bundling behavior by copying the WASM files to `<outdir>/<hash>.wasm`, which fixes the Next.js deploy problem.


Mkassabov (approved):
nextjs example passed

PR #1314, opened 2026-02-03
fix(cloudflare): update writeMiniflareResponseToNode() to handle multiple `set-cookie` headers

samcxps:
### Summary
This PR fixes a bug where multiple Set-Cookie headers emitted by a worker were being collapsed so only the last cookie reached the client when running alchemy dev. The change ensures all Set-Cookie headers returned by a worker are forwarded correctly to the client.

### Problem
When proxying responses from a worker during development, the server previously overwrote the Set-Cookie header instead of preserving multiple values. This is a result of the [setHeader()](https://nodejs.org/api/http.html#requestsetheadername-value) method on Node.js `http.ServerResponse` which will overwrite headers of the same name. If a worker returned multiple Set-Cookie headers (e.g. for different cookies or cookie attributes), only the final header value was sent to the browser and earlier cookies were dropped.


alchemy/src/cloudflare/miniflare/miniflare-worker-proxy.ts
```diff
@@ -191,12 +191,23 @@ const writeMiniflareResponseToNode = (
...
   res.headers.forEach((value, key) => {
     // The `transfer-encoding` header prevents Cloudflare Tunnels from working because of the following error:
     // Request failed error="Unable to reach the origin service. The service may be down or it may not be
     // responding to traffic from cloudflared: net/http: HTTP/1.x transport connection broken: too many
     // transfer encodings: [\"chunked\" \"chunked\"]"
-    if (key !== "transfer-encoding") {
+    //
+    // Also ignore set-cookie since we handle it above
+    if (key !== "transfer-encoding" && key !== "set-cookie") {
       out.setHeader(key, value);
```
john-royal: Question: have you tried using `out.appendHeader` instead? Wonder if that would eliminate the need for cookie-specific handling.

samcxps: I have not. I will test though.

samcxps: `appedHeader()` worked as well. Updated PR to reflect.

PR #1315, opened 2026-02-04
feat(cloudflare): add Send Email support to Worker

harshil1712:


alchemy/src/cloudflare/worker-metadata.ts
```diff
@@ -339,13 +348,46 @@ export async function prepareWorkerMetadata(
...
+  }
+
   // Prepare metadata with bindings
   const meta: WorkerMetadata = {
     compatibility_date: props.compatibilityDate,
     compatibility_flags: props.compatibilityFlags,
     tail_consumers: props.tailConsumers?.map((consumer) =>
       isWorker(consumer) ? { service: consumer.name } : consumer,
     ),
+    send_email: props.sendEmail?.map((config) => ({
```
john-royal, line 385: Is this always at the top level instead of under `bindings`?

I understand it's at the top level in a wrangler.jsonc but the API docs seem to indicate this goes in the `bindings` array of the worker metadata object ([link](https://developers.cloudflare.com/api/resources/workers/subresources/scripts/methods/update/)).

alchemy/test/cloudflare/send-email.test.ts
john-royal, line 1: Were you able to run these? No problem if not, just wondering

alchemy/src/cloudflare/worker.ts
```diff
@@ -601,6 +609,36 @@ export interface WorkerPlacementHostname {
...
+
+  /**
+   * Restrict emails to a single destination address (mutually exclusive with allowedDestinationAddresses)
+   */
+  destinationAddress?: string;
+
+  /**
+   * Allowlist of destination addresses (mutually exclusive with destinationAddress)
+   */
+  allowedDestinationAddresses?: string[];
```
john-royal: This is a bit of a nitpick but generally if properties are mutually exclusive, we like to model that at the type level.

alchemy/test/cloudflare/send-email.test.ts
```diff
@@ -0,0 +1,222 @@
...
+        id: string;
+        send_email?: Array<{
+          name: string;
+          destination_address?: string;
+          allowed_destination_addresses?: string[];
+          allowed_sender_addresses?: string[];
+        }>;
+      }>(
+        `get worker "${name}"`,
+        api.get(`/accounts/${api.accountId}/workers/workers/${name}`),
```
john-royal, line 218: Love to see the new Workers API being used!

alchemy/src/cloudflare/worker-metadata.ts
harshil1712: Yes, it is at the top level. The API might change, I am not 100% sure.

alchemy/test/cloudflare/send-email.test.ts
harshil1712: It was missing the remote flag. I have now added that. Will test, and let you know.

john-royal:
Appreciate the contribution but going to have to close this for now. I'll incorporate what you've done into #1319 and make sure you're credited as well.

Just a note: in Alchemy, any bindings should always be defined in `WorkerProps["bindings"]` - this ensures consistency and lets us infer the type of your `env` object, which is important to us (Alchemy favors inference over codegen as a design tenet).

Happy to discuss further here or in Slack/Discord.

PR #1317, opened 2026-02-04
feat(cloudflare): AI Search

G4brym:
Main changes are:
- target branch was using user owned tokens, this one now uses account owned tokens (there was a issue on cf side that prevented this previously)
- Remove rag oauth scopes and add ai-search scopes (i also updated scopes for the alchemy app on cf side)

### DIFF
--- a/alchemy/src/cloudflare/ai-search.ts
+++ b/alchemy/src/cloudflare/ai-search.ts
@@ -418,7 +418,7 @@ export const AiSearch = Resource(
     let instance: AiSearch.ApiResponse;
     if (this.phase === "update" && this.output?.id) {
       const replace =
-        "type" in this.output &&
+        "source" in this.output &&
         (payload.type !== this.output.type ||
           payload.source !== this.output.source);
       // the development version of this resource had different properties, so check those to avoid an unnecessary replacement


alchemy/src/cloudflare/bucket.ts
```diff
@@ -689,16 +689,23 @@ const _R2Bucket = Resource(
 
     if (this.phase === "delete") {
       if (props.delete !== false) {
+        // Use bucket's actual jurisdiction from output, falling back to props
+        // This ensures EU/FedRAMP buckets are correctly addressed during cleanup
+        const deleteProps: BucketProps = {
+          ...props,
+          jurisdiction: this.output?.jurisdiction ?? props.jurisdiction,
+        };
+
```
john-royal, line 698: Good catch here :)

G4brym:
Hey, asked another team member to send a message to you in slack, and unfortunately since there is no cloudflare oauth scope to create account owned tokens, it will be difficult to support ai search into alchemy today.
It worked on my local setup because i was using an api token with permission to create other tokens, so the e2e tests passed.
Just as a bit of context, why ai search need a service token: https://developers.cloudflare.com/ai-search/configuration/service-api-token/
We are working on something internally that will solve this issue, will post back in this pr in a few weeks

alchemy/src/scope.ts
```diff
@@ -342,12 +342,21 @@ export class Scope {
...
+    let name = [this.appName, ...this.chain.slice(2), id, this.stage]
+      .map((part) => part.replaceAll(/[^a-z0-9_-]/gi, delimiter))
       .join(delimiter);
+    while (maxLength && name.length > maxLength) {
+      name = name
+        .split(delimiter)
+        .map((part) => part.slice(0, -1))
+        .join(delimiter);
+    }
+    return name;
```
john-royal, line 359: @sam-goodwin @Mkassabov What do you think about this?

The reason: AI search names have a max length of 32, but I was getting 37-40 character names in examples if the result from this `createPhysicalName` function wasn't truncated. But this approach is a little crude - results in names like `cloudflar-a-searc-searc-johnroya`. Any better ideas?

Mkassabov: I'd rather see `cloudflare` shortened to `cf` and `ai-search` shortened to `aisearch` or even `ais`; that way you get `cf-ais-search-johnroyal` which is a short enough prefix we can go back to using max len 32

john-royal: `cloudflare` is part of the app name. The app name was `cloudflare-ai-search`, the ID was `search`, and the stage `johnroyal`.

This change is for the scope `createPhysicalName` function, to let it generate a name with a max length. My guess is that if I ran into it in examples, users would have a similar issue. But app, logical ID, and stage are decided by the user, so simply swapping out "cloudflare" or "ai-search" wouldn't work.

Are you proposing we have a dictionary of potential abbreviations to run every name through?

Mkassabov: Nah didn't realize that was the app name.

IIRC what sst does is it just truncates with enough room to add an id at the end for differentiation.

Idk if we want to do that but it's an approach

PR #1318, opened 2026-02-05
fix(cloudflare-state-store): clear scope state when forceUpdate is en…

acoyfellow:
…abled

**Problem:**
When deploying with forceUpdate: true, the CloudflareStateStore updates the worker but doesn't clear the cached state data in the Durable Object. This causes deployments to fail when trying to delete resources that no longer exist (404 errors on D1 databases, workers, etc).

**Solution:**
1. Add a new 'clearScope' operation to SQLiteStateStoreOperations that deletes all resources in the current scope
2. Invoke clearScope automatically when provision() is called with forceUpdate: true
3. Log the clear operation for visibility

**Example:**
User deletes D1 database manually and recreates with new UUID. Alchemy state still has old UUID. Deploy fails trying to delete old DB. With this fix: forceUpdate: true clears the state, deploy succeeds.

**Testing:**
Tested on myfilepath.com project where D1 database ID changed from 3fc71d65-c2e2-44af-bdf6-1b1b3a6cdc99 to 11c62299-1d8c-418f-b250-ff2598c699c6. Before fix: Deploy failed with 404 on old database. After fix: State cleared, deploy proceeds with new database.

Fixes issue where forceUpdate doesn't actually force a clean slate.


john-royal:
I understand the issue you ran into - drift is a classic issue in IaC and every "solution" has tradeoffs.

The `forceUpdate` option is meant to force a re-deploy of the state store worker itself (mainly useful if you lose your token), not to modify the data in the state store itself. We probably could've named the property better, but adding this behavior to an existing property would be a potentially dangerous breaking change.

Happy to talk about alternative solutions. For example, would a CLI command that wipes your state be acceptable?

acoyfellow:
> I understand the issue you ran into - drift is a classic issue in IaC and every "solution" has tradeoffs.
> 
> The `forceUpdate` option is meant to force a re-deploy of the state store worker itself (mainly useful if you lose your token), not to modify the data in the state store itself. We probably could've named the property better, but adding this behavior to an existing property would be a potentially dangerous breaking change.
> 
> Happy to talk about alternative solutions. For example, would a CLI command that wipes your state be acceptable?

100% yes, i don't know why i didn't think to go that route in the first place!

took a stab at an approach: 
https://github.com/alchemy-run/alchemy/compare/main...acoyfellow:alchemy:feat/cli-state-clear

PR #1320, opened 2026-02-06
Add previews_enabled configuration option for Workers

kltdwrds:
## Summary

- Adds optional `preview_urls` boolean to `BaseWorkerProps` (matching Wrangler's config name)
- Threads through to `WorkerSubdomain`, which calls the Cloudflare subdomain API: `POST /accounts/{id}/workers/scripts/{name}/subdomain` with `{ enabled, previews_enabled }`
- Previously `enableWorkerSubdomain` hardcoded `previews_enabled: true` — now respects the user's setting

## Why this approach?

Preview URLs are **not** part of the multipart upload metadata (unlike `logpush`). They're managed via a separate POST to the `/subdomain` endpoint — this is exactly how [Wrangler handles it](https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/triggers/deploy.ts). The first commit in this PR incorrectly added it to metadata; the second commit fixes this.

## Usage

```ts
  preview_urls: false, // Required when using Durable Objects
  bindings: {
    MY_DO: DurableObjectNamespace("my-do", {
      className: "MyDurableObject",
    }),
  },
});
```

## Files changed

- **`worker.ts`** — Added `preview_urls?: boolean` to `BaseWorkerProps`, passed to `WorkerSubdomain`
- **`worker-subdomain.ts`** — Added `previewsEnabled` prop, threaded to `enableWorkerSubdomain()` instead of hardcoding `true`
- **`worker-metadata.ts`** — No net changes (reverted incorrect metadata approach)

## References

- [Cloudflare Preview URLs docs](https://developers.cloudflare.com/workers/configuration/previews/)
- [Limitation: no preview URLs with Durable Objects](https://developers.cloudflare.com/workers/configuration/previews/#limitations)
- [Wrangler config: `preview_urls`](https://developers.cloudflare.com/workers/wrangler/configuration/)

## Test plan

- [ ] Deploy a Worker with Durable Objects and `preview_urls: false` — succeeds
- [ ] Deploy a Worker without the flag — behavior unchanged (defaults to `true`)
- [ ] Deploy a Worker with `preview_urls: true` — preview URLs enabled

### DIFF
--- a/alchemy/src/cloudflare/worker-metadata.ts
+++ b/alchemy/src/cloudflare/worker-metadata.ts
@@ -204,7 +204,6 @@ export interface WorkerMetadata {
     };
   };
   logpush?: boolean;
-  previews_enabled?: boolean;
   migrations?: SingleStepMigration;
   main_module?: string;
   body_part?: string;
@@ -353,9 +352,6 @@ export async function prepareWorkerMetadata(
       enabled: observability?.enabled !== false,
     },
     logpush: props.logpush ?? false,
-    ...(props.previews_enabled !== undefined && {
-      previews_enabled: props.previews_enabled,
-    }),
     // TODO(sam): base64 encode instead? 0 collision risk vs readability.
     tags: [
       // encode a mapping table of Durable Object stable ID -> binding name
--- a/alchemy/src/cloudflare/worker-subdomain.ts
+++ b/alchemy/src/cloudflare/worker-subdomain.ts
@@ -35,6 +35,15 @@ interface WorkerSubdomainProps extends CloudflareApiOptions {
    * @internal
    */
   dev?: boolean;
+  /**
+   * Whether to enable preview URLs for this worker.
+   *
+   * When `undefined`, defaults to `true` (matching the behavior of enabling
+   * the workers.dev subdomain).
+   *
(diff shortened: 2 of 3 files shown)


kltdwrds:
Hey @sam-goodwin 👋 — would love your eyes on this when you get a chance. Unsure how best to test ...

We're using Alchemy to deploy Workers with Durable Objects at przm, and currently have to work around the lack of `preview_urls` support by manually PATCHing `/script-settings` with `{ previews_enabled: false }` before calling `Worker()`. It's fragile and races with Alchemy's own upload.

This PR adds `preview_urls` as an optional prop on `Worker`, threaded through to the subdomain API endpoint (which is how Wrangler handles it — not via upload metadata as I initially assumed).

cc @peterp — you helped us get Alchemy set up for preview deploys on przm, figured you'd want to know about this one.

Mkassabov:
Thanks for the contribution. We didn't see this until we had realized the issue was actually much deeper and required more of a change.

We've implemented your fix of adding a property to the worker (alongside everything else needed to fix this) in this PR here: https://github.com/alchemy-run/alchemy/pull/1321

So I'm gonna close this PR in favor of our in house one, but the change was very much appreciated. thank you!

PR #1330, opened 2026-02-13
fix(state): set sqlite busy_timeout to 5s

julienroubieu:
When running multiple apps sharing the same SQLite state store file, concurrent access causes errors. This has been biting us in a monorepo setup with `alchemy dev`.

```
LibsqlError: SQLITE_BUSY_RECOVERY: database is locked
```

This PR adds an arbitrary 5s delay before SQLite returns an error.

### DIFF
--- a/alchemy/src/state/sqlite-state-store.ts
+++ b/alchemy/src/state/sqlite-state-store.ts
@@ -162,6 +162,7 @@ async function createBunSQLiteDatabase(
     create: true,
     ...bunOptions,
   });
+  client.exec("PRAGMA busy_timeout = 5000;");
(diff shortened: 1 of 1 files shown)


Mkassabov:
running tests on this right now; I expect it to be fine though

Mkassabov:
might be some delay on this, we are having issues with our testing account

gingerhendrix:
@Mkassabov can we get a pkg.pr.new?

Mkassabov:
@gingerhendrix 

> @Mkassabov can we get a pkg.pr.new?

done. apologies on the delay, dropped this over the weekend, running tests now will likely merge shortly

PR #1347, opened 2026-02-26
chore: build improvements

aleclarson:
### Minor Changes

* **Formatted `package.json`** using `oxc` (c65b7d5).
* **Tweaked VSCode workspace settings** to improve the format-on-save experience (571d09b).
* **Cleaned up publish script** by removing a duplicate `tsdown` call (dd2e361).

### Main Changes (aa1f6e1)

**Architecture & Imports**

* **Refactored CLI imports (`bin/*.ts`):** Replaced relative imports (e.g., `../../src/*`) with self-referencing imports (`alchemy/private/*`).
* **The Problem:** The old method caused duplicate module instantiations and bloated the npm package with duplicate code (one version in `lib/`, another bundled in the CLI).
* **The Solution:** Added `"alchemy": "link:."` to `devDependencies` so the CLI can reference itself seamlessly during local development and rely on standard `node_modules` resolution once installed.

* **Disabled dependency bundling** into `bin/alchemy.js`, allowing package managers to effectively use their global caches.

**Build & NPM Package Optimization**

* **Trimmed package payload:** Removed `src/*`, `*.map`, and `*.ts` files from the final npm package.
* **Removed Bun entry points:** These were stripped out since they relied on the removed `*.ts` files.
* **Disabled map generation for production:** Sourcemaps and declaration maps are no longer generated during the pre-publish build step (they remain active during development).

**Developer Experience**

* **Enhanced `dev` script:** Integrated [Picorun](https://github.com/alloc/picorun) to run both `tsc -b` and `tsdown` concurrently in watch mode during local development. *(Note: Can be swapped for an alternative runner if preferred).*


aleclarson:
**Note:** We still need to update the Vitest configuration if we want Vitest to compile the `alchemy` CLI source code on the fly during testing.

.vscode/settings.json
john-royal: Good choice, been meaning to make these changes.

PR #1348, opened 2026-02-26
feat: include healthcheck status in Docker container `inspect()` result

aleclarson:
Feel free to edit this one as you see fit!

### Summary
This PR introduces support for monitoring and waiting on Docker container health checks. It updates the internal Docker API types to expose health status and adds a new method to the `Container` resource to synchronize actions based on a container's health state.

### Key Changes
*   **API Enhancements**: Updated `ContainerState` to include the `Health` object, which tracks status (`starting`, `healthy`, `unhealthy`), failure counts, and log output.
*   **New Method: `waitForHealth()`**: Added a method to the `Container` class that polls the container's state until it reports as `healthy`. 
    *   Throws an error if the container is marked `unhealthy`.
    *   Throws an error if the container does not have a health check configured.
*   **Testing**: Added a comprehensive test suite in `container.test.ts` to verify the inspection logic and `waitForHealth` behavior across various states (healthy, unhealthy, starting, and unconfigured).

### Impact
This allows users to ensure a service inside a container is fully operational and passed its internal health checks before proceeding with dependent tasks, rather than relying solely on the container's "running" status.

### DIFF
--- a/alchemy/src/docker/container.ts
+++ b/alchemy/src/docker/container.ts
@@ -340,7 +340,9 @@ export const Container = Resource(
       return toRuntimeInfo(info);
     };
 
-    const waitForHealth: Container["waitForHealth"] = async (timeout = 60000) => {
+    const waitForHealth: Container["waitForHealth"] = async (
+      timeout = 60000,
+    ) => {
       const startTime = Date.now();
       while (Date.now() - startTime < timeout) {
         const [info] = await api.inspectContainer(containerName);
--- a/alchemy/src/docker/container.ts
+++ b/alchemy/src/docker/container.ts
@@ -218,6 +218,12 @@ export interface Container extends ContainerProps {
    * Inspect the container to get detailed information
    */
   inspect(): Promise<ContainerRuntimeInfo>;
+
+  /**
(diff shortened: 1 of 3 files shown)


alchemy/src/docker/container.ts
```diff
@@ -325,6 +331,46 @@ export const Container = Resource(
...
+          );
+        }
+
+        // Wait 500ms before next check
+        await new Promise((resolve) => setTimeout(resolve, 500));
+      }
+      throw new Error(
+        `Timed out waiting for container ${containerName} to become healthy`,
+      );
+    };
```
Mkassabov, line 372: I'm not sure if `waitForHealth` makes sense as an alchemy feature.
1. a service can be unhealthy and then transition to healthy (which might not make sense here) so throwing an error on unhealthy seems a little bit unexpected.
2. I think wait for checks often have enough logic tied into them that a generic solution might not be great (e.g. timeout being 500)
3. this almost makes more sense as part of the resources creation lifecycle rather than a method on it.

Could you briefly explain why you want this as part of alchemy?

PR #1349, opened 2026-02-26
feat: handle existing Docker network creation gracefully

aleclarson:
Feel free to edit this as you see fit!

### Summary
This PR makes Docker network creation idempotent. Instead of failing when a network with the same name already exists, the system now gracefully handles the conflict by inspecting the existing network and adopting its configuration.

#### `DockerApi` Enhancement
- Introduced a `NetworkInfo` type to represent Docker network metadata.
- Added `inspectNetwork(nameOrId)` to retrieve details for a specific network.
- Added `networkExists(name)` helper method to check for network availability.

#### `Network` Resource Logic
- Updated the `create` method to catch "already exists" (409 Conflict) errors from the Docker API.
- If a conflict occurs, the resource now performs an inspection to retrieve the existing network's ID and properties, ensuring the internal state is correctly populated.

#### Testing
- Added a new test suite in `Network.test.ts`.
- Verified that the resource successfully creates a new network when one doesn't exist.
- Verified that the resource correctly attaches to an existing network when a name collision occurs.

### DIFF
--- a/alchemy/src/docker/network.ts
+++ b/alchemy/src/docker/network.ts
@@ -114,7 +114,9 @@ export const Network = Resource(
       } catch (error: any) {
         const errorMessage = error.message || String(error);
         if (
-          errorMessage.includes(`network with name ${networkName} already exists`)
+          errorMessage.includes(
+            `network with name ${networkName} already exists`,
+          )
         ) {
           const [details] = await api.inspectNetwork(networkName);
           if (details) {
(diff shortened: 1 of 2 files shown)


Mkassabov (commented):
@aleclarson Do we need changes to tests / and or needing new tests here?

Mkassabov:
> @aleclarson Do we need changes to tests / and or needing new tests here?

oh I'm blind this has test I didn't see that... will run tests soon and approve if they merge

PR #1351, opened 2026-03-01
chore: fix onExit missing error

Mkassabov:
IDK how needed this change is. alchemy was working fine for me but tests were all failing with this error about onExit not existing.

Claude tracked it down to using `proper-lockfile` depending on `signal-exit` v3 but if we downgrade `proper-lockfile` to v3 it breaks `docker/api.ts` in incomprehensible ways;

the biggest change that v4 breaks is just the way it exports onExit so patching proper-lockfile is quite a simple fix.

but I also absolutely hate this so if anyones got better ideas...

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -816,6 +816,9 @@
       },
     },
   },
+  "patchedDependencies": {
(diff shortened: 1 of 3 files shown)


john-royal:
"tests are running fine on main now"

PR #1358, opened 2026-03-10
feat(astro): update Cloudflare support for Astro 6

AmanVarshney01:
Updates Alchemy's Astro Cloudflare support to match Astro 6 output and adapter conventions.

## Changes

- switch Astro Cloudflare server defaults to `dist/server/entry.mjs`
- switch Astro Cloudflare assets defaults to `dist/client`
- update Astro package support to Astro 6 and `@astrojs/cloudflare` 13
- update the Astro template package versions and Wrangler template paths
- normalize the Astro plugin's local Cloudflare proxy config for the v13 adapter API


Mkassabov (commented):
could this also update the cloudflare-astro example?

examples/cloudflare-astro/package.json
```diff
@@ -1,26 +1,28 @@
...
   "private": true,
   "version": "0.0.0",
   "type": "module",
   "scripts": {
     "build": "astro check && astro build",
-    "deploy": "alchemy deploy --env-file ../../.env",
-    "destroy": "alchemy destroy --env-file ../../.env",
-    "dev": "alchemy dev --env-file ../../.env",
+    "deploy": "alchemy deploy",
+    "destroy": "alchemy destroy",
```
sam-goodwin: this might break our tests

alchemy/src/cloudflare/astro/astro.ts
```diff
@@ -86,8 +86,16 @@ export async function Astro<B extends Bindings>(
     dev: spreadDevProps(props, `${runner} astro dev`),
     entrypoint:
       props.entrypoint ??
-      (output === "server" ? "dist/_worker.js/index.js" : undefined),
-    assets: props.assets ?? "dist",
+      (output === "server" ? "dist/server/entry.mjs" : undefined),
```
sam-goodwin, line 89: Does it still work for older astro?

AmanVarshney01: Yeah, this would break Astro 5.

Astro 5 still outputs:

`dist/_worker.js/index.js`

I’m not sure we can reliably infer v5 vs v6 inside the resource

I thinkmaybe we just update the docs?

keep these defaults aligned to Astro 6
document the Astro 5 override explicitly

Something like:
```
await Astro("website", {
  entrypoint: "dist/_worker.js/index.js",
  assets: "dist",
  wrangler: {
    main: "@astrojs/cloudflare/entrypoints/server.js",
  },
});
```

Mkassabov: maybe a version prop? we default to `astroVersion: 5` so we don't break it for existing users and anyone who wants to use astro 6 would explicitly say `astroVersion: 6`.

or we just version the entire resource but that feels like overkill. We should really avoid breaking userspace

jcheese1:
upvote this

PR #1374, opened 2026-04-02
Fix: Ignore timestamps when comparing props for consistency

gingerhendrix:
Prevents getting stuck in a 'Resource ... is not in a stable state after...' loop.

### DIFF
--- a/alchemy/src/apply.ts
+++ b/alchemy/src/apply.ts
@@ -144,16 +144,20 @@ async function _apply<Out extends ResourceAttributes>(
       ) {
         // Remove timestamps from props before comparing, as they can cause non-deterministic differences
         const stripNonDeterministic = (obj: any): any => {
-          if (!obj || typeof obj !== 'object') return obj;
+          if (!obj || typeof obj !== "object") return obj;
           if (Array.isArray(obj)) return obj.map(stripNonDeterministic);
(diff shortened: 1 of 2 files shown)


Mkassabov (commented):
Hmmm not sure this looks like the right change.

IIRC we use updatedAt and createdAt in monorepos to wait for a consistent state across projects in a monorepo; this might cause damage to that process.

This change would also break any resources that legitimately have a need for updatedAt and createdAt; hardcoding prop names doesn't seem like a good idea since it introduces some "magic" to whats going on here.

can you provide more details on the resources causing this to happen as I suspect its a resource specific issue in 1 or more resources?

mgandolfi7:
I was able to [reproduce](https://github.com/mgandolfi7/alchemy-race-condition-repro) this with turborepo, pnpm, and node by adding a durable object binding to the [**Example Monorepo**](https://github.com/alchemy-run/alchemy/tree/main/example-monorepo).

Mkassabov:
We had a conversation about this in discord: https://discord.com/channels/1359694195782320389/1489262247426527252

TL;DR
this change wouldn't actually do anything since alchemy waits for consistency; it would just paper over the actual error and introduce a race-condition somewhere else.
the root problem here is the root resource in the example not being stable

PR #1379, opened 2026-04-06
fix: resolve zone domain names to IDs in EmailRule and EmailRouting

### LINKED ISSUE #1378: EmailRule and EmailRouting should resolve zone domain names to zone IDs
## Problem

`EmailRule` and `EmailRouting` accept a `zone` parameter typed as `string | Zone`. When a string is passed, the implementation uses it directly as the zone ID:

```typescript
const zoneId = typeof props.zone === "string" ? props.zone : props.zone.id;
```

This means passing a domain name like `"example.com"` results in API calls to `/zones/example.com/email/routing/rules`, which returns a 404:

```
CloudflareApiError: Error 404 create email rule: 7003: Could not route to /client/v4/zones/sauna.dev/email/routing/rules, perhaps your object identifier is invalid?
```

## Expected behavior

Passing a domain name string should work, consistent with the documentation examples:

```typescript
await EmailRouting("custom-email-routing", {
  zone: "example.com",
  enabled: true,
});
```

Other Alchemy resources like `CertificatePack` already resolve domains to zone IDs using `findZoneForHostname()`. `EmailRule` and `EmailRouting` should do the same when the string doesn't look like a zone ID.

## Workaround

Use a `Zone` resource to resolve the domain first:

```typescript
const zone = await Zone("my-zone", { name: "example.com" });

await EmailRule("my-email-rule", {
  zone,
  // ...
});
```

## Affected resources

- `EmailRule` (`src/cloudflare/email-rule.ts`)
- `EmailRouting` (`src/cloudflare/email-routing.ts`)

nyacg:
## Summary
- `EmailRule` and `EmailRouting` now resolve zone domain name strings (e.g. `"example.com"`) to zone IDs, matching the existing pattern in `RedirectRule`
- Previously, passing a domain name string resulted in API calls to `/zones/example.com/...` which returned 404

## Changes
- Apply the same zone resolution logic from `RedirectRule` to both `EmailRule` and `EmailRouting`: if the string contains a `.`, resolve it via `getZoneByDomain`; otherwise treat it as a zone ID
- Add error handling for unresolvable domains

Fixes #1378


Mkassabov (changes requested):
can you please add test cases for this. otherwise great change!

sam-goodwin:
Was this resolved by https://github.com/alchemy-run/alchemy/pull/1388?

image

sam-goodwin:
Closing as I believe this was fixed by another PR that also added EmailSender binding

PR #1380, opened 2026-04-07
feat(cloudflare)!: scope workflow names with app and stage

Mkassabov:
Workflow names now default to `${app}-${id}-${stage}` to avoid cross-stage collisions in ephemeral environments.


alchemy/src/cloudflare/workflow.ts
```diff
@@ -62,6 +62,12 @@ export type Workflow<PARAMS = unknown> = {
   limits?: {
     steps?: number;
   };
+  /**
+   * Whether the workflow name was explicitly provided by the user.
+   * When false, the Worker resource will generate a scoped name.
+   * @internal
+   */
+  _nameExplicit?: boolean;
```
sam-goodwin: Why do we need this?

Mkassabov: we dont. I had told claude to preserve backwards compat if it could and it did some dumbshit; pushed a cleaner fix

PR #1386, opened 2026-04-15
feat(cloudflare): add AI Search namespace + standalone binding support

G4brym:
## Summary

Adds `AiSearchNamespace` resource and the two official Cloudflare AI Search worker bindings (`ai_search` and `ai_search_namespace`), replacing the legacy `env.AI.autorag()` path.

## New resources & bindings

- **`AiSearchNamespace` resource** — CRUD lifecycle for AI Search namespaces, with support for:
  - `adopt: true` (pre-flight existence check, not status-code matching — no false positives on unrelated 400s)
  - Rename triggers replace (delete old + create new at scope finalize)
  - `delete: false` opt-out
  - The reserved `"default"` namespace (adopt-only; delete is silently skipped since it cannot be deleted via the API)
  - Three-way semantics on `description`: absent = leave untouched, `string` = set, `null` = explicit clear
- **`ai_search` Worker binding** — binds directly to a single AI Search instance. `Bound<AiSearch>` maps to the runtime `AiSearchInstance` class from `@cloudflare/workers-types`.
- **`ai_search_namespace` Worker binding** — scoped to a namespace for dynamic multi-instance access. `get(id) / list() / create() / delete() / search() / chatCompletions()` at runtime. `Bound<AiSearchNamespace>` maps to the runtime `AiSearchNamespace` class (name-collision handled via `_`-prefixed import aliases).

## `AiSearch` updates

- **Source is now optional** — omit `source` for a built-in-storage instance that accepts manual uploads via the items API.
- **`namespace: string | AiSearchNamespace`** prop (defaults to `"default"`); changing it on an existing instance triggers a replace.
- **Service tokens are only auto-created for R2 sources** — web-crawler and sourceless instances don't need them.
- **All API endpoints migrated** to the namespace-scoped URL shape (`/ai-search/namespaces/{ns}/instances/...`).

## Validation

- Single-instance `ai_search` bindings validate at deploy time that the bound instance is in the `default` namespace; error message points at the `AiSearchNamespace` binding alternative.
- Namespace names validated locally against CF's documented regex before any API call.

## Lifecycle hardening

- `deleteAiSearchInstance` and `deleteAiSearchNamespace` actively wait for `GET` to return 404 before returning, presenting a strongly-consistent "resource is gone" contract to callers (child token delete, user-facing teardown assertions, downstream resources).
- `deleteAiSearchToken` briefly retries on 7076 `token_in_use_by_instances` to cover residual cross-colo lag. A persistent 409 throws an actionable error — a post-retry 409 means the token is genuinely shared by another instance and should not be silently leaked.
- Child `AiSearchToken` honors the parent `AiSearch`'s `delete: false` so preserving an instance also preserves its referencing token.

## Usage

```ts
import { AiSearch, AiSearchNamespace, Worker } from "alchemy/cloudflare";

// Namespace binding — dynamic multi-instance access
const ns = await AiSearchNamespace("docs", { name: "docs" });

// Single instance (built-in storage, no R2 or token needed)
const search = await AiSearch("kb", { name: "knowledge-base" });

// Instance in a custom namespace
const search2 = await AiSearch("blog", {
  name: "blog",
  namespace: ns,
});

  bindings: {
    SEARCH: search,  // ai_search binding
    DOCS: ns,         // ai_search_namespace binding
  },
});
```

```ts
// Worker runtime
// Single instance binding — simple query shape
const results = await env.SEARCH.search({ query: "How does caching work?" });
await env.SEARCH.items.upload("faq.md", "# FAQ\n\n...");

// Namespace binding — access instances dynamically
const blog = env.DOCS.get("blog");
const stream = await blog.chatCompletions({
  messages: [{ role: "user", content: "explain caching" }],
  stream: true,
});

// Namespace-level operations
const instances = await env.DOCS.list();
await env.DOCS.create({ id: "tenant-123" });
await env.DOCS.delete("old-docs");
```

## Test coverage

24 tests against the real Cloudflare API. Highlights:

| Test | Coverage |
|---|---|
| `adopt preserves underlying AiSearch instance (createdAt + internalId stable)` | Catches silent-recreate regressions on adopt |
| `adopt preserves underlying namespace resource (createdAt match)` | Same, for namespaces |
| `rename triggers replace (old deleted, new created)` | Verifies delete-old-then-create-new semantics |
| `description round-trips through the API (set → update → clear with null)` | Three-way `description` semantics including `null` clear, round-tripped to the API |
| `adopt on reserved default namespace binds without create` | Reserved `default` namespace adopt + destroy no-op |
| `rejects names that violate Cloudflare's character rules` | Local regex validation (offline) |
| `delete=false preserves namespace on scope destroy` | Opt-out deletion pattern |
| `sourceless instance update changes maxNumResults without replacing the instance` | Sourceless = no token auto-created; updatable without replace |
| `changing namespace on an existing instance triggers replace` | Namespace immutability + replace correctness |
| `namespace prop accepts a string (not just a Resource)` | Both input shapes |
| `single-instance binding rejects non-default namespace at deploy time` | Deploy-time validation with actionable error |
| `AI Search with RAG response generation via Worker` | Full E2E of the (legacy) `env.AI.autorag()` path via deployed Worker |

Full suite runs in ~2 min against a live Cloudflare account.

## Files changed

| Category | Files |
|---|---|
| New resource | `alchemy/src/cloudflare/ai-search-namespace.ts` |
| Updated resource | `alchemy/src/cloudflare/ai-search.ts`, `ai-search-token.ts` |
| Binding integration | `bindings.ts`, `bound.ts`, `worker-metadata.ts`, `miniflare/build-worker-options.ts`, `wrangler.json.ts` |
| Exports | `index.ts` |
| Tests | `alchemy/test/cloudflare/ai-search.test.ts` (24 tests total) |
| Docs | new `ai-search-namespace.md`, updated `ai-search.md`, new `cloudflare-ai-search.mdx` guide |
| Example | `examples/cloudflare-ai-search/` (updated `alchemy.run.ts` + `worker.ts`, new `README.md`, registered in root tsconfig) |
| Deps | `@cloudflare/workers-types: ^4.20260420.1`, `wrangler: ^4.83.0` (for native miniflare AI Search support) |

## Notes

- Local dev: `miniflare@4.20260415.0+` (via `wrangler@^4.83.0`) natively proxies AI Search bindings to the deployed instance via `remote: true` — same mechanism as `ai`, `vectorize`, `browser`, etc.
- Binding resolution uses `isAiSearch()` / `isAiSearchNamespace()` type guards (via `ResourceKind`) since `AiSearch.type` is the *source* type (`"r2" | "web-crawler"`), not a binding discriminant.
- `bound.ts` places AI Search arms FIRST in the `Bound<T>` conditional chain (required for the same reason — `AiSearch.type` would otherwise false-match structural discriminators later in the chain).
- `AiSearchItems.upload()` accepts `ReadableStream | Blob | string` — docs and examples updated accordingly.

### DIFF
--- a/alchemy/src/cloudflare/bound.ts
+++ b/alchemy/src/cloudflare/bound.ts
@@ -1,6 +1,9 @@
+/// <reference types="@cloudflare/workers-types" />
 import type { Pipeline } from "cloudflare:pipelines";
 import type { Secret } from "../secret.ts";
 import type { Ai as _Ai } from "./ai.ts";
+import type { AiSearchNamespace as _AiSearchNamespace } from "./ai-search-namespace.ts";
+import type { AiSearch as _AiSearch } from "./ai-search.ts";
 import type { AnalyticsEngineDataset as _AnalyticsEngineDataset } from "./analytics-engine.ts";
 import type { Assets } from "./assets.ts";
 import type { Binding, Json, Self } from "./bindings.ts";
@@ -9,6 +12,7 @@ import type { R2Bucket as _R2Bucket } from "./bucket.ts";
 import type { Container as _Container } from "./container.ts";
 import type { D1Database as _D1Database } from "./d1-database.ts";
 import type { DurableObjectNamespace as _DurableObjectNamespace } from "./durable-object-namespace.ts";
+import type { EmailSender as _EmailSender } from "./email-sender.ts";
 import type { HyperdriveRef } from "./hyperdrive-ref.ts";
 import type { Hyperdrive as _Hyperdrive } from "./hyperdrive.ts";
 import type { Images as _Images } from "./images.ts";
@@ -18,7 +22,6 @@ import type { RateLimit as _RateLimit } from "./rate-limit.ts";
 import type { SecretKey } from "./secret-key.ts";
 import type { SecretRef as CloudflareSecretRef } from "./secret-ref.ts";
 import type { Secret as CloudflareSecret } from "./secret.ts";
-import type { EmailSender as _EmailSender } from "./email-sender.ts";
 import type { VectorizeIndex as _VectorizeIndex } from "./vectorize-index.ts";
 import type { VersionMetadata as _VersionMetadata } from "./version-metadata.ts";
 import type { VpcService as _VpcService } from "./vpc-service.ts";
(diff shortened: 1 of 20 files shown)


G4brym:
This is still very much WIP

G4brym:
this pr is now ready for review

G4brym:
Hey @sam-goodwin can you take a look at this pr?

sam-goodwin:
Have you run the tests and verified this works @G4brym

alchemy/src/cloudflare/bound.ts
```diff
@@ -39,78 +42,88 @@ type BoundWorker<
...
-        : T extends _Worker<any, infer RPC> | WorkerRef<infer RPC>
+// NOTE: AiSearch / AiSearchNamespace MUST come FIRST in this conditional
+// chain, before any structural discriminator like `{ type: "kv_namespace" }`.
+// `AiSearch.type` is the data-source type (`"r2" | "web-crawler"`), NOT a
+// binding-type discriminator — so we match by *resource-type import* via the
+// `_AiSearch` / `_AiSearchNamespace` aliases (underscore-prefixed to avoid
+// colliding with the identically-named Cloudflare runtime classes on the
+// right-hand side of the conditional).
+export type Bound<T extends Binding> = T extends _AiSearch
+  ? AiSearchInstance
```
sam-goodwin, line 53: What happened here? Is this assumeing they one day add `type: "kv_namespace"`?

PR #1319, opened 2026-02-05
feat(cloudflare): send email binding

john-royal:
Adapted #1315 to follow Alchemy conventions for bindings, account for local development, and use ([what I believe to be](https://github.com/alchemy-run/alchemy/pull/1315#discussion_r2769251581)) the correct Cloudflare API for bindings.

Treats `SendEmail` as an object that you call and pass in as a binding rather than a separate property*. For example:

```ts
const worker = await Worker("my-worker", {
  bindings: {
    EMAIL: SendEmail({
      allowedDestinationAddresses: ["john@chipotle.co.uk"]
    })
  }
});
```

* Still a separate property in `wrangler.jsonc`.

## Tasks

- [x] Remote implementation
- [x] Local implementation
- [ ] Docs and tests (will throw at Claude)
- [x] Scope (discuss at standup)

PR #1388, opened 2026-04-20
feat(cloudflare): send email binding support

Dawsson:
## Summary
- add a first-class `SendEmailBinding()` helper for Cloudflare Workers
- serialize `send_email` bindings in worker metadata, wrangler output, and local Miniflare config
- support `dev.remote: true` for local email sending through Cloudflare
- document the binding and add a focused wrangler serialization test

## Validation
- `bun test test/cloudflare/wrangler-json.test.ts -t "with send email binding"`
- `bun run --filter alchemy build`

## Notes
- I also updated the remote binding proxy so local remote email sends support both raw `EmailMessage` inputs and the object-builder overload accepted by Cloudflare's `SendEmail.send(...)`.


Dawsson:
Follow-up credit: I reviewed #1319 from @john-royal while finishing this Email Service work and pulled in a few ideas from that draft.

Specifically, this PR now incorporates the same general direction for:
- type-level exclusivity between `destinationAddress` and `allowedDestinationAddresses`
- the Miniflare logging tweak so local `send_email` activity is visible without noisy ready-line duplication
- a dedicated `examples/cloudflare-email` example package

The rest of the implementation here still differs in a few places, especially around the current `SendEmailBinding()` API shape, docs/tests completion, and the remote binding proxy support for both `EmailMessage` and builder-style sends. But #1319 was useful reference material and should be credited.

Dawsson:
image
image

examples/cloudflare-email/alchemy.run.ts
```diff
@@ -0,0 +1,22 @@
+import alchemy from "alchemy";
+import { SendEmailBinding, Worker } from "alchemy/cloudflare";
+
+const app = await alchemy("cloudflare-email");
+
+export const worker = await Worker("email-worker", {
+  entrypoint: "./worker.ts",
+  compatibility: "node",
+  bindings: {
+    EMAIL: SendEmailBinding({
```
sam-goodwin: Not sure I like the name, would `EmailSender` be better?

sam-goodwin: Maybe something like:
`EmailAddress`
`EmailSender`
`EmailRecipient`

Dawsson: John also used EmailSender, I don't personally have a preference. I asked Codex "While following naming conventions and standards of the app" it preferred SendEmailBinding. I'm happy to change that though, you guys obviously know this library better than I do

Dawsson:
Name has been updated, I left `sendEmailBindings` for example in `alchemy/src/cloudflare/wrangler.json.ts:266` as it utilizes Cloudflare's own `send_email` binding type

PR #1391, opened 2026-04-24
fix(cloudflare): repair AI Search binding support in `alchemy dev`

G4brym:
## Summary

Follow-up to #1386. Four bug fixes that together make AI Search bindings (`ai_search`, `ai_search_namespace`) actually work in `alchemy dev`. Before this change, running a Worker with either binding type in dev mode failed with a mix of CF API rejections and runtime `TypeError`s on `env.SEARCH` / `env.DOCS`.

Opened as a draft because fix #4 (Access probe) can't be end-to-end verified locally on my test account (unrelated ZT config issue) — the code change itself is correct and mirrors what wrangler's equivalent path does, but the specific runtime interaction needs a reviewer with a workers.dev subdomain not gated behind Access / WARP.

### 1. `raw: true` on `ai_search` / `ai_search_namespace` remote bindings — CF API rejects with 10333

Cloudflare's `/edge-preview` API rejects the `raw` property on these two binding types:

```
[10333] Binding 'DOCS' of type 'ai_search_namespace' does not support
        being configure with the 'raw' property.
```

Wrangler's [`create-worker-upload-form.ts`](https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/deployment-bundle/create-worker-upload-form.ts) deliberately omits `raw` for these two binding types (while keeping it for `ai`, `browser`, `vectorize`, etc.). This PR mirrors that shape when assembling `remoteBindings` for the `remote-binding-proxy` worker upload.

### 2. `scope.local` mock branches break the remote-binding-proxy flow — 10359 / 10360

`AiSearch` and `AiSearchNamespace` both had `if (this.scope.local) return mock;` short-circuits. But AI Search is an always-remote binding: the `remote-binding-proxy` worker needs the namespace/instance to **actually exist** on Cloudflare at preview-token creation time, or the binding validation fails with:

```
[10359] AI Search binding 'DOCS' references namespace 'foo' which does not exist.
[10360] AI Search binding 'SEARCH' references instance 'bar' … not found.
```

Removed the mock branches. Matches the existing pattern used by Vectorize, D1, R2, etc. (none of which mock in `scope.local`). Left the separate R2-bucket-needs-`dev: { remote: true }` guard in `ai-search.ts` in place — that's a distinct developer-UX check.

### 3. miniflare pinned too old — no `ai-search` plugin, so bindings are undefined at runtime

Alchemy pinned `miniflare: 4.20260310.0`. The `ai-search` plugin was added in [workers-sdk#13027](https://github.com/cloudflare/workers-sdk/pull/13027), merged 2026-03-27. Without it, miniflare has no idea what `aiSearchNamespaces` / `aiSearchInstances` options mean — `env.SEARCH` / `env.DOCS` are simply never populated in the Worker runtime and you get `TypeError: Cannot read properties of undefined`.

Bumped to `4.20260329.0` (first release containing the plugin).

### 4. Access detection can't see Access 302 on preview workers without the preview token

`domainUsesAccess(host)` does a bare HEAD probe to detect Access. For a workers.dev preview-worker URL, CF only surfaces the Access 302 when the request carries `cf-workers-preview-token`; a bare HEAD returns 404 regardless of Access configuration. So on accounts that gate workers.dev behind Access, alchemy would silently report "no Access", skip `cloudflared access login`, and every subsequent proxy request would 302-redirect to Access login and fail.

Also had to restructure `createRemoteProxyWorker` to obtain the preview token *before* the Access probe (they were running in `Promise.all` previously, so the probe never had the token available).

## Verification

- `bunx tsc --noEmit --project alchemy/tsconfig.json` — clean
- `bunx tsc --noEmit --project alchemy/tsconfig.test.json` — clean
- `bun format` — clean
- Ran 5 AI Search tests against real CF (adopt / name validation / namespace prop / binding integration) — all green
- Manually reproduced each bug path against a live CF account with the pkg.pr.new tarball of #1386, confirmed fix resolves it

Specifically for (1) and (2) I traced the failing `/edge-preview` POSTs via curl to confirm CF's exact error codes, and for (4) I captured the preview token from the running proxy and verified HEAD-with-token returns 302→cloudflareaccess.com while bare HEAD returns 404.

## Test snippet

```ts
import { AiSearch, AiSearchNamespace, Worker } from "alchemy/cloudflare";

const app = await alchemy("ai-search-dev-test");

const ns = await AiSearchNamespace("docs", {
  name: "my-docs",
  adopt: true,
});

// Instance in a custom namespace — exercised via env.DOCS.get(name)
const nsInstance = await AiSearch("ns-kb", {
  name: "my-ns-kb",
  namespace: ns,
  adopt: true,
});

// Instance in the default namespace — bound via single-instance ai_search
const defaultInstance = await AiSearch("default-kb", {
  name: "my-default-kb",
  adopt: true,
});

  entrypoint: "src/worker.ts",
  url: true,
  bindings: {
    SEARCH: defaultInstance,  // ai_search binding
    DOCS: ns,                  // ai_search_namespace binding
  },
});

await app.finalize();
```

Worker implementation:

```ts
export default {
  async fetch(request, env) {
    const url = new URL(request.url);
    if (url.pathname === "/search") {
      return Response.json(await env.SEARCH.search({ query: "hello" }));
    }
    if (url.pathname === "/ns/list") {
      return Response.json(await env.DOCS.list());
    }
    return new Response("ok");
  },
};
```

Before this PR: `alchemy dev` fails with 10333 (and even after bypassing that, 10359 / 10360 / undefined-binding).

After this PR: `alchemy dev` creates both namespace + instance on CF, uploads the remote-binding-proxy worker with valid binding metadata, miniflare wires `env.SEARCH` / `env.DOCS` through the proxy, and the Worker serves `/search` + `/ns/list`.

### DIFF
--- a/alchemy/package.json
+++ b/alchemy/package.json
@@ -195,7 +195,7 @@
     "glob": "^10.0.0",
     "jszip": "^3.0.0",
     "libsodium-wrappers": "^0.8.0",
-    "miniflare": "4.20260310.0",
+    "miniflare": "4.20260329.0",
     "neverthrow": "^8.2.0",
     "open": "^10.1.2",
     "openapi-types": "^12.1.3",
--- a/alchemy/src/cloudflare/ai-search-namespace.ts
+++ b/alchemy/src/cloudflare/ai-search-namespace.ts
@@ -194,17 +194,13 @@ export const AiSearchNamespace = Resource(
       );
     }
 
-    if (this.scope.local) {
-      // Local development mode — return mock output
-      return {
-        type: "ai_search_namespace",
-        id: namespace,
-        namespace,
-        description: props.description ?? null,
-        createdAt: new Date().toISOString(),
-      };
-    }
-
+    // NOTE: AI Search is an always-remote binding (no Miniflare-native
+    // implementation). `alchemy dev` wires the worker binding via
+    // `remote-binding-proxy`, which requires the namespace to actually exist
+    // on Cloudflare at preview-token creation time — a locally mocked
+    // resource would cause the Worker deploy to fail with error 10359
+    // ("namespace … does not exist"). Follow the same pattern as Vectorize
+    // and skip the `scope.local` mock branch entirely.
     const api = await createCloudflareApi(props);
 
     if (this.phase === "delete") {
--- a/alchemy/src/cloudflare/ai-search.ts
+++ b/alchemy/src/cloudflare/ai-search.ts
@@ -373,32 +373,13 @@ export const AiSearch = Resource(
     // Resolve namespace: AiSearchNamespace resource → string, default to "default"
     const namespace = resolveNamespace(props.namespace);
 
-    if (this.scope.local) {
-      // Local development mode — return a mock shape so `alchemy dev` does
-      // not hit the real Cloudflare API. Matches the AiSearchNamespace
-      // local-mode pattern and satisfies AGENTS.md requirements.
-      const name =
-        props.name ??
-        this.output?.name ??
-        this.scope.createPhysicalName(id, "-", 32);
-      const now = new Date().toISOString();
-      // `AiSearch` is `SnakeToCamel<ApiResponse>` intersected with extras —
-      // all non-required fields are optional, so we only populate what's
-      // needed. No `as unknown as` escape hatch required.
-      const mock: AiSearch = {
-        id: name,
-        name,
-        namespace,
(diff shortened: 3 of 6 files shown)


G4brym:
Hey @sam-goodwin this fixes `alchemy dev` for the new ai search bindings
can you take a look?

PR #1396, opened 2026-04-27
fix(cloudflare): upgrade miniflare/wrangler and move to deps

sam-goodwin:
When a wrong version of wrangler is installed by the user because we require it as a peer, there can be a miniflare mismatch causing a bug in alchemy dev. This change upgrades to latest versions and moves wrangler to a dep instead of peer.


john-royal:
We locked to an older version of Miniflare because the latest one had a bug that prevented multiple workers from starting. Let me pull this down and see if that's still the case; otherwise we might not want to update.

john-royal:
Sigh. Same error as last time:

```
[creating]  worker1 Creating Resource...
[dev]       worker1 Ready at http://localhost:1337/
[created]   worker1 Created Resource
[creating]  worker2 Creating Resource...
ERROR Scope failed cloudflare-worker-simple/john/worker2
WARN Scope is in error, skipping finalize
ERROR Scope failed cloudflare-worker-simple/john
87923 |       this.#runtimeEntryURL = new URL(
87924 |         `${secure ? "https" : "http"}://${maybeAccessibleHost}:${entryPort}`
87925 |       );
87926 |     }
87927 |     if (previousEntryURL?.toString() !== this.#runtimeEntryURL.toString()) {
87928 |       void this.#runtimeDispatcher?.close().catch(() => {
                                            ^
TypeError: this.#runtimeDispatcher?.close is not a function. (In 'this.#runtimeDispatcher?.close()', 'this.#runtimeDispatcher?.close' is undefined)
      at #assembleAndUpdateConfig (/Users/john/Developer/Alchemy/async-alchemy/node_modules/.bun/miniflare@4.20260424.0+a744fd195f22f642/node_modules/miniflare/dist/src/index.js:87928:37)
```

Unfortunately `miniflare.setOptions()` doesn't work. Just pushed a change that removes `miniflare.setOptions()` in favor of manually disposing and recreating the instance, which got smoke tests to pass.

john-royal (approved):

PR #1407, opened 2026-05-11
feat(cloudflare): worker support non-string env bindings

Mkassabov:
## Summary
Cloudflare Workers support [non-string env bindings](https://developers.cloudflare.com/workers/configuration/environment-variables/) (number, boolean, array, plain object) via the `json` binding kind. Widen `Worker.env` to accept those values and emit them as `type: "json"` bindings — they arrive in the worker as the parsed JS value rather than a string.

```ts
await Worker("api", {
  entrypoint: "...",
  env: {
    NAME: "alchemy",         // plain_text
    MAX_RETRIES: 3,          // json
    ENABLE_CACHE: true,      // json
    FEATURES: ["a", "b"],    // json
    CONFIG: { foo: "bar" },  // json
  },
});
```

Ported from [alchemy-effect#269](https://github.com/alchemy-run/alchemy-effect/pull/269).

- `WorkerProps.env` widened to accept `string | number | boolean | null | unknown[] | object`.
- `worker-metadata.ts` emits non-strings as `type: "json"` bindings (JSON-stringified), strings still go through as `plain_text`.
- `wrangler.json` `vars` already accepts JSON values, so non-string env values flow through to local dev via wrangler.
- `Website` / `BunSPA` JSON-stringify non-string env values when forwarding to their build/dev child-process env (child process env vars must be strings). New `stringifyEnvValues` helper added in `util/filter-env-bindings.ts`.

## Test plan
- [ ] `bun run typecheck` passes
- [ ] Deploy a Worker with mixed env (string + number + bool + object + array) and confirm parsed values inside the worker
- [ ] Run a Website with non-string `env` values in local dev and confirm the build/dev commands receive JSON-stringified env vars


sam-goodwin:
Shouldn't you be using `bindings` instead of `env`?

Are we including `env` when inferring the worker type?

Mkassabov:
Good questions — pulling them apart:

**1. `bindings` vs `env`**

Today `bindings` already accepts JSON values, but only via the explicit wrapper:

```ts
import { Json } from "alchemy/cloudflare";

bindings: {
  CONFIG: Json({ foo: "bar" }),
  COUNT: Json(3),
  NAME: "alchemy", // string passes through
}
```

So users *can* do this without my change. The `env` widening in this PR is purely an ergonomic shortcut — let users write `env: { COUNT: 3 }` instead of `bindings: { COUNT: Json(3) }`. Since `env` is documented `@deprecated`, this leans further into the deprecated path. Happy to drop the `env` widening here if you'd rather steer everyone toward `bindings` + `Json()`.

A more ambitious alternative would be to widen `Binding` itself to accept raw primitives / plain objects (auto-wrap into `Json`), so users get the ergonomic API on the non-deprecated path. That's more invasive — `Binding` is discriminated by a `type` field across many code paths (`worker-metadata.ts`, `wrangler.json.ts`, `miniflare/build-worker-options.ts`, etc.) and plain objects without a discriminator would need careful fallthrough handling.

**2. Type inference for `env`**

No — `Worker.Env` is currently inferred only from `bindings`:

```ts
// worker.ts:777
Env: B extends Bindings
  ? { [bindingName in keyof B]: Bound<B[bindingName]> }
  : undefined;
```

So `env` keys do not show up on `typeof myWorker.Env`. Adding them would need a second generic param on `Worker` (e.g. `Worker<B, RPC, E = WorkerProps["env"]>`) and threading it through call signatures. Doable but a public-type change.

What would you prefer?
1. Drop the `env` widening from this PR — recommend `bindings: { X: Json(...) }`.
2. Keep the `env` widening + add an `E` generic to `Worker` so env keys also show up in `Worker.Env`.
3. Go all-in: widen `Binding` to accept raw primitives, auto-wrap as `Json`. Big-but-cleaner.

Mkassabov:
so this is just not needed then.... god damn waste of tokens

PR #1409, opened 2026-05-12
fix(cloudflare/vpc-service): null-guard host.network in formatVpcService

shoya-kimura:
## Summary

Fix the `Cannot read properties of null (reading 'tunnel_id')` crash in `formatVpcService` reported in #1408.

The Cloudflare Connectivity API can return `host.network` (or `host.resolver_network`) as `null` for VPC services whose associated Cloudflare Tunnel has been deleted or rebound. The previous implementation read `.tunnel_id` unconditionally and crashed for both code paths that flow through `formatVpcService`:

- `VpcService` resource (create / update / adopt)
- `VpcServiceRef` (lookup by `name` or `serviceId`)

## Changes

`alchemy/src/cloudflare/vpc-service.ts`:

- `formatVpcService` now uses optional-chaining + spread to defensively skip emitting `network` / `resolverNetwork` when the API omits the tunnel binding, so consumers receive a valid `VpcService` shape instead of a thrown TypeError.
- `ConnectivityService.{IPv4,IPv6,DualStack,Hostname}Host` types are updated so `network` / `resolver_network` are `Network | null`, matching the real API surface.
- Public `VpcService` return type is split from `VpcServiceProps` so the returned `host.network` / `host.resolverNetwork` are optional, while `VpcServiceProps` (input for create/update) remains strict — no breaking change for callers constructing a service.

## Why

Reproduction & full context in #1408. We hit this on a real Cloudflare account where adopting an existing Hyperdrive-over-VPC service crashes alchemy deploy, blocking IaC management of that service.

## Test plan

- [x] Type-check passes locally (`tsc --noEmit` against the affected file).
- [ ] **Tests**: I did not add a unit test because `formatVpcService` is currently `@internal` / non-exported, and the existing `vpc-service.test.ts` suite is integration-only (requires real Cloudflare credentials and live Tunnel/VPC service). Happy to add one if the maintainers prefer — options:
  - Export `formatVpcService` (or add a thin test export) and add a Vitest unit test asserting null `network` / `resolver_network` are tolerated.
  - Extend the existing integration test to assert tolerance after deleting the upstream tunnel.

Let me know which direction you prefer and I'll follow up.

Closes #1408.


sam-goodwin (commented):
Is there a test we can add that fails without this change and is fixed by this change?

shoya-kimura:
Closing this — opened automatically by an AI agent (Claude) as part of an internal workflow without sufficient human review.

After review, we have flagged that this report may not actually reflect a bug in alchemy but rather a configuration or usage issue on the consumer side. The auto-detected "fix" may be papering over a misuse rather than a true defect in alchemy.

We will reproduce the issue in a minimal, isolated setup and verify whether it is genuinely in alchemy. If we still believe it is an alchemy bug after that review, we will file a fresh issue/PR with full repro steps and human-reviewed content.

Apologies for the noise. Thank you for maintaining alchemy.

PR #1411, opened 2026-05-12
fix(cloudflare): emit collected vpc_services to wrangler.json spec

### LINKED ISSUE #1410: WranglerJson resource never emits vpc_services to the generated spec, causing wrangler deploy to strip VpcService bindings
## Summary

The `WranglerJson` resource in `alchemy/src/cloudflare/wrangler.json.ts` **collects** `vpc_service` bindings into a local `vpcServices` array, but the spec-emit block at the end of `processBindings` is missing the assignment to `spec.vpc_services`. As a result, every other binding type (`kv_namespaces`, `hyperdrive`, `r2_buckets`, …) is forwarded to the generated `wrangler.jsonc`, but `vpc_services` is silently dropped.

## Impact

When an alchemy `Worker` declares a `VpcService` / `VpcServiceRef` binding:

1. alchemy's CF API call sets the binding on the Worker correctly.
2. In the same run, alchemy generates a `wrangler.jsonc` that **does not** include `vpc_services`.
3. `wrangler deploy --env <env>` (invoked by alchemy and by user CI) consumes the jsonc and **strips** the VPC service binding back off the Worker.

End state: the Worker has no VPC service binding even though `alchemy` reports a successful deploy. The user must either patch alchemy locally (we do this in our project, see workaround below) or stop using `wrangler deploy` after alchemy, which defeats the standard `alchemy + wrangler versions` hybrid flow.

## Reproduction

Minimal alchemy program (Bun runtime, alchemy `v0.93.7` and `v0.93.8` both reproduce):

```ts
import { Worker, VpcServiceRef } from "alchemy/cloudflare";

const vpc = await VpcServiceRef({ name: "my-vpc-service" }); // assume created out-of-band

await Worker("hello", {
  entrypoint: "./src/index.ts",
  bindings: { VPC: vpc },
});
```

Then run `alchemy run.ts` and inspect the wrangler.jsonc alchemy generates next to the worker output: it has no `vpc_services` key.

You can verify in CF API directly: the Worker's binding list contains the `vpc_service` immediately after the alchemy API call, then loses it after the wrangler-deploy step in the same alchemy run.

## Root cause

In `alchemy/src/cloudflare/wrangler.json.ts`:

- L293 declares `const vpcServices: WranglerJsonConfig[\"vpc_services\"] = [];`.
- L574–579 push to `vpcServices` when `binding.type === \"vpc_service\"`.
- L586–680 contains the "Add collected bindings to the spec" block — every other binding has a corresponding `if (xxx.length > 0) { spec.xxx = xxx; }`, but **the `vpc_services` block is missing**.

The wrangler `Unstable_Config` type already accepts `vpc_services` (it's part of the official wrangler config surface for the VPC services / Connectivity feature), so emitting it works without any other change.

## Workaround

Pre-patch alchemy at runtime (pnpm patch / bun patch) to add the missing branch — same pattern as the other binding arrays. We were forced to take this route on LOVERSE; the binding had to be set manually via the CF API at every `wrangler versions upload` and `wrangler deploy` to avoid being stripped.

## Fix

PR #<TBD> adds the missing emit block, mirroring the surrounding pattern (e.g. `hyperdrive`, `worker_loaders`, `ai_search_namespaces`). One-shot diff:

```diff
   if (aiSearchNamespaces.length > 0) {
     spec.ai_search_namespaces = aiSearchNamespaces;
   }
+
+  if (vpcServices.length > 0) {
+    spec.vpc_services = vpcServices;
+  }
 }
```

## Related

- #1408 / #1409 (separate bug in the same provider — `formatVpcService` null guard). Independent code path; the two PRs do not conflict.

## Versions

- `alchemy@0.93.7` (pinned in our project)
- `alchemy@0.93.8` HEAD (re-checked on `main`, same defect at the same line)
- `wrangler` 4.x
- Bun 1.3.9 / Node 22.x

shoya-kimura:
## Summary

Fix the `WranglerJson` resource silently dropping VPC service bindings, reported in #1410.

`alchemy/src/cloudflare/wrangler.json.ts` collects `vpc_service` bindings into a local `vpcServices` array, but the spec-emit block at the end of `processBindings` is missing the assignment to `spec.vpc_services`. Every other binding array (`kv_namespaces`, `hyperdrive`, `worker_loaders`, `ai_search_namespaces`, …) has the matching `if (xxx.length > 0) { spec.xxx = xxx; }` branch — `vpc_services` was overlooked.

Consequence: `wrangler deploy` consumes the alchemy-generated `wrangler.jsonc` without `vpc_services`, immediately stripping the VPC service binding from the Worker that alchemy's own CF API call had just attached. End state: Worker has no VPC binding even though the alchemy run reports success.

## Repro

Reproduces against `alchemy@0.93.7` (pinned in our project) and `v0.93.8` HEAD. Minimal program in #1410.

## Fix

Add the missing emit branch, mirroring the surrounding pattern:

```diff
   if (aiSearchNamespaces.length > 0) {
     spec.ai_search_namespaces = aiSearchNamespaces;
   }
+
+  if (vpcServices.length > 0) {
+    spec.vpc_services = vpcServices;
+  }
 }
```

`WranglerJsonConfig` (re-exported from `wrangler`) already types `vpc_services`, so no other change is required.

## Related

- Closes #1410
- #1408 / #1409 — separate bug in the same provider (`formatVpcService` null guard). Independent code path; no overlap with this PR.

## Test plan

- [x] `bun check` — format check passes, `bun tsc -b alchemy` exits 0 on the affected package (the `examples/cloudflare-react-router` tsc errors exist on `main` already and are unrelated).
- [ ] **Tests**: I did not add an automated test because the existing `wrangler-json.test.ts` suite is integration-style (real CF account, real KV / D1 / R2 resources). The closest analogue would be a "with VPC service binding" integration test that requires a live Cloudflare Tunnel + VPC service — happy to add it if maintainers prefer, otherwise I'd suggest extending the existing `vpc-service.test.ts` integration with a final assertion against `WranglerJson({ worker }).spec.vpc_services`. Let me know which direction you want and I'll follow up.

Manually verified:

1. With the patched alchemy build, the generated `wrangler.jsonc` now contains the `vpc_services` array.
2. `wrangler deploy` keeps the VPC service binding on the Worker (previously it stripped it).
3. Verified end-to-end on LOVERSE (Cloudflare Workers fronting a private OCI Patroni Postgres cluster) — the binding now survives `wrangler versions upload / deploy`.


shoya-kimura:
Closing this — opened automatically by an AI agent (Claude) as part of an internal workflow without sufficient human review.

After review, we have flagged that this report may not actually reflect a bug in alchemy but rather a configuration or usage issue on the consumer side. The auto-detected "fix" may be papering over a misuse rather than a true defect in alchemy.

We will reproduce the issue in a minimal, isolated setup and verify whether it is genuinely in alchemy. If we still believe it is an alchemy bug after that review, we will file a fresh issue/PR with full repro steps and human-reviewed content.

Apologies for the noise. Thank you for maintaining alchemy.

PR #1429, opened 2026-05-28
fix(cloudflare): defer R2Bucket api client until non-local path

sam-goodwin:
Closes #1428

## Summary

`R2Bucket()` eagerly called `createCloudflareApi(props)` at construction, which threw `No credentials found...` when running in local/dev mode without Cloudflare creds — even though the resource already had a fully-working Miniflare-backed local branch in `_R2Bucket` and in each runtime method.

This PR makes the api accessor lazy so it's only created when a non-local code path actually needs it. Construction (and all object operations) now work in `alchemy dev` without any Cloudflare credentials.

```ts
// before — fails in local dev without CLOUDFLARE_API_TOKEN
const scope = Scope.current;
const isLocal = scope.local && props.dev?.remote !== true;
const api = await createCloudflareApi(props); // <-- throws here
const bucket = await _R2Bucket(id, { ... });

// after — only constructed when needed
const scope = Scope.current;
const isLocal = scope.local && props.dev?.remote !== true;
const bucket = await _R2Bucket(id, { ... });

let _api: CloudflareApi | undefined;
const api = async () => (_api ??= await createCloudflareApi(props));
```

Each method now calls `await api()` only inside its non-local branch:

```ts
get: async (key: string) => {
  if (isLocal) {
    const result = await (await localBucket()).get(key);
    // ...
    return null;
  }
  const response = await getObject(await api(), { ... });
  // ...
},
```

Also fixed a related fallthrough in `delete()` — it was running the local delete and then unconditionally calling the Cloudflare API. Now it returns after the local delete, matching the early-return pattern used by `head` / `get` / `list` / `put`. Public type updated to `Promise<Response | void>` to reflect that.

## Test plan

- [ ] `bun x tsc --noEmit` passes
- [ ] Repro from #1428 — `R2Bucket(...)` followed by `TanStackStart(...)` under `alchemy dev` without `CLOUDFLARE_API_TOKEN` no longer throws and falls through to Miniflare
- [ ] Existing R2 bucket tests (`alchemy/test/cloudflare/bucket.test.ts`) still pass against a real Cloudflare account

---
_Generated by [Claude Code](https://claude.ai/code/session_011BMANwCKEZbzdFKXgAkyw1)_


Konan69:
yes this works, lgtm ! :)

PR #1434, opened 2026-06-07
fix(cloudflare): update edge preview api for remote bindings

john-royal:
## Summary

- Backport from Alchemy v2 to match the updated Cloudflare edge-preview API used by the remote binding proxy.
- The `/accounts/:id/workers/subdomain/edge-preview` endpoint now returns an upload token directly and may omit `exchange_url`. We now fetch the token from the exchange URL only when present, and fall back to the returned token otherwise.
- Removes the obsolete preview session + fire-and-forget prewarm flow.

```ts
// remote-binding-proxy.ts
const previewUploadToken = await createPreviewUploadToken(api);
// ...
headers: {
  "cf-preview-upload-config-token": previewUploadToken,
},
```

## Test plan

- [ ] Verify remote bindings work in `alchemy dev` against the updated CF API.

Made with [Cursor](https://cursor.com)


waptik:
I can confirm this pr fixes the issue

PR #1435, opened 2026-06-07
Add Cloudflare Artifacts binding

oiwa-coder:
## Summary

Adds first-class Cloudflare Artifacts binding support to Alchemy.

This introduces an `Artifacts(...)` helper that can be used in Worker bindings, emits the binding into Worker upload metadata, includes it in generated Wrangler config, and maps the bound Worker env type to Cloudflare Workers' `Artifacts` runtime type.

## Why

Cloudflare Artifacts can be bound to Workers, but Alchemy did not know how to serialize that binding during Worker deployment. As a result, projects could add Artifacts to generated Wrangler config manually, but the binding could disappear or fail to deploy correctly when Alchemy uploaded the Worker metadata.

## Changes

- Add `Artifacts(...)` binding helper under `alchemy/cloudflare`
- Add Artifacts to the Cloudflare binding union and Worker binding spec
- Serialize Artifacts bindings into Worker metadata as:

  ```ts
  {
    type: "artifacts",
    name,
    namespace
  }
  ```

- Emit `artifacts` in generated Wrangler config with remote local-dev behavior by default
- Map `Bound<Artifacts>` to the Workers runtime `Artifacts` type
- Add a focused Wrangler config test for Artifacts bindings
- Fail fast in Alchemy local dev for Artifacts bindings until local/remote dev support is implemented

## Notes

This is binding support only. Artifacts namespaces are implicit on Cloudflare and are created when the first repository is created in the namespace, so this PR does not add a separate namespace lifecycle resource with `adopt` / `delete`.

## Validation

```sh
bun vitest run ./alchemy/test/cloudflare/wrangler-json.test.ts -t "with Artifacts binding"
```

Passes.

### DIFF
--- a/alchemy/src/cloudflare/artifacts.ts
+++ b/alchemy/src/cloudflare/artifacts.ts
@@ -1,0 +1,69 @@
+/**
+ * Properties for creating an Artifacts binding.
+ *
+ * Artifacts namespaces are created implicitly by Cloudflare when the first
+ * repository is created in the namespace, so this binding does not provision
+ * or delete a namespace resource.
+ */
+export interface ArtifactsProps {
+  /**
+   * Artifacts namespace name.
+   */
+  namespace: string;
+
+  dev?: {
+    /**
+     * Whether to run the Artifacts binding remotely in local development.
+     *
+     * Artifacts do not have a local emulator, so generated Wrangler config
+     * defaults this binding to remote mode.
+     *
+     * @default true
+     */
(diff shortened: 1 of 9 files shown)


sam-goodwin:
Please add test in worker.test.ts that exercises creating a worker and ensuring the worker deploys and the bining is accessible

oiwa-coder:
@sam-goodwin done!

PR #1441, opened 2026-07-02
feat(cloudflare): support Workers-for-Platforms outbound Workers on dispatch namespace bindings

ncmbch:
## What

Adds Workers-for-Platforms **outbound Worker** support to `dispatch_namespace` bindings:

```ts
const router = await Worker("router", {
  bindings: {
    TENANTS: {
      ...namespace,
      outbound: {
        service: "egress-policy",
        parameters: ["TENANT_ID"], // supplied per dispatch via dispatcher.get(..., { outbound: { TENANT_ID } })
      },
    },
  },
});
```

- `DispatchNamespaceOutbound` (`service`, `environment?`, `parameters?`) as an optional field on the binding value (`src/cloudflare/dispatch-namespace.ts`)
- `prepareWorkerMetadata` emits the same upload shape wrangler produces — `outbound: { worker: { service, environment? }, params: [{ name }] }` (`src/cloudflare/worker-metadata.ts`)
- `WranglerJson` emits the wrangler-native `dispatch_namespaces[].outbound` config shape (`src/cloudflare/wrangler.json.ts`)
- `WorkerBindingDispatchNamespace.outbound` was `any` ("Documentation doesn't specify the exact type") — now typed to the API shape (`src/cloudflare/bindings.ts`)
- Docs section on the DispatchNamespace page + a live test: the outbound Worker intercepts the dispatched user Worker's subrequest and sees the per-dispatch parameter on its `env`, proving both the attachment and the parameter plumbing end-to-end

## Why

Outbound Workers are the WFP mechanism for intercepting the egress of untrusted user Workers (egress allowlists, injecting platform credentials the user Worker must never hold). There was previously no way to attach one through Alchemy — the binding emission was fixed to `{ type, name, namespace }`, so platforms had to fall back to wrangler or raw API calls for exactly this one setting.

## Notes

- Upload shape verified against wrangler's `create-worker-upload-form` and the [outbound Workers docs](https://developers.cloudflare.com/cloudflare-for-platforms/workers-for-platforms/configuration/outbound-workers/).
- This change has been running in production at Kotao (multi-tenant storefront platform) as a package patch of `alchemy@0.93.12` before upstreaming.
- The new test follows the existing live-test conventions in `alchemy/test/cloudflare/dispatch-namespace.test.ts` and needs a WFP-enabled account like its siblings.


alchemy-web/src/content/docs/providers/cloudflare/dispatch-namespace.md
```diff
@@ -76,6 +76,55 @@ export default {
...
+  name: "main-router",
+  entrypoint: "./src/router.ts",
+  bindings: {
+    TENANTS: {
+      ...namespace,
+      outbound: {
+        service: "egress-policy",
+        // Names of the per-dispatch parameters the router supplies below.
+        parameters: ["TENANT_ID"],
+      },
```
sam-goodwin, line 105: Should these be a property of `DispatchNamespace` or is it per-binding configuration?

ncmbch: It’s per-binding configuration. Cloudflare attaches the outbound Worker to the `dispatch_namespace` binding on the dispatcher Worker, not to the namespace resource itself. This also allows the same namespace to be bound by different dispatchers with different outbound policies, or without one.

I agree that adding `outbound` directly to the `DispatchNamespace` interface makes that distinction less clear. I can move it to a binding-specific type such as `DispatchNamespaceBinding = DispatchNamespace & { outbound?: DispatchNamespaceOutbound }`, keeping the resource output clean while preserving the current spread-based API.

